Issue #1526, opened 2026-07-03
design(dataflow): multi_tenant models use id-alone PK — two tenants cannot share a natural key


esperie:
## Summary

Surfaced while fixing #1518. A `multi_tenant=True` DataFlow model with a natural-key `id` (e.g. `id: str`) generates a table whose primary key is **`id` alone**:

```sql
CREATE TABLE "tenant_docs" (
    "id" TEXT PRIMARY KEY,
    "tenant_id" TEXT NOT NULL,
    ...
)
```

Because `id` is globally unique, **two tenants cannot hold the same natural key**. When tenant-b upserts a key that tenant-a already owns, the write fails closed with `UNIQUE constraint failed: tenant_docs.id` (SQLite) / the equivalent on PostgreSQL.

This is **fail-closed and safe** (no cross-tenant leak — verified in the #1518 regression `test_issue_1518_cross_tenant_upsert_cannot_corrupt_other_tenant_row`), but it means the #1518 acceptance criterion *"two tenants sharing a natural key get separate, correctly-scoped rows"* is structurally unmet: shared natural keys require the app to use globally-unique ids (e.g. UUIDs).

## Design question

Should `multi_tenant=True` models scope uniqueness by `(tenant_id, id)` (composite PRIMARY KEY / UNIQUE) instead of `id` alone, so tenants can share natural keys? This is a schema-generation change with real blast radius:

- **Migration / backward-compat** for existing `id`-alone tables.
- **SQLite + PostgreSQL** DDL parity.
- **Cross-SDK (EATP D6)** — the Rust SDK's multi-tenant schema must match semantics.
- **RETURNING / pre-check / upsert** paths that assume single-column PK.
- Interaction with `conflict_on` (#1508/#1519) — the conflict target must include `tenant_id`.

Alternatively, this may be **intended** (multi-tenant = globally-unique surrogate ids), in which case the fix is documentation + an actionable error naming `tenant_id` when a natural-key collision occurs across tenants.

## Related

- #1518 (tenant_id mis-map on upsert) — the fix that surfaced this. The `:pN` mis-map is independent and already fixed; this issue is the separate schema-semantics question.
- Siblings: #1519 (bulk `conflict_on`), #1520 (PG `conflict_on` error DX).

Not scoped into #1518's fix (distinct subsystem: schema generation vs query interception; exceeds a single shard).

PR #1650, opened 2026-07-10
fix(dataflow): surface actionable cross-tenant collision on bulk paths (#1526 follow-up)

esperie:
## Summary
Follow-up to PR #1646 (issue #1526). The merged fix added an actionable, tenant-scoped `TenantNaturalKeyCollisionError` on the **single-record** create/upsert path; the `bulk_create` / `bulk_upsert` paths were not wired for it (they use partial-failure-dict semantics, a distinct mechanism).

This PR wires the SAME actionable, tenant-scoped, no-cross-tenant-leak diagnostic into both bulk paths — delivered as a `collision` entry in each row's failure dict, **without** converting the bulk contract to raise-on-first-error.

- Extracted a shared message builder `format_tenant_natural_key_collision_message(...)`; the single-record message stays byte-identical.
- Added `_maybe_bulk_tenant_natural_key_collision` (bulk sibling of the single-record guard); `bulk_create` and `bulk_upsert` both enrich their `success: False` rows with an actionable `error` + `collision` entry.
- Added WARN partial-failure logging to both bulk paths (observability Rule 7).
- Fixed a pre-existing `bulk_upsert` silent-swallow (a failure was flattened into an empty-records dict with no error surfaced — zero-tolerance Rule 3).

## Security invariants (same as #1646)
The diagnostic names only the caller's own `tenant_id` + the caller's own supplied ids; tenant-scoped reads with `cache_ttl=0`; never reads another tenant's data; fails closed (an ambiguous/intra-batch case declines to a conservative refusal rather than an over-broad cross-tenant claim). No f-string SQL; no new cross-helper kwargs.

## Testing
- New Tier-2 regression suite (real DB, no mocking): `13 passed, 3 deselected` (PG-gated rows deselected — no Docker in the authoring session; SQLite equivalents + a direct real-DB helper test cover the same contract).
- Cross-tenant-no-leak test asserts tenant B's failure structure never contains tenant A's id or field values.
- Broader sweeps green: 44 (#1526 + bulk-upsert), 118 (bulk/express/tenant unit), 14 (integration).

## Note for reviewers
An orthogonal observation surfaced during implementation (NOT addressed here — flagged for the redteam): on SQLite, `bulk_upsert(conflict_on=["id"])` for a cross-tenant id collision resolves the conflict by UPDATE (`updated: 1`), which may mutate the other tenant's row via the ON CONFLICT path — a distinct cross-tenant-**write** concern, orthogonal to this diagnostic wiring. Under redteam verification; a tracking issue will follow if confirmed.

## Related issues
Follow-up to #1526 / #1646.

### DIFF
--- a/packages/kailash-dataflow/src/dataflow/core/exceptions.py
+++ b/packages/kailash-dataflow/src/dataflow/core/exceptions.py
@@ -19,7 +19,7 @@ loop into a single, loud, fail-fast error at the next access.
 from __future__ import annotations
 
 import re
-from typing import Optional
+from typing import Optional, Sequence
 
 # Re-export the public DataFlow base error so callers can `except
 # DataFlowError` to catch both legacy and new exception types without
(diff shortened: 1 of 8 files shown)


esperie:
Cross-SDK inspection (cross-sdk-inspection Rule 1): the same cross-tenant upsert write breach very likely exists in the Rust SDK's DataFlow. Paired issue filed on the Rust SDK BUILD repo (the Rust SDK's #1712).

Issue #2003, opened 2026-08-06
Parameter envelope doubles caller arguments persisted into workflow_runs.metadata (2x storage)


esperie:
Found by a residual sweep during the #1720 work on `fix/issue-1720-forest-drain`. Measured, not inferred. LOW-MEDIUM: nothing breaks and nothing leaks; durable storage is consumed at 2x on every run.

## The defect

The parameter envelope binds `{**P, "parameters": P}` so a workflow can read either bare top-level names or `parameters.get(...)`. That is correct at CALL time. But on the sync `LocalRuntime` path the BOUND mapping is what gets serialized into durable storage — so every caller argument is stored twice, in every run-history row, permanently.

Path: `src/kailash/tracking/storage/database.py:463::save_run` (the SQLite `workflow_runs.metadata` column), reached even with no explicit TaskManager via `local.py:4139::_flush_deferred_storage_sqlite` -> `deferred.py:144::flush_to_sqlite`.

Affected bind sites: `src/kailash/channels/api_channel.py:305`, `cli_channel.py:632`, `mcp_channel.py:465` and `:628`, `src/kailash/servers/enterprise_workflow_server.py:366`.

## Measurements

Paired against a RAW-mapping control through identical config:

```
save_run                          sentinels raw=3 / bound=6
                                  25452 -> 50154 bytes      ratio 1.971
workflow_runs.metadata @ 8 KiB     8470 -> 16690 bytes      ratio 1.970
workflow_runs.metadata @ 1 MiB  1048854 -> 2097458 bytes    ratio 2.000
local.py:305::_safe_serialize                               ratio 2.002
```

## What this is NOT — established, not assumed

**Not a size-cap halving.** The 10 MB cap lives in `nexus/validation.py` and applies only on the Nexus HTTP path, which was measured to have ZERO serializing sinks (AsyncLocalRuntime, no `idempotency_key`). The channels listed above apply no size cap at all.

**Not a checkpoint-key problem.** `durable.py:362::build_checkpoint_key` does `json.dumps` the doubled mapping, but yields a 64-hex digest and the persisted checkpoint row carries no inputs (verified with a sentinel). It is unreachable regardless: AST enumeration of all 10 bind-consumer call sites shows none passes `idempotency_key`.

Both were initial hypotheses that the measurement disconfirmed. Recorded so nobody re-opens them.

## Expected vs actual

[...]

Issue #1995, opened 2026-08-03
chore: pre-commit rewrites 2,022 files — repo-wide isort drift vs known_first_party


esperie:
Running `pre-commit run --all-files` on a clean tree rewrites **2,022 files**. The cause is
isort, not black — `black --check` is clean.

`pyproject.toml` declares:

```toml
[tool.isort]
profile = "black"
known_first_party = ["kailash", "kaizen", "kaizen_agents", "dataflow", "nexus", "pact"]
```

so isort wants all six in ONE import block, while the committed code separates them:

```diff
 from kailash.utils.url_credentials import fingerprint_secret
-
 from kaizen.utils.credential_scrub import scrub_credentials
```

Measured on `fix/issue-1720-forest-drain` @ `f7c30101f`: 2,022 files, of which **1,984 are
untouched by that branch** — this is repo-wide, pre-existing drift, not branch-introduced.
Split: 492 single-line (newline/whitespace class), 1,530 multi-line (import-regroup class).

## 1. Blocking-safety note

Touches no shipped runtime path. No workflow on any push or PR path runs isort or black —
`unified-ci.yml` runs ruff (`--select=E9,F63,F7,F82`), pyright (`continue-on-error: true`),
and pytest only. So this drift cannot red CI today; it is purely a local `pre-commit`
experience defect. Verified clean on the same tree: `ruff ... All checks passed!`, Tier-1
pytest Passed.

## 2. Value-anchor

Per the operator's instruction in the 2026-08-03 session — *"do a /sweep and consolidate
according to our new procedural directives"* — and the ratified disposition to file the
repo-wide normalization as its own item rather than bury it inside a bugfix PR. The user
value is that `pre-commit` is currently unusable as a pre-push gate: it reports 2,022
spurious modifications, which trains operators to ignore or bypass it.

## 3. Full-fix acceptance criteria

- [ ] Decide the intended contract: either the code adopts isort's grouping, or
      `known_first_party` is corrected to match the code's actual intent.
- [ ] Apply repo-wide in a single mechanical commit, no behavioural change.
- [ ] `pre-commit run --all-files` on a clean tree exits 0 and modifies **zero** files.
- [ ] Landed as its own PR, not folded into a feature or bugfix branch.

## 4. Revisit trigger

`after-milestone:forest-release`

Deliberately sequenced after the #1970/#1971/#1972/#1974/#1981 forest ships — a 2,000-file
[...]

Issue #1720, opened 2026-07-13
kaizen LLM: consolidate legacy providers/llm onto the four-axis LlmClient (retire the redundancy)


esperie:
## Context
#1717 (PR #1719) gave the four-axis `LlmClient` a real completion send path (Vertex-Claude/Gemini, Bedrock, OpenAI-family), closing the gap that made it orphaned. This unblocks — but does not perform — retiring the parallel **legacy** `providers/llm/` layer.

## Why the redundancy can't be removed yet (evidence)
The legacy layer is still the **sole production completion path**:
- `nodes/ai/llm_agent.py:2189` `_provider_llm_response` → `get_provider().chat(...tools)` (CRITICAL hot path)
- `core/base_agent.py:333` `_simple_execute_async` → `OpenAIProvider().chat_async` (multimodal)
- `nodes/ai/embedding_generator.py:661`, `providers/registry.py:165/206`
- ~10 distinct src consumer files, 34 legacy-touching test files.

The four-axis `CompletionRequest` has **no tools / structured-output / multimodal / mock-preset** — so a naive deletion regresses agent tool-calling. Four-axis is a config superset but a runtime subset.

## Proposed program (additive, ~6–10 autonomous sessions)
1. Bring four-axis to feature parity: tool/function-calling, structured output, streaming-with-tools, multimodal, a `mock` preset, remaining embed dispatch (cohere/hf/azure).
2. Dual-run behind `_provider_llm_response` (four-axis + legacy, compare) — no user-visible change.
3. Migrate consumers one at a time (llm_agent → base_agent → embedding_generator → registry).
4. Delete legacy `providers/llm/` + `base.py` + `registry.py` LAST, gated on a `tests/regression/` **four-axis quickstart** running `LlmClient.from_deployment(...).complete(...)` end-to-end (unit-green ≠ release-green).

## Highest risk
streaming-with-tools delta accumulation; `_provider_llm_response` tool-loop + structured post-parse parity; the mock-preset gap; the release-gate quickstart.

Full parity delta + consumer/test counts in the #1717 workspace journal (`workspaces/issue-1717-vertex-claude/journal/0002-parity-delta-legacy-vs-four-axis.md`).

Related: #1717.

esperie:
## Precise legacy-vs-four-axis head-to-head (2-agent, file:line-evidenced)

Confirms this is **NOT a true redundancy** — the two layers are different *tiers*, each more complete on a different axis. Neither is a superset.

[...]

Issue #2013, opened 2026-08-08
nexus: Nexus(enable_auth=True) and app.enable_auth() are inert — a documented security control that installs nothing


esperie:
## Summary

**`Nexus(enable_auth=True)` and `app.enable_auth()` are inert. Enabling authentication is indistinguishable from not enabling it.**

Measured — the control line is decisive:

```
CASE Nexus(enable_auth=True)   -> unauthenticated POST returns 200   _auth_enabled=True
CASE no auth at all (control)  -> unauthenticated POST returns 200   _auth_enabled=None
```

Ordering is NOT the cause. Both orderings behave identically, so this is not a "call `enable_auth()` before `register()`" footgun:

```
ORDER=auth-first   unauthenticated POST -> 200
ORDER=auth-last    unauthenticated POST -> 200
```

## Scope — what is NOT broken

**Nexus authentication itself works.** `NexusAuthPlugin` (`nexus/auth/plugin.py`) is real and wired: it installs `JWTMiddleware`, `RBACMiddleware` and `TenantMiddleware` via `app.add_middleware`. That is the functioning path.

**What is inert is the convenience facade** — `Nexus(enable_auth=True)` and `app.enable_auth()`.

> Anyone using `NexusAuthPlugin` is protected. Anyone following the README is not.

## Root cause — three dead links

1. `Nexus.enable_auth()` guards on `hasattr(gw, "enable_auth")`. **No gateway defines `enable_auth`** anywhere in `src/` or `packages/*/src/` — the only definition in the repo is `Nexus.enable_auth` itself. Dead branch.
2. It falls through to `use_plugin("auth")` → `AuthPlugin.apply()`, which constructs a real `MiddlewareAuthManager`, then attaches it via `if hasattr(gateway, "set_auth_manager")`. **`set_auth_manager` does not exist anywhere.** The manager is built and dropped.
3. The plugin sets `_auth_enabled = True` and `_auth_manager`. Across the whole nexus source those names are **written in 3 places and read in 0**.

A flag with no consumer is `zero-tolerance.md` Rule 2. A facade with no production call site is `orphan-detection.md` Rule 1. This is both.

## Why this is a documentation-backed security defect, not an internal wart

`README.md:48` lists `enable_auth=False,  # Default: False` under *"Optional: Configure enterprise features"*.

`README.md:188`, under *"DevOps Engineers — Add production features progressively"*:

```python
app = Nexus(enable_auth=True, enable_monitoring=True)
[...]

PR #2016, opened 2026-08-09
fix(#1720): close the credential-disclosure class in kaizen_agents + drain the forest

### LINKED ISSUE #1970: kaizen: MultiProviderNode leaks raw provider exceptions to user surface — #1953-parity sanitize sweep
Surfaced by the #1953 holistic redteam (Wave-R Round-2 correctness reviewer).

## Finding
`packages/kailash-kaizen/src/kaizen/nodes/ai/ai_nodes.py` `MultiProviderNode.run` (~L1547-1563) stores raw `str(e)` / `f"...{e}"` from a provider exception into a WARN log (L1548) AND into `result["primary_error"]` (L1554) / `result["error"]` (L1557/1563) — a genuine sibling of the credential-leak class #1953 closed in `iterative_llm_agent.py`, but in an unchanged module.

## Why it matters
A provider exception (bad-key/rate-limit) can embed a credential in its message; this reaches a user-visible error field + a WARN log unsanitized. #1953 (kaizen 2.44.0) added `sanitize_provider_error` at the iterative-agent surfaces; this node was outside that PR's scope.

## Scope
This is a kaizen-WIDE #1953-parity sweep, not a one-line fix — grep all `str(e)` / `{e}` sites reaching a user-visible dict or log across `packages/kailash-kaizen/src/` and route each provider-error path through `sanitize_provider_error`. Behavioral regression per gap.

## Acceptance
- [ ] MultiProviderNode error surfaces route through `sanitize_provider_error`
- [ ] Kaizen-wide sweep: no raw provider `str(e)` reaches a user-visible surface unsanitized
- [ ] Behavioral regression test injecting a fake-credential exception per fixed site

### LINKED ISSUE #1971: DataFlow: generated SQL identifier exceeds 63-char PostgreSQL limit in multi-database e2e
Surfaced by the Wave-R Round-2 closure reviewer (broader no-regression check).

## Finding
`test_multi_database_e2e` fails with `Invalid SQL identifier ... exceeds 63-char PostgreSQL limit (len=65)` on node `create_user`. DataFlow generates an identifier >63 chars, which PostgreSQL rejects (silent truncation risk / hard error). Orthogonal to the kaizen provider work — pre-existing.

## Why it matters
PostgreSQL truncates identifiers at 63 bytes; a 65-char generated identifier either errors (as here) or collides after truncation. Real correctness bug for multi-database DataFlow usage.

## Acceptance
- [ ] DataFlow generates PostgreSQL-safe identifiers (≤63 bytes, deterministic hashing/truncation on overflow)
- [ ] `test_multi_database_e2e` passes
- [ ] Regression test for the long-identifier case

### LINKED ISSUE #1972: kailash-nexus: 3 pre-existing unit failures (pydantic AnyUrl special-chars + FastMCP tool-registry KeyError)
Surfaced during #1959 (Nexus lifecycle) work — these reproduce identically on unmodified main, distinct bug class, correctly scoped out of #1959.

## Findings (reproduce on main)
- `test_core_comprehensive.py::...::test_special_characters_in_name` — pydantic `AnyUrl` rejects `workflow://...!@#$%^&*()`.
- `test_nexus_shared_runtime.py::TestNexusMCPToolClosure::test_workflow_tool_closure_uses_shared_runtime` — FastMCP tool-registry `KeyError`.
- `test_nexus_shared_runtime.py::TestNexusMCPToolClosure::test_workflow_tool_does_not_create_new_runtime` — same FastMCP `KeyError`.

## Why it matters
MCP/FastMCP-layer version-sensitive failures in kailash-nexus's own suite; pre-existing zero-tolerance debt.

## Acceptance
- [ ] AnyUrl special-char handling fixed or the workflow-name validation aligned
- [ ] FastMCP tool-registry closure tests pass
- [ ] All three green

### LINKED ISSUE #1974: kaizen: error_sanitizer pattern-coverage gaps — non-http conn strings, Slack tokens, bare JWTs
Surfaced by the Wave-F holistic security redteam (#1960 review). These are under-redaction gaps OUTSIDE error_sanitizer's stated LLM-provider threat model — deferred from the #1960 fix (which closed the AWS/uppercase-hex/Azure gaps).

## Gaps (all in `packages/kailash-kaizen/src/kaizen/nodes/ai/error_sanitizer.py`)
1. **Non-http(s) connection-string credentials** — `_URL_WITH_AUTH` anchors on `https?://`, so `postgres://user:pass@`, `redis://:pass@`, `mongodb://user:pass@`, `mysql://user:pass@` embedded passwords are NOT redacted. Concrete: `sanitize_provider_error(RuntimeError("postgresql://admin:hunter2@db:5432/x"), "x")` leaves `hunter2`. One-line scheme-broaden: `(\w+://)[^@\s]+:[^@\s]+@`.
2. **Slack tokens** (`xoxb-…`) and **bare (non-Bearer) JWTs** (`eyJ…`) slip — segment separators break the 40-char run and there's no prefix pattern. (Bearer-prefixed JWTs ARE covered.)
3. **Test hardening**: add a >40-char URL/path negative-vector to the sanitizer test suite (current negatives are all short).

## Why it matters
If these providers/surfaces enter scope, these are genuine unredacted-credential shapes. Low exploitability today (the module sanitizes LLM-provider exceptions, which rarely carry DB URLs / Slack tokens).

## Acceptance
- [ ] `_URL_WITH_AUTH` broadened to any `scheme://user:pass@`
- [ ] Slack + bare-JWT patterns added
- [ ] Long-path/long-URL negative vector confirms no over-redaction regression
- [ ] Behavioral regression per shape

### LINKED ISSUE #1981: kaizen: A2A capability scores are always 0.0 on providers without structured output
## Affected API

`kaizen.llm.reasoning.llm_capability_match` / `llm_text_similarity`, and the agents behind them (`CapabilityMatchAgent`, `TextSimilarityAgent`). Reached in production via `A2AAgentCard.calculate_match_score` → `A2ACoordinatorNode._find_best_agents_for_task`.

## Summary

`CapabilityMatchAgent` and `TextSimilarityAgent` do not enforce structured output. On any provider that returns prose rather than parseable JSON, structured-output parsing fails, `match_score` is absent, and the score coerces to `0.0`.

Consequence: **every** capability score is `0.0`, so A2A agent ranking is arbitrary — `_find_best_agents_for_task` returns all candidates tied at zero and the "best" agent is whichever the sort happens to emit first.

## Observed

Live run against a real provider, after the #1973 plumbing fix:

```
_find_best_agents_for_task -> [('eng', 0.0), ('des', 0.0)]

raw agent result:
{'response': '...## Confidence Score\n\n**0.92**\n\nThis is a high-confidence match...',
 'error': 'JSON_PARSE_FAILED'}
```

The model reasoned correctly and produced **0.92**. Only the transport of that number failed.

## Expected vs actual

- **Expected:** a capability match returns the model's score, or fails loudly enough that ranking is not silently degraded.
- **Actual:** returns `0.0` indistinguishable from a genuine no-match, and ranking degrades to arbitrary order.

## Already landed (partial mitigation, not a fix)

PR #1980 (#1973) made the condition observable: a degraded result now logs `*.degraded` at WARN with the underlying error, and is no longer cached. Before that it logged `*.ok` and **cached the fabricated zero**, pinning it for the process lifetime.

That change makes the failure loud. It does not make scoring correct.

## Severity

HIGH — a shipped feature silently produces meaningless output on an entire class of providers. No exception is raised, so nothing surfaces at the call site.

## Acceptance criteria

- [ ] `CapabilityMatchAgent` and `TextSimilarityAgent` request structured output (`response_format` / `structured_output_mode="explicit"`) where the provider supports it.
- [ ] A parse failure on a provider that does NOT support structured output is distinguishable from a genuine `0.0` score at the API surface (not only in logs).
- [ ] Cross-provider test matrix covering at least one structured-output provider and one without, asserting a non-zero score is returned for a high-confidence match.
- [ ] `_find_best_agents_for_task` does not return an all-zero ranking when the underlying judge degraded.

## Notes

Scoped out of #1973 deliberately: the fix spans provider configuration for shared reasoning agents and needs its own cross-provider matrix, with blast radius well beyond the coroutine bug #1973 addressed.

esperie:
## Summary

Closes the credential-disclosure class in `kaizen_agents` and drains the #1720 forest. Every
provider/exception path that reached a log record unscrubbed in that tree is now wrapped, pinned,
and adversarially reviewed.

**Scope, stated honestly up front: this sweeps ONE tree of four.** Re-derived on this branch:

| tree | scrub call sites |
| ---- | ---------------: |
| `packages/kaizen-agents/src` | 273 |
| `packages/kailash-kaizen/src` | 71 |
| `src/kailash` | **0** |
| `packages/kailash-nexus/src` | **0** |

The two zeros are the unswept surface. This PR does not claim the leak class is closed
platform-wide, and the follow-up shard is scoped from measurement rather than estimate (below).

## What lands

- **`kaizen_agents` sink sweep** — provider/exception text routed through `scrub_local_error` /
  `scrub_remote_error` at every reachable sink, with an AST scanner pinning the count so a new
  unscrubbed sink reds CI.
- **`safe_exception_frames` / `safe_callable_name` / `safe_type_name`** in
  `src/kailash/utils/secure_logging.py` — bounded, structurally inert identifiers for log records,
  applied at 14 sink sites across 7 modules.
- **Forest drain** — #1970, #1971, #1972, #1974, #1981.

## Diff

```
319 commits vs main · 418 files (356 .py)
production src: 28 files, +2267 / -238
```

## Test evidence

```
tests/regression/                                    1703 passed, 12 skipped   (1715 collected)
packages/kaizen-agents/.../test_local_error_sinks_are_scrubbed.py   379 passed
tests/unit/runtime + tests/unit/channels                            821 passed, 3 skipped
```

Every new pin was fail-first verified in an isolated worktree at its parent commit, and a
mutation matrix over 18 enumerated behaviours reds on all 18.

## Redteam history — 11 rounds, and it did not converge quietly

Reported as-is because the round count is the useful signal. Rounds 6–11 each found a real defect
in the previous round's fix. The recurring shape: **a fix hardened the layer it was looking at, and
the next round found the same class one layer over.**

- **R9** — the chokepoint (`_safe_identifier`) bounded identifiers and made them structurally inert.
- **R10** — that hardened the HELPER and left the SINK: 9 call sites logged `type(exc).__name__` raw
  as a sibling field in the same record. Measured: a class named `"A\nERROR …"` put a real newline
  into the record through the raw field while the helper's field rendered sanitized.
- **R11 (CRITICAL)** — `text in _CPYTHON_PSEUDO_IDENTIFIERS` is not byte-equality; it is `__hash__`
  then `__eq__`, both overridable on a `str` subclass. A subclass with a lying `__eq__` returned
  **5043 unsanitized characters** containing a newline, the chain separator and the frame anchor —
  bypassing the charset filter and the length bound together. Closed by normalizing to a plain
  `str` via `str.__str__` before any predicate runs.
- **R11 also** — `__suppress_context__` is shadowable by a class attribute that *lies* rather than
  raises, defeating `raise X from None` suppression. Closed by reading the walk attributes through
  `BaseException`'s own descriptors.
- **R11b** — a mutation matrix (62 mutations) found 13 pin gaps: behaviours that were CORRECT but
  defended by nothing, including both `_relative_frame_path` disclosure guards and
  `_DEFAULT_FRAME_LIMIT`, which is the live bound on 5 of 7 sinks. Nine re-derived as still open on
  this branch and are now pinned.

**A pin of ours was proven vacuous and deleted rather than left as reassurance.**
`test_safe_type_name_never_raises` used `@property def __class__`, but `type(obj)` never consults
`__class__` — it passed identically with the guard present or removed. Replaced with a metaclass
whose `__name__` raises.

**One linter suppression is load-bearing and deliberate.** Ruff `E721` ("use isinstance()") is
suppressed at the normalization line with an inline reason: `isinstance` admitting subclasses IS
the CRITICAL that line closes, so "fixing" E721 reintroduces the bypass verbatim.

## Known limits — please read before merging

1. **Tree-wide suites are NOT green-verified.** Root `tests/` does not run slowly, it **hangs** —
   42s CPU in 17 minutes, an asyncio selector wait, with no effective pytest timeout configured
   (`timeout=None` resolves for every tree). Adding `--timeout` makes it attributable;
   `--timeout-method=thread` kills the process on first hang, `signal` lets the run continue. The
   full three-tree run has not completed. `tests/regression/`, the sink scanner, and the
   runtime/channels unit suites are all green and were run repeatedly.
2. **~240 raw `type(<exc>).__name__` sites remain outside the 7 swept sink files** — measured:
   `src/kailash` 81, `kailash-kaizen` 60, `kailash-dataflow` 54, `kailash-nexus` 29,
   `kaizen-agents` 14. Most are ordinary Python idiom, not leaks; triaging them for
   attacker-controlled class names is its own shard, deliberately not scope-crept here.
3. **`base_async.py:304`** embeds the full `{e}` in an exception MESSAGE (not a log record) —
   runtime-hot-path shard, excluded by an exact-text rule in the sweep.
4. **#1995 (repo-wide isort drift, 2,022 files)** is deliberately NOT in this PR. Pre-commit was run
   scoped to this branch's 65 changed files; the repo-wide reformat lands after the merge so it does
   not make a 319-commit security diff unreviewable.

## Related issues

Fixes #1970
Fixes #1971
Fixes #1972
Fixes #1974
Fixes #1981


esperie:
## Accepted residuals — now issue-backed

The redteam produced accepted-pending residuals that lived only in session notes. Session notes are not a tracking surface and do not survive a workspace archive, so they are filed with revisit triggers and a **2026-09-08 calendar backstop** each:

- #2017 — sink scanner shape 9 misses message-bearing attribute chains in `extra={...}` (reachability in the swept tree is ZERO; blocked on a per-type allowlist design decision)
- #2018 — MCP `stop()` reports STOPPED over a still-live server thread (**pre-existing**, loudly WARNed, daemon thread)
- #2019 — MCP thread join blocks the event loop up to 5s (**pre-existing**; one-line fix, recommended for the #2013/#2014/#2015 shard rather than a third re-defer)
- #2020 — channel cleanup ordering skips the `close()` retry when `close()` raises (loud failure, not the false-success family)
- #2021 — `STOPPING` conflates a died event task with a live one (coupled: the M7 pin asserts current behaviour, so fixing it reds that test by design)

**Two of the seven were NOT filed.** The sweep report names `P2` and `P3` in its residual list, but no substantive content for either survives anywhere in the workspace artifacts — only the labels. Filing them would have meant inventing the findings, so they are reported as unrecoverable instead. If either was a real finding, it needs re-deriving from scratch.

Separately closed: #2003, at its fourth defer cycle, with owner approval and a re-open criterion. #2005 was deliberately NOT closed — its "tenant key falls open to a shared global namespace" reading is a tenant-isolation shape and wants re-triage as a possible BUG.

Issue #2027, opened 2026-08-10
HIGH/MED: sensitive values reach logs, cache keys, and a world-readable audit log (5 sites)


esperie:
## Summary

Five independent sites emit sensitive values to logs, cache keys, or world-readable files. Grouped into one issue because they share a single review lens — redact before emit (`observability.md` Rules 4 / 6.3 / 8) — and each fix is small and independent.

Each was confirmed by reading the cited line. Sibling lines that CodeQL flagged but which log only *names* or *fingerprints* are deliberately excluded here and dismissed separately.

---

### 1. MD5 over AWS secret material, used as a registry key — `src/kailash/gateway/resource_resolver.py:295` (CodeQL #11250)

`connection_config["aws_secret_access_key"] = credentials["secret_key"]` (line 286) is folded into `config_str = json.dumps(connection_config, sort_keys=True)`, then:

```python
s3_key = f"s3_{region}_{hashlib.md5(config_str.encode()).hexdigest()[:8]}"
```

MD5 over secret material, truncated to 32 bits, used as a key that reaches logs.

**Fix:** exclude secret fields from the key material entirely — region + endpoint + access-key-**id** identify the connection without the secret. If a digest is still wanted, `hashlib.blake2b(..., digest_size=8)`.

---

### 2. Workflow parameter **values** logged — `src/kailash/runtime/parameter_injector.py:411` (CodeQL #11246)

```python
f"Mapping workflow input {workflow_param} -> {node_param} for node {node_id} (value: {value})"
```

A workflow parameter named `api_key` / `password` / `token` is emitted verbatim. Note its immediate siblings at `:423` and `:445` log only `list(...keys())` and are correct — this one line diverges from the pattern the rest of the file follows.

**Fix:** drop `(value: {value})`, or emit `type(value).__name__` and length.

---

### 3. Raw pre-sanitization value logged — `packages/kailash-dataflow/src/dataflow/core/nodes.py:1034` (CodeQL #48)

```python
logger.debug(f"Sanitized SQL injection in field '{field_name}': {original_value} -> {value}")
```

Logs the user's raw input *before* sanitization. If the field is `password` or `token`, the secret lands in DEBUG. Also emits the schema field name, which `observability.md` Rule 8 requires be DEBUG-or-hashed.

**Fix:** log a hash of `field_name` plus a sentinel; never `original_value`.

(Its neighbour at `:1030` logs the *detector regex*, not user data, and is a false positive — excluded.)

---

[...]

Issue #2026, opened 2026-08-10
HIGH: production auth paths contain test stubs that forge tokens and log MFA codes


esperie:
## Summary

Two production auth paths contain test-compatibility shortcuts with security consequences: one **forges a successful OAuth token exchange** for attacker-influencable URLs, the other **logs the MFA one-time code in cleartext** while not sending anything.

Grouped because they are the same failure shape — a test convenience left in a shipped auth path — and want one reviewer pass on that theme.

---

## 1. SSO token exchange fabricates a successful response (CodeQL #105)

`src/kailash/nodes/auth/sso.py:924-931`:

```python
        except Exception as e:
            # For test compatibility, simulate successful token exchange if using example URL
            if "oauth.example.com" in token_url:
                return {
                    "access_token": "test_access_token",
                    "token_type": "Bearer",
                    "expires_in": 3600,
                    "refresh_token": "test_refresh_token",
                }
            else:
                raise ValueError(f"Token exchange failed: {str(e)}")
```

Two problems compounding:

- **The guard is a substring test, not a host test.** `https://oauth.example.com.attacker.tld/token` satisfies `"oauth.example.com" in token_url`. So does `https://evil.tld/?x=oauth.example.com`. Anyone who can influence `oauth_settings["token_endpoint"]` picks a URL that matches.
- **It triggers on the exception path.** The attacker does not need the exchange to succeed — they need it to *fail*, which is the easy direction. Point the endpoint at something that errors, and the function returns a synthetic "successful" authentication.

This is `zero-tolerance.md` Rule 2 (production stub) with an authentication-bypass consequence.

**Fix:** delete the branch. If a fixture genuinely needs it, gate on an explicit injected test double or an env flag — never on the content of a URL.

---

## 2. MFA SMS path logs the OTP at INFO (CodeQL #42)

`src/kailash/nodes/auth/mfa.py:29-32`:

```python
def _send_sms(phone: str, message: str) -> bool:
    """Module-level SMS sending function for test compatibility."""
    logger.info(f"SMS sent to {phone[-4:] if len(phone) > 4 else phone}: {message}")
    return True
```

`message` is the SMS body — i.e. the MFA code. The phone number is correctly truncated to its last 4 digits, which shows the author was thinking about disclosure; the OTP beside it is not redacted at all, at INFO, so it lands in any aggregator per `observability.md` Rule 4 and `security.md` § "No secrets in logs".

[...]

Issue #2025, opened 2026-08-10
CRITICAL: workflow-server proxy route is unauthenticated — arbitrary method/path/query forwarded to the backend


esperie:
## Summary

`WorkflowServer.proxy_workflow()` registers an **unauthenticated** catch-all route that forwards arbitrary methods, paths, and query strings to a proxied backend and returns the response body to the caller. Where the backend is an internal service — the usual reason to configure a proxy — this is an authentication bypass onto that service's entire surface.

## Evidence

`src/kailash/servers/workflow_server.py:761-770` — the route:

```python
        # Create proxy endpoints that forward requests to the remote server
        @self.app.api_route(
            f"/workflows/{name}/{{path:path}}",
            methods=["GET", "POST", "PUT", "DELETE", "PATCH"],
        )
        async def proxy_handler(request: Request, path: str, _url=proxy_url):
            """Forward requests to proxied workflow server."""
            import aiohttp

            target_url = f"{_url.rstrip('/')}/{path}"
```

and the forward, `:799-806`:

```python
            async with session.request(
                request.method,
                target_url,
                headers=headers,
                data=body,
                params=dict(request.query_params),
            ) as resp:
```

**No authentication exists on this path.** Grepping the whole file case-insensitively for `Depends|add_middleware|auth|api_key|require_auth` returns exactly one hit — `add_middleware` at :326, which is CORS. There is no auth dependency, no API-key check, no auth middleware.

Compounding it: the `Authorization` header is deliberately stripped on forward (`:779`, `_sensitive_headers`), so the backend cannot re-authorize the caller either. The caller's identity is removed and nothing replaces it.

## On the CodeQL framing

This was surfaced by CodeQL alert #11252 as `py/partial-ssrf`. **That framing is wrong and the issue should not be worked as an SSRF.** `_url` is developer-supplied via the `proxy_workflow(name, proxy_url, ...)` Python API; its only two production callers (`src/kailash/channels/api_channel.py:497`, `src/kailash/api/gateway.py:500`) are Python APIs, not HTTP endpoints, so scheme/host/port are fixed at registration. `path` is appended after a `/` and therefore lands in the path component — no authority pivot was constructible (`@evil.com/` yields `http://backend:8080/@evil.com/`).

The severity is right; the mechanism is not. Fixing "SSRF" here (e.g. validating `_url`) would close nothing.

## What is already right

[...]

Issue #2024, opened 2026-08-10
CRITICAL: FileSecretBackend stores secrets in cleartext while documenting encrypted storage


esperie:
## Summary

`FileSecretBackend` documents itself as encrypted storage and writes secrets to disk in **cleartext**. There is no cipher anywhere in the class.

## Evidence

`src/kailash/gateway/security.py:188-189` — the claim:

```python
class FileSecretBackend(SecretBackend):
    """Secret backend using encrypted file storage."""
```

`src/kailash/gateway/security.py:205-213` — what it actually does:

```python
    async def store_secret(self, reference: str, secret: Any) -> None:
        """Store secret in file."""
        file_path = os.path.join(self.secrets_dir, f"{reference}.json")

        with open(file_path, "w") as f:
            if isinstance(secret, str):
                f.write(secret)
            else:
                json.dump(secret, f)

        # Set restrictive permissions
```

and the read path, `:195-203`, is a plain `json.load`. The only protection applied is `os.chmod(file_path, 0o600)` after the write.

This is **not** a case of the module lacking crypto. The same file imports and uses Fernet:

```
16:from cryptography.fernet import Fernet
64:            self._cipher = Fernet(encryption_key.encode())
77:        self._cipher = Fernet(key)
```

— but in a different class. `FileSecretBackend` never touches `_cipher`.

## Why this is more than a docstring bug

The docstring is the contract a caller reads before deciding what is safe to put in this backend. A developer choosing `FileSecretBackend` for production credentials on the strength of "encrypted file storage" gets plaintext-at-rest, and `0o600` is not a substitute — it protects against other local users, not against backup capture, volume snapshots, image layers, or anything that reads the filesystem out of band.

Per `zero-tolerance.md` Rule 2 this is the fake-encryption pattern explicitly enumerated as BLOCKED.

## Acceptance criteria

Pick ONE of:

- [ ] **Encrypt at rest.** Reuse the module's existing Fernet setup; derive the key from an env-supplied KEK, never a generated-and-discarded one (note `:71` already warns about exactly that failure mode for the sibling class). Round-trip test: bytes on disk do not contain the plaintext secret.
- [ ] **Delete the class** and route callers to `EnvSecretBackend` / a real secret manager, with a `DeprecationWarning` shim per `zero-tolerance.md` Rule 6a.

Not acceptable: weakening the docstring to match the behaviour and leaving plaintext writes in place under a name that says otherwise.

[...]

PR #2028, opened 2026-08-10
chore(#1995): make pre-commit converge — 5 broken hook gates, 262 lint findings, 8 real bugs

esperie:
## Summary

Makes `pre-commit run --all-files` converge. It previously rewrote ~2,000 files on a clean tree and never reached a fixed point, so the gate was unusable as a pre-push check and operators had learned to bypass it. Four cycles of deferral.

**Acceptance criterion met** — on a clean tree:

```
$ git status --porcelain | wc -l
0
$ .venv/bin/pre-commit run --all-files
EXIT=0   ALL HOOKS PASSED
$ git status --porcelain | wc -l
0
```

## The issue described this as isort drift. It was five defects.

**1. Five competing isort configs disagreed on `known_first_party`.** This is the root cause.

| config | `known_first_party` |
|---|---|
| root `pyproject.toml` | kailash kaizen kaizen_agents dataflow nexus pact |
| `packages/kailash-kaizen` | *(none)* — `kailash` was THIRD-party here |
| `packages/kailash-dataflow` | *(none)* — same |
| `packages/kaizen-agents` | 3 of the 6 |

isort resolves config from the nearest file, so the same import was classified first-party or third-party depending on which config won for that path. That is the blank line between `import pytest` and `from kailash...` that flip-flopped every run. All three per-package blocks now carry the identical six-entry list.

**2. `src_paths` was unset**, so isort inferred first-party membership from whichever *batch* of files it was handed — `--all-files` and the commit-time hook (staged files only) classified the same import differently. Necessary but not sufficient on its own; pinned explicitly.

**3. Every directory carve-out in `.pre-commit-config.yaml` was inert.** Written `^(a/|b/)$`; pre-commit fullmatches the repo-relative path, so a `dir/` alternative under a `$` anchor can only match a path that IS literally `"dir/"`. None had ever fired — including the pre-existing ones.

**4. `end-of-file-fixer` was rewriting 304 audit fixtures, including `.expected` answer-key files.** Those bytes are what the detectors are measured against; normalizing them silently weakens the eval harness. Fixture trees are now excluded from the whitespace and JSON gates — the deliberately malformed JSON under them is malformed on purpose. **Audit fixtures are byte-untouched by this PR.**

**5. black/ruff crashed on scaffold templates** (`extend-exclude` is ignored for explicitly-passed paths, which is how pre-commit invokes black; `force-exclude` is the only form honoured), and `kailash-align` silently opted out of the root lint config since ruff resolves from the nearest `pyproject.toml`.

## 262 ruff findings the converged gate exposed — 8 were real bugs

Across 9 rule classes. The per-package configs select more than root, so these had never all been visible at once.

- **5x F821** — undefined `registry`/`tool_count` in `kaizen-agents` autonomy examples, left behind by the ToolRegistry→MCP migration. **These NameError at runtime today.** Fixed by describing current behaviour, not by inventing a registry.
- **3x `from __future__ import annotations` placed BEFORE the module docstring** (`trust/constraints/budget_store.py`, `budget_tracker.py`, `trust/plane/session.py`). That silently un-docstrings the module — the string becomes a dead expression statement.

~50 bare `except:` were narrowed **per-site** to the exception actually raised there (`asyncpg.UniqueViolationError`, `requests.RequestException`, `NodeConfigurationError`, `black.InvalidInput`, …) rather than blanket `except Exception:`, per `zero-tolerance.md` Rule 3. Guarded E402 imports kept their position with a `# noqa: E402` and a reason; unguarded ones moved to the top.

## Also fixed in passing

`packages/kailash-kaizen/grafana/prometheus.yml` had `[]` at the same indent as its own `targets:` key — not valid YAML.

## Verification

- Per-package `pytest --collect-only` across all 8 packages, no new import errors: 20239 + 15185 + 6876 + 2663 + 671 + 2602 + 4159 + 1756.
- Two pre-existing gaps unchanged: pact's uninstalled `hypothesis`, dataflow's 250 marker-deselections.
- Full 21-hook `pre-commit run --all-files` green and idempotent across consecutive runs.

**One flake observed and dispositioned, not hidden:** `test_concurrent_execution_performance` failed once during verification — it asserts a 3x speedup on a ~30ms workload and measured 0.031s vs 0.033s while this machine was running parallel agents. Passes 3/3 in isolation, and this diff is import ordering only. Filed separately rather than retried into silence.

## Risk

The 1,700-file import normalization carries no behavioural change. The load-bearing risk was import-order side effects, which is why collection was verified per-package rather than trusting a green hook.

## Related issues

Refs #1995


esperie:
CodeQL disposition — **artifact of the line-keyed diff baseline, not a regression.**

CodeQL reports "377 new alerts including 6 high severity". This PR reorders imports across ~1,700 files, which shifts nearly every line number in the repo; CodeQL keys its new-vs-existing diff on alert *location*, so every shifted alert re-reads as new.

Tested rather than assumed. Comparing the full alert sets at `(rule, security_severity, path)` granularity — deliberately NOT at line granularity, since line is precisely what the reorder perturbs, so a line-keyed comparison could not discriminate the hypothesis from its negation:

```
gh api ".../code-scanning/alerts?ref=refs/pull/2028/merge&state=open" --paginate ...  # 2319 alerts
gh api ".../code-scanning/alerts?ref=refs/heads/main&state=open"     --paginate ...  # 2345 alerts
comm -23 pr.txt main.txt   # present on PR, absent on main
comm -13 pr.txt main.txt   # present on main, absent on PR
```

Falsifying result, had the PR genuinely introduced alerts: `comm -23` would be non-empty.

- **`comm -23` → EMPTY.** Zero `(rule, severity, path)` tuples exist on this PR that do not already exist on main. The alert set is a strict subset of main's. `comm` on sorted input accounts for multiplicity, so per-tuple *counts* match as well, not merely membership — a new alert of an existing rule in an already-flagged file would still have surfaced.
- **`comm -13` → 26 rows**, every one `py/unused-import` or `py/unused-local-variable` in `kailash-ml` / `kailash-align` / `kailash-pact`. This PR *removes* them.
- Totals reconcile exactly: 2345 − 2319 = 26.

Net: this PR strictly **reduces** the CodeQL alert surface and introduces nothing.

The ~2.3k alerts standing on main are **not dismissed by this** — they remain open and are tracked in the burn-down plan, with several of these exact high-severity sites (`gateway/security.py` clear-text-storage, `base_agent.py` + `mfa.py` + `parameter_injector.py` clear-text-logging, `resource_resolver.py` weak-hashing, `cursor/hook.py` overly-permissive-file, `sso.py` incomplete-url-sanitization) already filed as #2024 / #2025 / #2026 / #2027 / #2030 and queued for this burn-down.

All other checks are SUCCESS on the pinned head `8c20f9ed46fa9b0f135c40e09957e55c70a6f050`.

Issue #2023, opened 2026-08-10
CI: five workflow steps install root kailash editable but resolve siblings from PyPI — latent version-skew class


esperie:
Filed from the PR #2016 session. **Latent, not currently failing** — no listed step exercises the affected path today. Filing so the CLASS is tracked rather than the one instance that broke.

## The class

A workflow step that installs the root `kailash` package editable from the branch, but lets a sibling package resolve from PyPI, is testing a combination the monorepo never ships: branch-core + released-sibling. When the branch changes a cross-package interface, that step fails in a way no source inspection explains — or, worse, passes while the real shipped pairing is broken.

The instance that broke, in `test-kailash-ml.yml` Coverage:

```
TypeError: _validate_identifier() missing 1 required keyword-only argument: max_length
.venv/lib/python3.12/site-packages/dataflow/core/nodes.py:3620
```

Root `kailash` came from the branch (carrying #1971's now-required keyword-only `max_length`); `kailash-dataflow` came from PyPI at 2.19.1, which still calls it positionally. 18 tests red. Fixed by installing dataflow editable — the discipline every other workflow already followed.

## Remaining sites

| file | line(s) |
| --- | --- |
| `.github/workflows/test-kailash-ml.yml` | 211, 247, 317, 392 |
| `.github/workflows/gpu-smoke.yml` | 117 |
| `.github/workflows/test-kailash-align.yml` | 109, 163, 198 |

## Why this is NOT a blanket "add the install everywhere" fix

None of these steps currently exercises the dataflow path: no kailash-align test imports `kailash_ml.features`, and the dl/rl marker filters do not select the feature-store files. Blanket-adding editable installs would lengthen every job and risk pulling heavy optional dependencies into jobs that do not need them, for no present benefit.

The correct fix is per-step: **install a sibling editable when the step actually exercises that sibling.** That is a judgment per job, not a sweep — which is precisely why it wants an issue rather than a drive-by commit.

## Acceptance criteria

- [ ] For each of the 8 sites, determine whether the step exercises a sibling package's code path (import graph or marker selection), and record the answer in the step comment so the next reader does not re-derive it.
- [ ] Where it does, install that sibling editable; where it does not, leave it and say why in-comment.
- [ ] Add a CI-lint or review-checklist item asserting: any step installing root `kailash` editable must declare, in a comment, which siblings it exercises and how each is resolved.
- [ ] Verify no step regresses in wall-clock beyond an agreed budget.

## Not claimed

[...]

Issue #2022, opened 2026-08-09
kailash-ml: km.from_brief fails end-to-end — BaseAgentConfig built without llm_provider (README Quick Start broken)


esperie:
Found while diagnosing PR #2016 CI. Not a blocker for that PR — filed separately.

## Affected API

`kailash_ml.from_brief` (and the documented README Quick Start that calls it).

## The defect

`packages/kailash-ml/src/kailash_ml/from_brief.py:736` constructs the agent config with no provider:

```python
agent_config = BaseAgentConfig(
    model=resolved_model,
    strategy_type="single_shot",
)
```

`llm_provider` is never passed. The kaizen `#1952` provider-integrity work (on main, kaizen 2.44.0, commit `4bf28de3b`) removed the silent `provider="mock"` default, so a config with an unresolved provider no longer degrades quietly. The `kailash_ml.from_brief` call site was not swept when that default was removed.

## Observed

`packages/kailash-ml/tests/integration/test_ml_from_brief.py` — 3 failures. The LLM result comes back structurally EMPTY and validation then rejects it:

```
raw = {"dataframe_schema": {}, "eval_metric": "", "feature_columns": [],
       "interpretation_confidence": None, ...}

pydantic_core._pydantic_core.ValidationError: 1 validation error for MLPlan
interpretation_confidence
  Input should be a valid number [type=float_type, input_value=None, input_type=NoneType]
src/kailash/_from_brief/validator.py:204: ValidationError
```

Affected: `test_from_brief_classification_returns_triple`, `test_from_brief_regression_returns_triple`, `test_readme_quickstart_executes_end_to_end`.

**Attribution caveat, stated rather than glossed:** an all-empty LLM result is consistent with EITHER an unresolved provider OR an unusable API credential in the local environment. The observation above does not by itself discriminate between them. What IS established independently of the environment is the structural defect: line 736 passes no `llm_provider`, and the mock default it used to rely on is gone. Someone reproducing this should confirm which failure they are seeing before assuming.

## Why CI does not catch it

The three tests are Tier-3 real-LLM and skip when no model env var is configured, so they never run on CI. `from_brief.py` lines ~702-794 show as uncovered in the CI coverage report. The failure is only visible where a `.env` is present — i.e. for the developer or user following the README, which is the worst place to discover it.

## Expected vs actual

Expected: the documented Quick Start (`km.from_brief(...)`) returns a plan triple.
Actual: raises `ValidationError` on an empty plan.

[...]

Issue #2015, opened 2026-08-09
core: 24 sites return a raw exception in an HTTP response body (detail=str(e)) — outside #2012's scope


esperie:
## Affected API

HTTP error paths in `kailash` core and the DataFlow starter templates. Enumerated below.

## Summary

24 sites pass a caught exception straight into an HTTP response body via `HTTPException(..., detail=str(e))`. The exception at these sites is whatever the workflow, driver, or provider raised — the same class of object shown elsewhere in this repo to carry DSNs, API keys and bearer tokens.

This is a **wider** surface than the log sinks swept so far: a log record is internal, whereas an HTTP 500 body is returned to whoever made the request, and on a gateway that can be an unauthenticated caller.

## Enumeration (production code, tests and build artifacts excluded)

```
10  src/kailash/visualization/api.py
 6  src/kailash/middleware/communication/api_gateway.py
 3  src/kailash/gateway/api.py
 2  packages/kailash-dataflow/templates/api_gateway_starter/example_app/routes/users.py
 2  packages/kailash-dataflow/templates/api_gateway_starter/example_app/routes/organizations.py
 1  src/kailash/channels/api_channel.py
--
24  total
```

Reproduce with:

```bash
grep -rnE 'detail=str\(e\)|detail=str\(exc\)' src/ packages/ --include='*.py' \
  | grep -v '/build/\|\.venv\|/tests/'
```

**24 is a floor, not a total.** That pattern matches only the `str()` spelling. At least one f-string variant exists and is not counted — `src/kailash/servers/durable_workflow_server.py:480` (`detail=f"Recovery failed: {e}"`). Any sweep should cover both spellings.

**The four template hits deserve separate handling.** They are scaffolding under `templates/api_gateway_starter/example_app/`, which users copy into their own applications — so they propagate the pattern rather than merely exhibiting it.

## Expected vs actual

**Expected:** an HTTP error body carries a caller-safe message and a correlation id; the raw exception text stays server-side.

**Actual:** the raw exception text is returned to the client.

## Severity

**MEDIUM-HIGH**, and it depends on the site. Severity is driven by (a) whether the endpoint is reachable unauthenticated and (b) whether the underlying exception can carry connection state. Per-site triage is needed — this is an enumeration, not 24 confirmed credential leaks.

## Why this is not covered by #2012

[...]

Issue #2006, opened 2026-08-07
fix(dataflow): CREATE sends explicit NULL for unset optional fields, defeating column DEFAULT


esperie:
## Affected API

`dataflow` CREATE path — `db.express.create(...)` / the generated `<Model>CreateNode`.
Observed on `kailash-dataflow` 2.12.0 (`dataflow/core/nodes.py`).

## Minimal repro

```python
from dataflow import DataFlow
from datetime import datetime

db = DataFlow("postgresql://localhost/test")

@db.model
class Event:
    id: str
    ts: datetime = None          # optional; the column has a server default

# Column exists as:  ts TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP
await db.express.create("Event", {"id": "e1"})
# asyncpg.exceptions.NotNullViolationError:
#   null value in column "ts" of relation "events" violates not-null constraint
```

Second, independent symptom — a callable default corrupts generated DDL with no error:

```python
def _now(): return datetime.now(timezone.utc)

@db.model
class Probe:
    id: str
    ts: datetime = _now          # registers without complaint

db._generate_create_table_sql("Probe")
# CREATE TABLE IF NOT EXISTS "probes" (
#     "id" TEXT PRIMARY KEY,
#     "ts" TIMESTAMP DEFAULT '<function _now at 0x10c1fb740>',   <-- repr() of the function
#     "created_at" TIMESTAMP DEFAULT CURRENT_TIMESTAMP,
#     "updated_at" TIMESTAMP DEFAULT CURRENT_TIMESTAMP
# );
```

## Expected vs actual

**Expected:** a field the caller did not supply, with no usable model default, is omitted from the
INSERT column list so the database applies its own column `DEFAULT`.

**Actual:** the CREATE path builds the column list from *every* model field, with exactly three
exemptions (`nodes.py` ~1706-1714):

```python
for name in model_fields.keys():
    if name == "id":
        if "id" in kwargs:               # omitted when absent
            field_names.append(name)
    elif name not in ["created_at", "updated_at"]:
        field_names.append(name)          # always included
```

Unsupplied optional fields are then filled with `None` (~1793-1795) and sent as an explicit `NULL`,
so a column `DEFAULT` is unreachable for every field except those three. Note the mechanism to omit
a column already exists and is applied to `id` alone.

**The two layers also disagree about callable defaults.** The INSERT path handles them
[...]

PR #2031, opened 2026-08-10
codify: specification-verification — an acceptance criterion is a hypothesis, not a measurement

esperie:
## Summary

Codifies the failure class found in the 2026-08-10 burn-down: **five issues had wrong SPECIFICATIONS rather than wrong code**, each of which a diligent session would have implemented faithfully and closed wrongly.

Both gating questions that held this draft for three sessions are now **resolved by measurement**. Ready for review.

## The failure

Every gate validates against the specification: the implementer writes the fix to the AC, the test author writes the test to the AC, the reviewer checks the diff against the AC, the issue closes because the AC is satisfied. **No gate measures whether the AC was true.** So when the specification is wrong, the wrongness is invisible to all of them at once — not because a gate was skipped, but because they share one unverified premise.

Structurally identical to a non-discriminating instrument (`instrument-discipline.md`): the check's output is the same whether the proposition is true or false. Here the instrument is the AC itself.

## Four MUST clauses

1. **Re-derive every load-bearing factual claim** before implementing — counts, paths, line numbers, "which branch", the named root cause. Cite the command.
2. **Test any PRESCRIBED remedy against the named threat** before adopting it. A remedy that does not close the threat MUST NOT be implemented just because the AC requires it.
3. **Build your own enumeration instrument**; never inherit the spec's grep — that reproduces its blind spots and reads as corroboration.
4. **Correct the specification** when measurement contradicts it. Silently implementing the corrected version leaves the next reader the original wrong premise.

## MUST-2 carries the sharpest evidence (#2004)

The issue prescribed routing a leaked credential through `mask_error_text`. Tested against the issue's own threat example:

```
RAW str(e):      spawn command 'npx -y @vendor/server --token=sk-live-ABCDEF123456' ...
mask_error_text: spawn command 'npx -y @vendor/server --token=sk-live-ABCDEF123456' ...
TOKEN STILL PRESENT AFTER MASK?: True
```

That helper masks URL userinfo and query params, not CLI flags — and no scrubber in the monorepo covered the shape. Implementing AC#1 literally ships a change that looks correct, passes a test written to the AC, and still leaks. **Following the specification exactly is what produces the vulnerability.**

---

## Resolution 1 — scope: path-scoped stands, MEASURED

The question was whether this ships `priority: 0 / scope: baseline` instead. Both halves were tested rather than argued.

**Budget.** `node .claude/bin/emit.mjs --all --dry-run` at this branch's base (`21deef321`):

| state | rules | emission | codex headroom | gemini headroom |
|---|---|---|---|---|
| base (this file path-scoped) | 11 | 53168 B | **13.46%** | **13.46%** |
| same command, this file flipped to `priority: 0 / scope: baseline` | 12 | 61465 B | **-0.04% → `headroom-floor BLOCK`** | **-0.04% → `headroom-floor BLOCK`** |
| final tree (path-scoped, all changes) | 11 | 53168 B | 13.46% | 13.46% |

Baseline is not merely expensive here — **the emitter refuses it**, 6169 B under the 10% floor on both lanes. Landing it would require ~8.3 KB of extraction out of the other eleven baseline rules. The corpus is already inside Rule 10's 15% proximity band with 2128 B of slack above the floor.

The falsifying result is named in the rule: a post-promotion headroom at or above 15% would have removed the budget objection outright. The third row is the leak check — byte-identical to base, so this rule costs zero baseline emission.

**Reachability.** The `issue-triage-routing.md` precedent applies when the trigger moment matches **no** glob (a `gh issue` triage touches zero files). This rule is not that shape: MUST-1/2/3 have a subject only when something is IMPLEMENTED, and implementation terminates in a file edit — which is why `ee8274122` widened the globs to catch the YAML-only CI case (#2023).

**Residual, stated in the rule rather than glossed:** a MUST-4-only session — measures a contradiction with Bash and `gh`, posts the correction, edits nothing — matches no glob and does not load this rule. Closing it costs ~200 B of pointer inside an already-loaded baseline rule, which fits the measured 2128 B slack but fires Rule 10's gate and so needs its own paired-extraction or named-rationale receipt. Filed as a separate shard, not silently dropped.

## Resolution 2 — probe set, registered (not deferred)

Registration was reported as possibly blocked. It was not blocked — it was **unconfigured**, and measurably so:

```
node .claude/bin/coc-manifest-integrity.mjs   -> exit 1
node .claude/bin/coc-eval-all.mjs             -> exit 1
"no eval-manifest at .../eval-manifest.json — the eval harness is UNCONFIGURED,
 which is a FAILURE, not a pass"
```

kailash-py had no `.claude/test-harness/eval-manifest.json` at all, so **both** the registration path and the `_deferred_probes` path were unreachable — the deferral declaration lives inside the manifest it would have to be written into. The honest fix is the manifest, not a deferral. Both gates now exit 0.

The manifest is **declared, not silently minimal**. Zero structural entries is the true state here, stated with reason + graduation + expiry (`2026-11-12`):

```
ls .claude/bin/*-readiness-check.mjs                          -> 0 matches
grep -rn 'coc-eval-all|test-harness|eval-manifest' .github/   -> 0 hits
grep -rlc 'runs-on' .github/workflows/                        -> 26 files   <- control
```

The third command is the control: it establishes the second's emptiness is a real absence rather than a broken search. `_declared_no_pins` is declared on the same evidence, converting an un-adopted WARN that asserts nothing into a stated exemption that expires.

**Suite — 5 bipolar pairs, 10 rows**, one pair per MUST clause plus a meta pair. A one-sided suite cannot discriminate, which is this rule's own thesis.

| pair | violation pole | compliant pole |
|---|---|---|
| `must1-rederive-claims` | implements to the filed count and "auth path worst" | re-derives both, posts the divergence |
| `must2-validate-prescribed-remedy` | adopts the prescribed helper untested | runs it on the threat input first, corrects the AC |
| `must3-independent-instrument` | re-runs the issue's own grep, reads agreement as corroboration | builds a defect-keyed instrument, reconciles the 25th site |
| `must4-correct-the-specification` | measures the contradiction, implements it silently | posts the correction before implementing |
| `meta-rule-conformance` | surface-matched adversarial rule, 6 breaches in content | synthetic rule meeting every applicable meta-rule |

Each violating transcript is a session doing everything else right — it plans, edits, writes a test, and the test passes. `flag-inherited-grep` even runs a command and quotes its output; the defect is that the command is the specification's own, so agreement is guaranteed rather than informative. A judge separating this corpus by "did the transcript run a command" scores at chance on three of the four efficacy pairs.

The meta pair is surface-matched: both poles share frontmatter, section skeleton, intro, MUST NOT clause, Origin and Wiring block, and sit **3.05% apart on size** (3275 B vs 3175 B). All six defects live in what the affordances contain — a `should` modal under a MUST heading, a paraphrased rather than verbatim BLOCKED corpus, a DO-only example, a four-sentence `Why:`, hedging inside the clause, and a Wiring block missing `Violation scope`.

**Registration verified end to end, not asserted:**

```
coc-probe-dispatch.mjs plan -> 10 rows to dispatch, 0 refused, 5 pairs
identifier-shuffle control  -> 0 divergences over 10 rows   (production rendering)
same control, pre-fix label -> 10 divergences               (so the control CAN fail)
```

The second control run is deliberate. A control that cannot fail is not evidence (`instrument-discipline.md` MUST-2b), so the pre-fix rendering is reproduced to show the fixture filename **would** be a live polarity channel if it reached a judge, and does not.

## Detection block

Was "no hook detector (deferred)" with a fixtures path that did not resolve — which `coc-artifact-eval-coverage.md` MUST-4 blocks. It now names the full binding (scanner `null`, fixtures dir, probe file), every path resolving.

**Phase 2 is declined, not deferred.** Whether a number was MEASURED or COPIED leaves no lexical trace, so a structural detector over the transcript would be the non-discriminating instrument `instrument-discipline.md` MUST-1 blocks. Saying "deferred" would imply a detector is coming; it is not, and the reason is this rule's own subject matter.

## Other design decisions

- **Deliberately NOT on `self-referential-codify.md`'s allowlist** — a universal implementation-quality rule firing in all sessions, the same disposition `zero-tolerance.md` carries. Recorded in Origin per the omission-precedent shape.
- **Canonical 8-field Trust Posture Wiring** present and verified one-each (`Severity` / `Grace period` / `Cumulative posture impact` / `Regression-within-grace` / `Receipt requirement` / `Detection mechanism` / `Violation scope` / `Origin`).
- **270 lines**, over the 200-line guidance, with a named rationale anchored at Origin. The count was re-derived with `wc -l` at the final edit; the prior `~217` predated the two receipt blocks. The first correcting edit stated `266` and was itself stale by four lines on write — the same defect at one remove, and why the number was measured again rather than reasoned about.

## Verification

- `coc-manifest-integrity.mjs` — PASS (was exit 1)
- `coc-eval-all.mjs` — exit 0 under the honest `NO STRUCTURAL COVERAGE` banner (was exit 1)
- `emit.mjs --all --dry-run` — validator-14/18 PASS, headroom unchanged from base
- `validate-xref-integrity.mjs` — this branch contributes **0** dangling refs; corpus-wide the count fell **212 → 206** (creating the eval-manifest resolved references `coc-artifact-eval-coverage.md` already made to it). The remaining 206 are pre-existing and overwhelmingly `journal/`, which a `coc-build` repo does not carry.
- pre-commit — full hook set green on both commits, no `--no-verify`.

## Not done in this shard

- The ~200 B baseline pointer that would close the MUST-4-only reachability residual (fires Rule 10; needs its own receipt).
- `.claude/rules/ci-runners.operator.local.example.md` has **no frontmatter**, which hard-fails `emit.mjs` validator-14 in this repo — a pre-existing break that made the headroom measurement above require a scratch-tree workaround. Reported, not fixed here: the correct classification for an operator-local template is a separate call.

## Related issues

Refs #2004, #2006, #2015, #1997, #2022, #2013, #2023


esperie:
## Tier-1 redteam complete — the rule failed its own MUST-1, and has been amended

Three agents (cc-architect, reviewer, analyst) per `self-referential-codify.md` Rule 1. Verdict: **land amended**. Amendments in `638a6c487`.

**The headline: a rule about verifying claims shipped three unverified claims.**

| # | Finding | Status |
|---|---|---|
| CRIT | MUST-1's DO example carried a **fabricated comparand** — "the 27 filed"; #2015's title files **24** | fixed |
| CRIT | **Internal contradiction** — 24 in MUST-1, 25 in MUST-4/extract, one measurement | fixed |
| CRIT | **#2013 replacement line numbers wrong** — draft said 4813/4826; actual 4814/4825, and 4782 is `"health_monitoring": True`, a status-dict key, not any method | fixed + public correction posted on #2013 |
| HIGH | **Scope under-reaches** — MUST-4 fires on `gh issue comment` (zero file edits); originating set includes a YAML-only CI fix (#2023) | stopgap globs added; see open question |
| HIGH | **No eval coverage** — no probe set, no manifest entry | OPEN |
| MED | "the test passes" reads as a general prohibition on green tests | reworded to "the test I wrote to the AC passes" + explicit note |
| MED | 217 lines, no named rationale | added |

**One reviewer correction was itself wrong.** It proposed `base_agent.py` 734/743; re-derivation returns **728/735/744**, so the original 735/744 stood and only 727 was off by one. Every amendment here is the orchestrator's own grep, not the redteam's numbers taken on trust — which is the rule's MUST-1 applied to the redteam.

**Two open questions before merge:**

1. **Should this be baseline rather than path-scoped?** The stopgap globs (md/yml/yaml/toml/json/sh/sql/tsx/jsx + workflows) cover the counterexamples found, but cc-architect's argument is structural: MUST-1/3 fire *before* the fix is written, and a path-scoped rule only loads at the first matching edit — i.e. after the decision point. That is the same reachability shape that sent `issue-triage-routing.md` baseline. Going baseline fires `rule-authoring.md` Rule 10's proximity gate, so it needs paired extraction or a named rationale.

2. **Eval coverage is genuinely absent.** `coc-artifact-eval-coverage.md` MUST-1 requires a probe set for a new prose artifact and MUST-4 blocks a Detection block that names none. Minimum: `probes/specification-verification.probes.json` with efficacy + no-false-positive + meta-compliance, each carrying both a `violation` and a `compliant` scenario, registered probe-only (`scanner: null`) — or a `_deferred_probes` declaration with reason + graduation + expiry.

**Analyst's proposed mechanical signal, worth considering separately:** don't try to detect "was this measured" (undetectable). Detect the *receipt* — require MUST-1's citation as a fixed trailer (`Re-derived: <cmd> → <output>`), then the check is: any PR body containing `Fixes #N` asserts ≥1 `Re-derived:` line. Structural, discriminating, seconds to run.

Staying **draft** until (1) and (2) are dispositioned.

esperie:
## Both open questions dispositioned — out of draft

Closing the loop on the two items this PR was held for.

**(1) Baseline vs path-scoped — path-scoped, and it is not a preference.** `emit.mjs --all --dry-run` with this file flipped to `priority: 0 / scope: baseline` returns `headroom-floor BLOCK: -0.04% < 10% floor (under by 6169B)` on **both** codex and gemini. The emitter refuses baseline; landing it would cost ~8.3 KB of extraction out of the other eleven baseline rules. Base is 13.46% headroom (already inside Rule 10's proximity band, 2128 B of slack above the floor), and the final tree measures byte-identical to base — so the rule costs zero baseline emission.

The reachability argument was the stronger one and it partly survives. `issue-triage-routing.md`'s precedent applies when the trigger matches **no** glob; here MUST-1/2/3 have a subject only when something is implemented, and implementation terminates in a matched file edit. The genuine residual is a MUST-4-only session — measures with Bash and `gh`, posts the correction, edits nothing. That is now **stated in the rule** with its price (~200 B of pointer inside an already-loaded baseline rule, which fits the slack but fires Rule 10) and filed as its own shard rather than dropped.

**(2) Eval coverage — registered, not deferred.** The blocker was misdiagnosed as "registration is hard"; it was that kailash-py had **no `eval-manifest.json` at all**, so `coc-manifest-integrity.mjs` and `coc-eval-all.mjs` both exited 1 and *both* the registration path and the `_deferred_probes` path were unreachable (the deferral declaration lives inside the file that did not exist). Manifest created with `_declared_empty` + `_declared_no_pins`, each carrying reason + graduation + expiry on measured evidence. Both gates now exit 0 — the structural one under the honest `NO STRUCTURAL COVERAGE` banner, which is the correct report for a repo with zero structural scanners.

Suite is 5 bipolar pairs / 10 rows, one per MUST clause plus a surface-matched meta pair (poles 3.05% apart on size, all six defects in content). `coc-probe-dispatch.mjs plan` → 10 rows, 0 refused. The identifier-shuffle control reports 0 divergences in production and **10 divergences** when driven with the pre-fix rendering — the second run is there because a control that cannot fail is not evidence.

**On the analyst's proposed `Re-derived:` trailer** from the redteam comment: not adopted, and worth saying why. It is a good structural signal for *whether a receipt was written*, but the rule's subject is whether a number was **measured**, and a trailer is as cheap to write from the issue's own figure as from a fresh grep. It would detect the ceremony, not the property — which is the same substitution MUST-3 blocks one layer up. Phase 2 is therefore recorded as **declined**, not deferred, so nobody waits for a detector that should not be built.

Two things I did not do, stated plainly: the ~200 B baseline pointer above, and `.claude/rules/ci-runners.operator.local.example.md`, which carries no frontmatter and hard-fails `emit.mjs` validator-14 in this repo — a pre-existing break that forced the headroom measurements through a scratch tree. The right classification for an operator-local template is a separate call, so it is reported rather than guessed at.

PR #2034, opened 2026-08-10
fix(#2015): route HTTP error bodies through one client-safe helper

### LINKED ISSUE #2015: core: 24 sites return a raw exception in an HTTP response body (detail=str(e)) — outside #2012's scope

esperie:
## Summary

Every HTTP error body that rendered a caught exception now routes through one shared helper, `kailash.utils.http_errors.safe_http_detail`. The caller gets a generic message plus a short correlation id; the full exception — type, masked text, frames — goes to the server log under that same id, so the fix does not trade a disclosure bug for an undebuggable service.

## The claim in the issue is wrong, and it matters for prioritisation

The issue says the auth path is worst. **Refuted, by measurement.**

`middleware/auth/auth_manager.py` has **1** leaking site. `visualization/api.py` has **10** — and that server has **no authentication of any kind**: no `Depends`, no `Security`, no auth dependency on any of its 14 routes. Anyone who can reach the port could read the exception. Driven through a real `TestClient`, the pre-fix body of an unauthenticated `GET /api/v1/runs` is:

```
{"detail":"...could not connect to postgresql://svc_user:sup3rs3cret@db.internal:5432/kailash"}
```

That is the headline case, and it is the one pinned by the regression test.

## Measured counts (both spellings, whole repo)

| Spelling | Occurrences | Files |
|---|---|---|
| `detail=str(e)` | **24** | **6** — 20 in `src/kailash/**` across 4 files, 4 in `packages/kailash-dataflow/templates/**` across 2 |
| `detail=f"` | **43** | 19 |

Two corrections to the brief this branch was given. The `detail=str(e)` total of 24 spans **six** files, not four — "exactly 4 files" holds only if you scope to `src/kailash/**`. And of the 43 `detail=f"` sites, only **1** interpolated an exception (`auth_manager.py`); the rest interpolate workflow ids, permission names and format lists, and were correctly left alone.

## The grep in the issue does not find the real surface

Counting `detail=` understates this bug. These sites render an exception into a response body without ever spelling `detail=`, and each was verified to leak before being fixed:

- `api/gateway.py` — `GET /mcp/tools` returned `{"error": str(exc)}` per server as the body
- `api/workflow_api.py` — the SSE `event: error` frame streamed `str(e)` to the subscriber
- `servers/connection_metrics_router.py` — `collect()`'s dict is returned by `GET /metrics` **and** `GET /pools`. Highest real-world credential risk in the set: the exception comes from a database driver, and driver connect failures quote the connection string
- `servers/enterprise_workflow_server.py` — 5 sites, incl. all four `/enterprise/health` probes. The secret-manager probe is the one likeliest of all to name a secret backend
- `channels/api_channel.py` — `ChannelResponse.error` and the emitted event payload

## Scope note: templates

`packages/kailash-dataflow/templates/api_gateway_starter/**` wrapped validation *and* DataFlow execution in one `try` and rendered `except ValueError` as a 400 with `str(e)`. So a database failure was reported as a client-side "Validation Error" carrying the DSN. Split into two blocks: validation keeps `str(e)` verbatim (that message is written for the caller and is the point of a 400), execution returns `INTERNAL_ERROR` at 500 via the helper. Fixed here rather than deferred because users copy templates into production code that never passes through this repo's review.

## Verification

Fail-first, not assumed. Every claim below is from a run, and each fix was confirmed to make a previously-red test go green:

- Regression suite red against the pre-fix code: **5 failed, 4 passed**, output quoting the leaked password in the response body
- Same suite after: **11 passed**
- The two new non-`detail=` tests independently confirmed red before their fixes
- `pre-commit run --all-files`: **exit 0**, every hook Passed, zero Failed (this is the form pre-commit.ci runs)
- **No regressions.** The selected suite was run against the pre-branch baseline in a throwaway worktree: baseline **47** failing, HEAD **46**, set difference of new failures **empty**. The 46 are pre-existing (MCP validation, security test helpers, durable-gateway e2e needing infra) and are reported, not absorbed here

Two existing tests asserted `"Execution failed" in response.error` — the vulnerability written down as expected behaviour, which is part of why it survived. Both inverted.

## Also fixed

- **`packages/kailash-dataflow/pyproject.toml` was missing `src_paths`.** The root config carries it with a comment naming this exact non-determinism (isort inferring first-party membership from whichever batch of files it is handed) and crediting it for the ~2,000 permanently-modified files #1995 cleaned up. isort resolves config from the nearest file upward, so nothing under `packages/kailash-dataflow/` ever got that fix, and the two template files oscillated between the commit-time and `--all-files` runs — the commit could not be made to stick. Blast radius verified: only the two files this branch already touches move.
- The WIP commit had silently converted `visualization/api.py` from CRLF to LF, rendering the change as 957 insertions / 916 deletions and burying the 20 lines that alter exception handling. Restored; the same change is now 61 / 20.

## Not fixed here — reported instead

- `api/gateway.py` `_sender()` swallows `asyncio.CancelledError` rather than re-raising. Different bug class; changing task-cancellation semantics inside a credential fix would hide an unreviewed concurrency change behind it.
- The three route handlers in `connection_metrics_router.py` log nothing on entry/exit.
- A framework-first hook flags `from fastapi import` in the two template files. That import is pre-existing (line 7 of both, before this branch) and untouched; converting a DataFlow FastAPI starter template to Nexus is a template redesign, not a credential-disclosure fix.
- **No leak sites were deferred for lane collisions.** The files held by other lanes were checked: the two `detail=f"` sites in `nexus/core.py` are a rate-limit message and a not-found message, and `nexus/auth/rate_limit/middleware.py` has none.

## Related issues

Fixes #2015
Follow-up to #1995 (dataflow isort `src_paths`)


esperie:
## Review follow-up: import defect does not reproduce; coverage gap did, and is closed

**On the reported `reportUndefinedVariable`** — it does not reproduce on this branch. The import has been present since `247b1c80a`:

```
src/kailash/servers/enterprise_workflow_server.py:21:from ..utils.http_errors import safe_http_detail
```

`pyright` reports **0 diagnostics** on that file. I did not stop at a clean result, since a clean run is only evidence if it could have come out the other way — deleting that single line reproduces the review's finding *exactly*, same rule and same lines:

```
reportUndefinedVariable: 5
  line 396 - "safe_http_detail" is not defined
  line 432 - "safe_http_detail" is not defined
  line 443 - "safe_http_detail" is not defined
  line 469 - "safe_http_detail" is not defined
  line 490 - "safe_http_detail" is not defined
```

The exact line-number match (396/432/443) places the original run against the pre-import state.

Swept all 13 call-site modules for the same omission, via AST rather than grep: every module that calls `safe_http_detail` binds it.

## The second point was right, and it was a real gap

Driving only some error paths was exactly the weakness described: a missing import in an `except` branch is invisible unless the branch executes. The prior tests drove the visualization routes, `/mcp/tools` and `collect()`, but *reasoned about* the rest. Now closed two ways:

1. **All four `/enterprise/health` probes are driven through their failure paths**, called unbound against a stub `self` so the real method bodies execute — including the module-global lookup — without standing up an enterprise server.
2. **A structural sweep** asserting every module calling the helper binds it. That covers the sites sitting behind servers too expensive to stand up, and keeps covering sites added later.

Both verified to fail when the defect is present. With the import deleted, the behavioural test fails with the precise request-time failure mode described in review:

```
src/kailash/servers/enterprise_workflow_server.py:443: in _check_resource_health
    resource_health["error"] = safe_http_detail(
E   NameError: name 'safe_http_detail' is not defined
```

and the structural test fails naming the offender. Restored: **13 passed**.

## Site inventory

`enterprise_workflow_server.py` was already listed in the PR body under the non-`detail=` sites, with its 5 sites and the note that all four health probes feed `GET /enterprise/health`. The full inventory and the reconciliation between the 24/43 counts and the files actually touched is in the body above.

## Gate

`pre-commit run --all-files` exit **0**, zero failed hooks. Regression + api_channel suites: **37 passed**.

esperie:
## H1 / H2 / H3 fixed — each with a test verified to RED without it (`58745033b`)

All three confirmed against the code and fixed. Two of my first-attempt tests **did not discriminate**, and I rewrote them rather than keep a green that proved nothing — detail below, because the failures are instructive.

### H1 — `gateway/api.py` no-op allowlist

Confirmed. `safe_types=(ValueError,)` under `except ValueError` meant every reachable exception was inside the allowlist by construction, so `_GENERIC_BY_STATUS` was unreachable at that site — and the `try:` spanned `WorkflowResponseModel(**response.to_dict())`, so a `pydantic.ValidationError` returned the offending input values as a **404**.

Fixed by scoping the `try:` to the lookup alone and **dropping the allowlist entirely**. I did not swap in a typed `WorkflowNotFound`: `enhanced_gateway.py:300` raises `ValueError(f"Request {request_id} not found")`, whose message tells the caller only what it already sent, so the generic fail-closed 404 costs nothing. Serialization now has its own 500 arm — a model failure is a server fault, not a missing resource.

### H2 — template `list_*` handlers

Confirmed, and **broader than reported**: `list_users` in `users.py` has the identical shape and is fixed too. `create_*` was split in the first pass; `list_*` was not.

I removed the `except ValueError` arm from the execute block rather than leaving it ahead of the generic one. A `ValueError` subclass surfacing from DataFlow is an infrastructure fault, not a client mistake, and routing it to a 400 was the mechanism.

### H3 — `workflow_api.py` masked only the branch that doesn't fire

Confirmed, and your framing was the important part: a real `NexusHandlerError` carries a dict or str body, so the two branches that fire in production were exactly the two skipping the mask. All three now route through a new `mask_response_body`, which walks nested structures, preserves non-string leaves and the response shape callers parse, and is **depth-bounded** so a self-referential body can't turn the sanitizer into a stack overflow.

### The two tests that didn't discriminate

- **H1**: my assertion searched the function source for `"safe_types"` — and matched the word inside its own explanatory comment. Now asserted over the **AST** (`ast.keyword` args, `ast.Try` count).
- **H3**: my first test exercised `mask_response_body` directly, so unwiring **both** call sites left it **green**. Now driven through the real `_execute_sync`; with the fix reverted it REDs on both shapes:

```
E   AssertionError: typed str body leaked: {"error":"connect failed: postgresql://svc_user:sup3rs3cret@db.internal:5432/kailash"}
```

Negative controls for the rest: H1 REDs with the old allowlist restored; H2 REDs naming all four pre-fix sites; M1 REDs with `RuntimeError: __str__ exploded` propagating out of the sanitizer.

## M1, M4, M5 — done in this pass, not deferred

- **M1**: totality is now *enforced* by `mask_exception_text` rather than claimed in prose, and the docstring says so. `url_credentials.py` is untouched (lane 1E holds it), so the guard lives in `http_errors.py`.
- **M4**: a companion test pins the allowlist's **real** reach — bearer tokens, CLI flags, cloud keys and paths are asserted **NOT** masked. Widening or narrowing it now forces the test and the docstring to move together.
- **M5**: the file now **raises at import** unless `kailash.utils.http_errors.__file__` resolves under its own `repo_root`. The AST sweep also no longer skips attribute-style `http_errors.safe_http_detail(...)` calls.

**M5 caught a real error of mine, worth recording.** Under `pytest` (rootdir = worktree, `pythonpath = src`) everything resolves to the worktree, so all my red/green evidence was genuinely measuring this branch. But bare `python -c "import kailash…"` resolves to the **main checkout** — so the `import OK` checks I cited earlier in this PR were inert, validating main's source rather than mine. The test evidence stands; those ad-hoc checks did not, and I've stopped using them.

## Gate

`pre-commit run --all-files` exit **0**, zero failed hooks. Regression suite **20 passed**. Still **zero** new test failures versus the pre-branch baseline (46 vs 47, set difference empty).

Nothing deferred from your list — H1/H2/H3 and M1/M4/M5 are all in `58745033b`. No follow-up issues needed.

Issue #803, opened 2026-05-03
bug(tests): test_mfa_functional.py — 9 test/production drifts beyond kwarg mismatch (#778 followup)


esperie:
## Affected API
`tests/integration/nodes/test_mfa_functional.py` against `kailash.nodes.auth.mfa.MultiFactorAuthNode`

## Context
After #778's `operation=` → `action=` kwarg rename (commit 23f6e5fd), test count moves from 16 fail / 7 pass / 2 skipped to 9 fail / 14 pass / 2 skipped. The remaining 9 failures are heterogeneous test/production assertion drift — the test file was authored at the same commit as the production code (`9364027b`) with assertion data that does not match the production response shape it was validating.

## Failures (9)

| Test | Failure | Root cause |
|------|---------|-----------|
| test_totp_setup_process | `'TestApp' in 'otpauth://totp/KailashSDK:user123?...&issuer=KailashSDK'` | Production hardcodes `issuer="KailashSDK"`; test expected configurable issuer |
| test_backup_codes_generation | `assert 10 == 5` | Production returns 5 default backup codes; test expects 10 |
| test_totp_verification_failure | `assert True is False` | Verify failure path returns success=True for invalid code |
| test_backup_code_verification | `KeyError: 'user_id'` | Production response dict missing `user_id` field |
| test_mfa_disable_process | `KeyError: 'user_id'` | Same |
| test_mfa_reset_process | `assert False is True` | `action="reset"` is unsupported by production dispatch (only setup/verify/generate_backup_codes/revoke/status/send_push/verify_push/trust_device/check_device_trust/set_preference/get_methods/disable/initiate_recovery) |
| test_mfa_status_check | `KeyError: 'user_id'` | Same as above |
| test_multiple_method_support | `KeyError: 'enabled_methods'` | Production response shape lacks `enabled_methods` key |
| test_empty_user_id_handling | response shape mismatch | Test expects `error` key OR `user_id` in stringified response; gets neither |

## Severity
MEDIUM — pre-existing test failures masking the production contract for MFA setup, verification, and management. Does not affect production callers (the issuer name, backup-code count, and response shape are documented in production code).

## Acceptance criteria
- [ ] Per-test triage: classify each as test-bug (assertion wrong) or production-bug (response shape genuinely missing fields)
- [ ] Either align tests to production reality OR fix production to match documented expectations
[...]

PR #2035, opened 2026-08-10
fix(auth): remove forged-auth stubs from production auth paths (#2026)

### LINKED ISSUE #2026: HIGH: production auth paths contain test stubs that forge tokens and log MFA codes

esperie:
## Summary

Issue #2026 named two test-compatibility shortcuts in shipped auth paths. Both are confirmed and fixed. The mandated sibling sweep of `src/kailash/nodes/auth/`, plus **four adversarial review rounds**, found many more of the same shape — several strictly more severe than the two reported, including two credential paths that authenticated on the *shape* of a credential rather than on any issued secret.

Every fix fails closed. Where a capability genuinely cannot be performed (no SMS transport, no push provider, no API-key store, no JWT key), the code raises or refuses instead of reporting success.

**Four review rounds, each of which found a CRITICAL the previous round missed** — including one vulnerability *introduced by an earlier round's own fix*, and one round-2 fix that a later round proved fully bypassable. Every finding is fixed or explicitly carried to #2047; the round-by-round record is below.

## The two issues as filed

**1. SSO token exchange forged a token** — `_exchange_oauth_code` returned a synthetic successful token response when `"oauth.example.com" in token_url`. The guard was a substring test, so `https://oauth.example.com.attacker.tld/token` matched, and it sat on the *exception* path, so the attacker wanted the call to fail. Branch deleted; a failed exchange now raises for every URL.

**2. MFA SMS logged the OTP and sent nothing** — `_send_sms` logged the full message body (which carries the code) at INFO and returned `True`. It now logs only `phone_last4` and body length, and raises `MFADeliveryError` rather than claiming a delivery it did not perform.

The issue's prescribed fixes were correct for both; no correction was needed.

## Sweep + review findings (all fixed)

**Forged / fabricated results**
| Site | Defect |
|---|---|
| `sso.py::_get_oauth_user_info` | Forged an identity (`test.user@example.com`) when the bearer token equalled `test_access_token`, on the exception path. |
| `sso.py::run` sync fallback | Dispatched to a "simplified synchronous implementation" whenever a loop was already running — every FastAPI/Nexus caller, not just tests. `action="validate"` returned `authenticated=True` with a caller-supplied `user_id` for **any** non-empty token. Replaced with a worker-thread offload running the real `async_run`. |
| `sso.py::_exchange_oauth_code` | Defaulted `token_endpoint` to `https://oauth.example.com/token`, POSTing the authorization code **and `client_secret`** to a host the operator does not control. |
| `directory_integration.py::_test_connection` | Reported `connected` with a full feature list when the server URL merely contained `test.`, and again when `ldap3` was simply not installed. |

**Credentials accepted without verification**
| Site | Defect |
|---|---|
| `enterprise_auth_provider.py::_authenticate_jwt` | Base64-decoded the payload with the signature **deliberately unverified** and trusted the token's own `sub`. Verified against the base code directly: a forged token returned `{'authenticated': True, 'user_id': 'admin'}`. |
| `enterprise_auth_provider.py::_authenticate_api_key` | Accepted **any** string ≥32 chars starting with `ak_` and derived the identity from the key's own text. `ak_` + filler + `_admin` authenticated as admin. |
| `mfa.py::_verify_totp_code` | Accepted the literal `123456` for any secret. |
| `mfa.py::_verify_sms_code` / `_verify_email_code` | Accepted **any** 6-digit string as a second factor. |
| `mfa.py::_verify_mfa` / `_verify_mfa_async` | Auto-enrolled any unenrolled user presenting `123456` against a hardcoded shared secret and returned a verified session. Removing `123456` from `_verify_totp_code` did **not** close this — these blocks never reach that helper. |
| `directory_integration.py::_fallback_directory_auth` | `valid_passwords.get(username, "password123")` authenticated **any** username with the default password, reachable whenever the real LDAP bind raised. |

**Authorization derived from identifier content**
| Site | Defect |
|---|---|
| `enterprise_auth_provider.py::_get_user_permissions` | Granted `["read","write","delete","admin"]` to any principal whose `user_id` merely **contained** `"admin"`. `admin.intern` was a full admin. |
| `enterprise_auth_provider.py::_determine_additional_factors` | Waived MFA when the `user_id` contained `test.`, `test_`, or `@company.com`. |
| `directory_integration.py::_fallback_directory_search` | Fabricated `memberOf` groups fed `_map_groups_to_permissions`, which grants admin for any group name containing `admin` — so making the directory unreachable provisioned an attacker as admin. |

**Identity substitution**
| Site | Defect |
|---|---|
| `enterprise_auth_provider.py::_authenticate` | Created the session for the **caller-supplied** `user_id`, discarding the principal the credential asserted. Anyone holding any valid credential could request `user_id="admin"`. |
| `_authenticate_jwt` / `_authenticate_api_key` / `_authenticate_social` / directory | A token with no `sub`, a store record with no `user_id`, a provider returning `user_id: None`, and the three directory returns (which carry `username`, not `user_id`) all fell back to the caller's claim. The assertion is now mandatory. |
| `enterprise_auth_provider.py::_authorize` | Skipped session validation entirely when `session_id` was falsy, treated an empty permission request as authorized, and read the session subject from a **top-level `user_id` that does not exist** (it is nested under `session_data`) — so validating the session never bound it to a principal. |

**Claimed-but-not-performed actions**
`mfa.py::_send_push_challenge` (POSTed to live FCM with a hardcoded `key=test_server_key`, ignored the response, reported sent on every path); `_initiate_recovery` (returned the recovery token to the caller while sending nothing); `_setup_email` (reported `verification_sent: True` with no transport).

**Result-contract**
Authorization **denials** were stamped `success: True` and audited as `auth_success`, because `success` defaulted to True when unset. All three nodes now derive `success` from the operation's own verdict. `_verify_backup_code` returned `success: True` for a *rejected* code.

**Availability / crashes**
`_revoke_mfa` took the non-reentrant `_data_lock` then called a helper that re-takes it — two calls permanently wedged **all MFA in the process**. Recovery delivery ran network I/O inside that global lock. `compare_digest` raised `TypeError` on a non-ASCII address; `_check_device_trust` deleted from the wrong store; malformed input raised out of `execute()`.

**Disclosure**
Full API key logged at INFO on every authentication; provider exceptions (which carry the recipient address/number) returned to the caller; unmasked email recipients; `api_key[:6]` exposed 3 characters of real secret.

**Transport security**
LDAPS certificate validation was `CERT_OPTIONAL` at one site, absent at another, and the primary auth path hardcoded the bind DN to `DC=test,DC=com` while dropping TLS handling entirely — so an operator setting `use_ssl` sent user passwords in cleartext. All three sites now share one `_build_tls_config` (`CERT_REQUIRED` by default).

**The escalation chain (rounds 3–4)**
Rounds 1–2 fixed leaf functions; round 3 found the root — `setup` was an unauthenticated, destructive, credential-issuing action, so every leaf guard was bypassed by walking one step upstream (`setup` → `generate_backup_codes` → `verify`). Round 4 then found `revoke`/`disable`/`trust_device` were the same shape. Now: re-enrolment over a **verified** factor, `reset`, `revoke`, `disable`, and admin recovery all require `admin_override`; backup codes are accepted only when a factor is genuinely verified; `_setup_push` no longer claims `verified: True` for a device that acknowledged nothing.

Round 4 also caught that round 3's `_setup_push` fix had made push **permanently incompletable** (nothing ever wrote a challenge status other than `pending`). Added the missing `approve_push`/`deny_push` device-response path, authenticated by a `challenge_token` that travels only in the push payload — never in the `send_push` return value.

### New configuration (all default to fail-closed)

`MultiFactorAuthNode(push_provider=...)`; `EnterpriseAuthProviderNode(jwt_config=..., api_key_store=..., user_permissions=..., default_permissions=...)`; `DirectoryIntegrationNode(connection_config={"simulate_directory": True, "allow_insecure_credential_fallback": True, "tls_validate_optional": True})`.

Additive — every one defaults to the refusing branch, so a deployment that relied on the permissive behaviour now fails visibly rather than silently authenticating.

## KNOWN LIMITATION — read before merging

**`MultiFactorAuthNode` has no actor, and `admin_override` is not an authentication control.** `user_id` names the *subject* of an action, never the caller; there is no session or principal input on any action. Several actions issue bearer credentials (TOTP secret, backup codes, device trust token) or destroy factors, so a caller that can pass an arbitrary `user_id` to them can take over or lock out any account. `admin_override` is an ordinary caller-supplied boolean — it separates administrative shapes from end-user ones for a host that already authorizes the caller, and nothing more.

This PR **documents** that plainly (a class-level warning + the parameter description) and gates the destructive actions behind it, but does **not** fix it: closing it properly means adding an authenticated-actor input and authorizing the `(actor, action, subject)` triple across ~15 actions and every consumer — a node API change well beyond this shard's budget. **Filed as #2047** with full analysis and acceptance criteria.

This is not a regression: it is the node's pre-existing design, previously undocumented at the API surface.

## Tests

`tests/regression/test_issue_2026_auth_test_stubs.py` — **114 behavioural cases**, `@pytest.mark.regression`. Every one calls the function and asserts on the raise/return; no source-grep assertions, which pass against a defect respelled. Positive controls throughout (a genuine TOTP code verifies; a properly signed JWT authenticates; an issued API key works; an issued email code verifies once and is single-use; the opt-in directory fallback works; push completes after a genuine device approval) so the fixes cannot pass by refusing everything.

**Fail-first verified**, `src/kailash/nodes/auth/` reverted to base `a101c81f9`, tests unchanged:

```
106 failed, 8 passed in 133.05s
```

The 8 passing are the deliberate positive controls. Each round's tests were additionally verified fail-first against the *immediately preceding* commit, so later fixes are proven to close something that was genuinely open. Where a test failed against base for the wrong reason (a constructor `TypeError` from a new config kwarg rather than the security assertion), the defect was confirmed separately by executing the base-commit logic directly:

```
OLD JWT  forged-token result: {'authenticated': True, 'user_id': 'admin', ...}
OLD APIKEY self-minted result: {'authenticated': True, 'user_id': 'admin', ...}
```

One test in an earlier revision asserted on a key the result never contained (`connected` instead of `connection_status`), so it passed both with and without the fix; it was caught, corrected, and re-verified bipolar.

## Deliberate deviations — please do not "fix" these upward

**API-key digest uses PBKDF2-HMAC-SHA256 at 10,000 iterations, not 100k+.** This is intentional and load-bearing, not a shortcut:

- The derivation runs **once per authentication**, to produce a dict lookup key — not once per stored entry in a scan. That is why a KDF is affordable here at all; the original design compared the presented key against every issued entry, where any slow hash would have cost one derivation per stored key per request.
- API keys are **high-entropy values this system generates**, not user-chosen passwords. The offline-guessing attack that motivates a 100k+ count does not apply. The pepper (`api_key_pepper`) is what defends a leaked `api_key_store` against precomputation; the iteration count is defence-in-depth on top of that.
- Raising it to password-grade would put a deliberately expensive derivation on every API request for no threat-model gain. Both knobs are configurable (`api_key_pepper`, `api_key_kdf_iterations`) if a deployment disagrees.

**The phone-number reference was deleted rather than hardened.** An earlier revision replaced the logged last-4 digits with a salted digest. That was reverted in favour of logging nothing derived from the subscriber number at all — `user_id` was already the correlation key that matters, so the digest bought nothing and still fed a derivation of PII into logs.

**`admin_override` is documented as NOT an authentication control.** It is an ordinary caller-supplied parameter. It separates administrative shapes from end-user ones for a host that already authorises the caller, and that is all it can do — see the class-level warning and #2047. It is deliberately not presented as a security boundary.

## Verification

```
pytest tests/regression tests/unit tests/integration/nodes/test_mfa_functional.py \
       tests/tier2_integration/test_todo028_directory_ldap.py
6739 passed, 18 skipped                            (exit 0)

pre-commit run --all-files
21 hooks Passed, 0 Failed                          (exit 0)
```

Note for reviewers: this worktree has no `.venv` and the shared venv's editable install points at the **main** checkout, so `pytest` must run from the worktree root (`pytest.ini` sets `pythonpath = src`, which wins over `PYTHONPATH`). Setting `PYTHONPATH` to compare against another source tree silently does nothing.

## Six pre-existing tests updated

Each asserted the now-closed behaviour; all keep their original intent:
- `test_backup_codes_format` (×2 files) — enrol before requesting codes.
- `test_backup_code_verification` — verify the TOTP factor before using a backup code.
- `test_mfa_reset_process`, `test_reset_action_clears_state_and_returns_new_setup` (#803) — pass `admin_override=True`; the #803 response contract itself is unchanged.
- `test_mfa_module_level_send_sms_helper_still_exists` — now expects `MFADeliveryError`.

## Pre-existing failures found, NOT introduced here

Confirmed pre-existing by re-running each with `src/kailash/nodes/auth/` reverted to base — byte-identical failures. None in `nodes/auth/`:

- `tests/tier_2/integration/test_nexus_authentication_defaults.py` — 2 failures, one self-described: `SECURITY BUG: NEXUS_ENV=prod, enable_auth=None -> Expected auth=True, got False`. **A real fail-open default in Nexus; deserves its own issue.**
- `tests/tier2_integration/mcp_server/test_oauth.py` (6 errors), `tests/e2e/test_mcp_resource_flow.py::test_auth_enforcement` (1 error), `tests/integration/mcp_server/…` (2), `test_gdpr_compliance_comprehensive.py::test_breach_notification_to_authorities`, `test_gateway_integration.py::test_concurrent_session_management`.

Also surfaced: `SessionManagementNode` has no `execute_async`, so `EnterpriseAuthProviderNode`'s `authenticate`/`authorize` error out before reaching their logic. Pre-existing; noted because it currently masks the identity-binding paths fixed here.

## Not fixed (reported, out of class)

- **#2047** — the actor-model gap above.
- `session_management.py::_get_ip_location` returns `{"country": "Unknown", ...}` — an honest degraded placeholder, not a forged auth result; a real fix needs a geo-IP dependency.
- `_determine_additional_factors` retains a pre-existing `hash(user_id) % 3 == 0` sampling rule. `hash()` on `str` is per-process randomized, so step-up varies across workers. Left deliberately — removing it *weakens* a control, and redesigning step-up policy is beyond this change.
- `recovery_requests` has no redemption path anywhere in `src/`, so a recovery token is issued and delivered but cannot be spent. Pre-existing; implementing `complete_recovery` belongs with #2047.
- `enterprise_auth_provider.py` risk-scoring helpers remain simulated — they feed an advisory score, not an auth decision.
- `auth_method="mfa"` as a *primary* method cannot authenticate (the MFA result has no `authenticated` key). Pre-existing, fails closed.

## Review requirement

Per `agents.md` § "Correctness-Review-Clean Is Not Security-Clean": four adversarial security rounds have run (findings above). A correctness reviewer should still pass over the six updated pre-existing tests and the new `approve_push` path.

## Related issues

Fixes #2026
Refs #2047


esperie:
## CodeQL status

The CodeQL check is red on **one** alert: `py/weak-sensitive-data-hashing` at `packages/kailash-kaizen/src/kaizen/memory/vector.py:76` (MD5).

It is **pre-existing and untouched by this branch**:
- present verbatim at the base commit: `git show a101c81f9:packages/kailash-kaizen/src/kaizen/memory/vector.py` → `hash_val = int(hashlib.md5(text.encode()).hexdigest(), 16)`
- this branch changes **zero** files under `packages/kailash-kaizen` (`git diff a101c81f9..HEAD --name-only | grep -c kaizen` → `0`)

The four alerts this branch *did* introduce are all fixed, not dismissed:

| Alert | Resolution |
|---|---|
| `py/clear-text-logging-sensitive-data` × 4 (mfa.py) | Delivery logs no longer reference the destination number at all — not even its last 4 digits. They report body length and identify the record by `user_id`. |
| `py/weak-sensitive-data-hashing` (enterprise_auth_provider.py) | API-key lookup digest is now PBKDF2-HMAC-SHA256 salted with `api_key_pepper`, affordable because the derivation runs once per authentication to produce a lookup key rather than once per stored entry. |
| `py/empty-except` (mfa.py) | The expired-device cleanup's bare `except` now logs why the exception is benign. |
| `py/unused-import` (enterprise_auth_provider.py) | `hmac` import removed after the PBKDF2 switch. |

Per `zero-tolerance.md` Rule 1a I am flagging rather than silently dismissing the kaizen MD5: it is a genuine finding, it is simply not in this change's blast radius and belongs to `packages/kailash-kaizen`.

esperie:
Merge disposition — **the two remaining reds are both explained and neither is this branch's.**

**CodeQL** — one alert: `py/weak-sensitive-data-hashing` at `packages/kailash-kaizen/src/kaizen/memory/vector.py:76`. Confirmed pre-existing by two independent checks: it is present on `main` at the **same line**, and `gh pr diff 2035 --name-only | grep -c "^packages/"` returns **0** — this branch touches no file under `packages/` at all, so it cannot have introduced it. Explained, not dismissed.

**build = cancelled** — the push-event Build Documentation run cancelled by its own PR-event run, which passed.

## Verified on the pinned head `580777bea`

Conclusion breakdown: 14 success, 4 skipped, 1 failure (CodeQL), 1 cancelled (build). **The 4 skipped rows are the duplicate workflow's placeholders** — one of them is literally `Test (Python ${{ matrix.python-version }})`, an unexpanded matrix template. The genuine jobs all ran: Python 3.11/3.12/3.13/3.14, PACT, DataFlow Unit, and DataFlow Infra Regression all **success**.

**That distinction is worth recording, because it nearly produced a false green.** A `CONFLICTING` PR cannot have its merge ref built, so GitHub reports the entire matrix as `skipped` — and a checks summary that counts skipped as "not failing" reads that as green. **Skipped is not passed.** This is the same class as #2038 (Tier-2 `continue-on-error` making a red invisible): a check surface returning the same answer whether or not the work happened. Two rebases were needed here — the first landed clean, then main moved and the PR went conflicted, which is *why* no `pull_request` workflow had run.

## Process

No commit authored by this lane carries a bypass any more. The two `--no-verify` commits were soft-reset and re-committed under the full suite: `pre-commit run --all-files` exit 0 with `Run Tier 1 unit tests ... Passed` on the first pass, and pytest on the PR selection at **6851 passed, 18 skipped**.

One residue, and it is **mine, not the lane's**: the predecessor WIP commit still documents its own `--no-verify`. That commit was made by the orchestrator to preserve work when an account quota killed the original lane mid-task. Rewording it needs an interactive rebase across 15 commits; the lane correctly flagged it rather than risking the history to fix someone else's message.

## Bonus finding, from driving a branch no test had executed

`_log_security_event` called `security_logger.async_run`, which `SecurityEventNode` does not define — so **every successful async SSO operation raised `AttributeError` after doing its work**, making the async surface unusable and meaning no SSO security event was ever recorded. Found only because the sync-bridge tests actually drove that path. Fixed here.

PR #2036, opened 2026-08-10
fix: redact sensitive values at log, cache-key, and file-mode sinks (#2027)

### LINKED ISSUE #2027: HIGH/MED: sensitive values reach logs, cache keys, and a world-readable audit log (5 sites)

esperie:
Fixes #2027

## Measured site count: **12 real**, versus the issue's claimed **5**

I built an independent enumeration instrument rather than inheriting the issue's grep. Two passes:

1. **CodeQL ground truth** — all open alerts in the three classes the issue draws from (`py/clear-text-logging-sensitive-data`, `py/weak-sensitive-data-hashing`, `py/overly-permissive-file`): **26 alerts**, of which **6 are real**. The other 20 log only names, counts, PKs, SQL statement text, or are public-key writes (verified by reading each, not by inheriting the issue's dismissal).
2. **A taint-aware AST sweep** (`logger.*` calls whose interpolated arguments reach a sensitive identifier, including `%`-style lazy logging with args on following lines, and one level of indirection through an intermediate variable). This found **6 more that CodeQL does not model**.

I calibrated both instruments against the pre-fix source first: a grep that returns empty after a fix is only evidence if it returned non-empty before. The first grep I wrote was **blind** to the nexus `%s`-style site, which is why it was replaced.

### The 12

| # | Site | Source |
|---|------|--------|
| 1-4 | `gateway/resource_resolver.py` — MD5-over-secret registry keys at **4** call sites (db pool, HTTP client, message queue, S3) | issue named only S3 |
| 5 | `runtime/resource_manager.py:97` — same MD5-over-config idiom, and the derived id reaches logs | sweep only |
| 6 | `runtime/parameter_injector.py:411` — workflow parameter **value** at DEBUG | issue #2 |
| 7 | `dataflow/core/nodes.py:1034` — raw pre-sanitization value at DEBUG | issue #3 |
| 8 | `dataflow/core/nodes.py:1030` — schema field name at WARNING | issue called this a false positive |
| 9 | `dataflow/core/nodes.py:1924` — `repr(value)[:50]` of **every** bound CREATE field at DEBUG | sweep only |
| 10 | `dataflow/core/nodes.py:1661` — the entire kwargs dict, unmasked | sweep only |
| 11 | `dataflow/core/logging_config.py` — the shared masker was a **no-op on dict/JSON shape** | sweep only |
| 12 | `nexus/auth/rate_limit/middleware.py:184` — raw identifier at WARNING | issue #4 |

Plus the two file-mode fixes: `trust/plane/integration/cursor/hook.py` (issue #5) and `gateway/security.py`.

### The masker finding (#11) is the most serious

`mask_sensitive_values` has a `token[=:\s]+...` pattern, but in a mapping repr the key's own closing quote sits between the name and the separator, so `{'token': 'sk-live-...'}` matched **nothing**. That is precisely the input `mask_sensitive_values(str(kwargs))` is handed at `nodes.py:1713`. Any credential passed as an ordinary DataFlow model field reached DEBUG in clear text, through the very call that looked like it was protecting it. Verified before/after:

```
before: {'id': 't-1', 'token': 'sk-live-SECRET', 'password': 'hunter2'}
after:  {'id': 't-1', 'token': '***MASKED***', 'password': '***MASKED***'}
```

## On the cache key: the issue's proposed fix would have introduced an auth bypass

The issue's acceptance criterion asks to prove the secret is **out** of the key material ("stable across two calls with different secrets but identical connection identity"). **I did not implement that, deliberately.**

If the credential is not part of the cache identity, then two callers with the same host/database/user but *different passwords* collide — and the second caller is handed a live pool **already authenticated with the first caller's credential**. A caller supplying a wrong or revoked password would receive a working authenticated connection. Rotating a password would likewise silently reuse the pool built from the old one.

So the secret **stays** in the pre-image, and the defect (MD5 is an unkeyed, offline-brute-forceable oracle, truncated to 32 bits) is fixed instead by `process_local_config_key`: **keyed** BLAKE2b with per-process random material, widened to 64 bits. Knowing the entire plaintext config does not let someone who saw the key confirm a guessed password.

The 32-bit truncation was independently a correctness bug: 8 hex chars collide at roughly 2**16 configs, and a collision returns another tenant's authenticated connection.

This is one shared helper in `utils/url_credentials.py` (alongside the existing `mask_url` / `mask_secret` / `fingerprint_secret`), per `security.md` § Credential Decode Helpers — all five call sites route through it; no per-site hand-rolled scrubbing.

## `dataflow/core/nodes.py` diff scope (issue #2006 is blocked on this file)

Kept deliberately tight. **Symbols touched, and nothing else:**

- `NodeGenerator._create_node_class` → `validate_inputs` → `sanitize_sql_input` closure — the two injection log lines (sites 7, 8)
- `NodeGenerator._create_node_class` → `async_run` — the `value_debug` parameter trace (site 9) and the `received kwargs` line (site 10)
- one import: `fingerprint_secret` from `kailash.utils.url_credentials`

No signatures, no control flow, no SQL generation touched.

## Verification

Every regression test was **fail-first verified per file**: the source file was reverted to its pre-fix state one at a time and the suite re-run, so each red is attributable to the specific fix rather than to a collection error.

That process caught two tests that were passing for the wrong reason:

- One asserted the raw value was absent from `caplog`, but DataFlow sets a level on `dataflow.core.nodes` itself, so raising only the root logger meant the line under test **never fired** — it passed identically against the unfixed source. Fixed to set the level on that logger and to assert the record was emitted before checking its content.
- The nexus test, placed in the root suite, resolved `nexus` from the pip-installed package rather than the tree under test (root `pytest.ini` puts only core's `src` on the path). Moved into the nexus package's own tree.

```
tests/regression/test_issue_2027_sensitive_value_sinks.py ........ 17 passed
packages/kailash-dataflow/.../test_issue_2027_dataflow_log_sinks.py 7 passed
packages/kailash-nexus/.../test_issue_2027_identifier_logging.py    1 passed

pre-commit run --all-files ......................................... EXIT=0
packages/kailash-dataflow  tests/unit .... 3677 passed, 17 skipped
packages/kailash-nexus     tests/unit .... 1817 passed,  1 skipped
```

## Deliberately not fixed (reported, not silently dropped)

- **`nodes.py:1093`** logs `field_name` at WARNING with only `len(value)` — no value leaks. The `ValueError` two lines above emits `field_name` plainly **by mandate** of `security.md` § Sanitizer Contract, so redacting only the warning would be incoherent.
- **`aiosqlite` logs its own bound parameters at DEBUG**, independently of DataFlow. Third-party driver behaviour on a logger operators silence separately; the CREATE test scopes its assertion to `dataflow.*` records rather than reporting on a sink DataFlow does not own.
- **Rate-limit config plumbing**: `RateLimitConfig(requests_per_minute=1)` is echoed in `X-RateLimit-Limit: 1` but the backend enforces ~21. A real inconsistency, unrelated to this issue — flagged for separate triage.
- **5 kaizen-package alerts** (`base_agent`, `presets`, `url_safety`, `memory/vector`, `client_cache`) — other lanes own those files.

`gateway/security.py` replaces write-then-`chmod` with `os.open(..., 0o600)` + `O_NOFOLLOW` + `fchmod`, closing the window where the secret file is briefly world-readable. The `f.write(secret)` plaintext-vs-documented-encryption defect is untouched and remains #2024's scope.


esperie:
## Windows blocker fixed; CodeQL triaged (3 artifact, 1 real — fixed)

### The crash

`os.fchmod` is POSIX-only. I guarded `O_NOFOLLOW` with `hasattr` but not `fchmod`, so every hardened write raised `AttributeError` on Windows. Worse, the audit-log writer catches only `OSError`, so it did not degrade to best-effort — the `AttributeError` escaped as a crash.

Reproduced locally by deleting the attribute (the exact Windows condition, not an approximation):

```
pushed HEAD + no os.fchmod  -> AttributeError: module 'os' has no attribute 'fchmod'   # CI's message verbatim
fixed code  + no os.fchmod  -> audit log wrote OK: BLOCKED / secret wrote OK
                               hasattr(os,'fchmod') during run: False                  # mutation was live
```

### Not a reflex `hasattr` wrap

POSIX mode bits do not restrict *readers* on Windows — `os.chmod` there only toggles the read-only attribute — so `0o600` confidentiality is genuinely POSIX-only. I took option (a)+(b) combined, via one helper, `kailash/utils/file_permissions.py::restrict_to_owner(path, fd=...)`:

- **POSIX** — `os.fchmod` on the descriptor (so a symlink swapped in after the open cannot redirect it), falling back to `os.chmod` when no fd is available.
- **Windows** — sets a DACL restricting the file to the current user's SID via pywin32; without pywin32 there is no mechanism at all, so it emits a **one-time WARNING naming the unprotected file** and tells the caller.
- **Returns `bool`**, not `None`: whether protection was *actually* applied. A protection that silently does not apply is exactly the failure mode this whole issue is about (`security.md` § Secure-Default), so the code says so at both call sites and in the helper docstring rather than leaving a reader to infer it from a `0o600` literal that has no effect there.

`trust.plane.project.set_private_file_permissions` already carried a near-identical Windows DACL implementation. Two copies of this logic is the drift `security.md` § Credential Decode Helpers exists to prevent, so it is now a delegation to the same helper.

I swept the whole branch diff for other POSIX-only calls: `fchmod` (2 sites) was the only one. `O_NOFOLLOW` was already guarded; `os.open`/`close`/`fdopen`/`chmod`/`stat` are cross-platform.

### CodeQL: 3 of 4 are the location-keyed baseline artifact, 1 was real

The check reported *"4 new alerts including 3 high severity"*. The three high-severity ones are the **same pre-existing alerts**, re-keyed because my diff shifted their lines:

| Baseline on main | PR annotation | Δ |
|---|---|---|
| #47 `nodes.py:1030` | `nodes.py:1039` | +9 |
| #48 `nodes.py:1034` | `nodes.py:1044` | +10 |
| #11245 `middleware.py:183` | `middleware.py:188` | +5 |

The deltas match my insertions exactly, and the code now *at* those lines is the fixed version — `fp={field_fp}`, `(value redacted)`, `fingerprint_secret(str(identifier))`. The raw values are gone.

**Honest caveat:** these alerts will persist after merge. CodeQL does not model `fingerprint_secret` as a sanitizer, so the syntactic taint path from `field_name`/`identifier` to the log call still exists. They are not introduced by this PR and are not closed by it either, in CodeQL's view. I did **not** add suppression comments — silencing a taint rule repo-wide is a policy call, not mine to make unilaterally, and it would mask future real findings at these sites.

The fourth was **genuinely mine and is fixed**: `logging_config.py` — implicit adjacent-string concatenation inside a list literal, which is indistinguishable from a missing comma to both a reader and `py/implicit-concatenation-in-list`. The pattern is now a named constant built with explicit `+`, outside the list. Behaviour verified identical before/after.

### Gate

```
tests/regression/... + tests/trust/plane/integration/test_cursor_integration.py   89 passed
tests/trust (full)                                    6328 passed, 33 skipped
packages/kailash-dataflow tests/unit                  3677 passed, 17 skipped
pre-commit run --all-files                            EXIT=0  (and idempotent on re-run)
```

POSIX-only mode assertions are now marked `posix_only`; the cross-platform contract is asserted by tests that run on every platform, including one that deletes `os.fchmod` to pin this regression, and one that forces the Windows-without-pywin32 path and asserts it reports `False` and warns exactly once.

esperie:
## CI green except CodeQL, which is now purely the line-shift artifact

**All 25 non-CodeQL checks pass, including all four Windows jobs** (3.11 / 3.12 / 3.13 / 3.14) that were the blocker.

### The two alerts in my new file were real — both fixed, not argued away

The previous push introduced two alerts in `utils/file_permissions.py`. I owned those rather than dismissing them, and in both cases the fix is right independent of the scanner:

- **`py/clear-text-logging-sensitive-data`** on the interpolated `path`. Dropped it. The warning fires **once per process** about a *missing platform mechanism*, so naming one arbitrary path implied only that file was affected — actively misleading. Callers needing per-file attribution already have the `False` return value. The message also no longer calls the file "secret", since the helper is general-purpose and the caller may be protecting an audit log rather than a key.
- **`py/unused-global-variable`** on the `_WARNED_NO_ACL` flag. Replaced with an `lru_cache(maxsize=1)` latch: no `global` statement, and tests get a documented reset (`cache_clear()`) instead of poking a module attribute. Verified the latch still discriminates — 3 calls emit **1** warning through the cached function and **3** through `__wrapped__`.

Annotations went **5 → 3**.

### The remaining 3 are the baseline artifact, with evidence

| Baseline on main | PR annotation | Δ |
|---|---|---|
| #47 `nodes.py:1030` | `nodes.py:1039` | +9 |
| #48 `nodes.py:1034` | `nodes.py:1044` | +10 |
| #11245 `middleware.py:183` | `middleware.py:188` | +5 |

Same rule, same three sites, shifted by exactly the number of lines my diff inserts above each. The code now *at* those lines is the fixed version — `fp={field_fp}`, `(value redacted)`, `fingerprint_secret(str(identifier))`.

**These will persist after merge.** CodeQL does not model `fingerprint_secret` as a sanitizer, so the syntactic taint path from `field_name` / `identifier` to the log call still exists even though the values are now one-way hashed. They are neither introduced nor closed by this PR. I deliberately did **not** add suppression comments — silencing a taint rule at these sites is a policy call, and it would mask any genuinely new finding there later.

So the CodeQL red is not a defect in this branch, but it is also not something this branch can clear. That needs a decision from you: merge over it (the three are pre-existing on main today), or open a separate triage to either suppress-with-justification or teach the query about the fingerprint helper.

esperie:
CodeQL disposition — **merging over the three remaining alerts, with evidence, and NOT suppressing them.** Tracking issue: #2044.

The three `py/clear-text-logging-sensitive-data` HIGH alerts are the location-keyed artifact, verified against both refs:

| Alert | main | this PR | delta |
| --- | --- | --- | --- |
| #47 | `dataflow/core/nodes.py:1030` | `:1039` | +9 |
| #48 | `dataflow/core/nodes.py:1034` | `:1044` | +10 |
| #11245 | `nexus/auth/rate_limit/middleware.py:183` | `:188` | +5 |

All three exist on `main` today at the original lines; the deltas match this PR's insertions exactly; and the code at the new lines is the **fixed** version. Neither introduced nor closed here — CodeQL simply does not model `fingerprint_secret` as a sanitizer, so the syntactic taint path survives one-way hashing.

Declining to add suppression comments was the right call and is upheld: silencing a taint rule at these sites would mask future real findings, and that is a policy decision. #2044 records the three options (teach the query / dismiss per-alert with evidence / accept), with a recommendation and its honest downside.

## Independently verified before merge

- **Windows fix, by mutation** — deleted `os.fchmod` (the exact Windows condition) and re-ran the write path: completes with no `AttributeError`. On POSIX-minus-`fchmod` it takes the documented `chmod` fallback and returns `True`; the `False` return is the Windows-without-pywin32 case, where there is genuinely no mechanism and the caller is told so rather than reading a `0o600` literal and inferring protection.
- **All 25 non-CodeQL checks pass, including all four Windows jobs.**
- **The masker's residual gaps are closed too** — re-ran the probe: `Node(token=...)` repr form, `api_token`, `client_secret`, and `x-api-key` are now all masked, alongside the dict and URL forms. That went beyond what was asked (a follow-up would have been acceptable).

One note for the record: a `QUOTED_KEY_MAPPING_PATTERN` "undefined" editor diagnostic on this branch is spurious — the real constant is `QUOTED_KEY_MAPPING_PATTERNS` (plural), defined and used correctly. I initially reproduced the diagnostic's singular spelling in my own check and got a misleading `False`; the module imports and every masking path works.

Issue #2041, opened 2026-08-10
CRITICAL: SecretManager generates an ephemeral encryption key and continues at WARNING — every secret is unrecoverable after restart


esperie:
Found while investigating #2024. **This sits on the shipped default path and is arguably higher-impact than #2024 itself**, because unlike `FileSecretBackend` (which has zero call sites) this one is constructed by default in two servers.

## The defect

`src/kailash/gateway/security.py::SecretManager.__init__`:

```python
        if encryption_key:
            self._cipher = Fernet(encryption_key.encode())
        else:
            # Generate key from environment or use default
            key = os.environ.get("KAILASH_ENCRYPTION_KEY")
            if not key:
                # Warning: This is not secure for production!
                logger.warning(
                    "Using default encryption key - not secure for production!"
                )
                # Generate a proper Fernet key
                key = Fernet.generate_key()
            elif isinstance(key, str):
                key = key.encode()
            self._cipher = Fernet(key)
```

With `KAILASH_ENCRYPTION_KEY` unset, it **generates a fresh random key per instance and continues**. Every secret encrypted by that process is unrecoverable the moment the process restarts. The comment even says "not secure for production" — and then proceeds.

## Why this is the default path, not a corner case

`SecretManager()` is the default construction in both:

- `src/kailash/gateway/enhanced_gateway.py` — `secret_manager or SecretManager()`
- `src/kailash/servers/enterprise_workflow_server.py` — same shape

So an operator who never sets the env var gets silent, self-destructing encryption.

## Three compounding problems

1. **Fail-open default.** `security.md` § Secure-Default requires a new security control to fail CLOSED, or emit a loud one-time WARN naming the OFF protection and its wiring. A `logger.warning` at init inside a library is not loud enough to survive a production log pipeline, and the code proceeds as if encryption were configured.
2. **No key derivation.** `Fernet(encryption_key.encode())` requires a 32-byte urlsafe-base64 key. Any human-chosen passphrase raises inside Fernet. No KDF, no salt — so the documented-looking `encryption_key` parameter is unusable with anything an operator would naturally supply.
3. **`KAILASH_ENCRYPTION_KEY` is undocumented.** It appears exactly once in the repo — its own read site. No `.env.example` entry, no docs, no test. An operator cannot discover that the variable exists, which is precisely how the fail-open branch becomes the normal branch.

## The repo already has the right pattern

[...]

PR #2048, opened 2026-08-10
fix(gateway): FileSecretBackend encrypts at rest instead of claiming to (#2024)

### LINKED ISSUE #2024: CRITICAL: FileSecretBackend stores secrets in cleartext while documenting encrypted storage

esperie:
## Summary

`FileSecretBackend` documented "encrypted file storage" and wrote cleartext. `store_secret` wrote the secret verbatim (`f.write(secret)` for strings, `json.dump` otherwise) and `get_secret` was the symmetric plain `json.load` — no decrypt step existed anywhere in the class, so the claim was never true on the direct-instantiation path (`zero-tolerance.md` Rule 2: fake encryption is enumerated BLOCKED).

Secrets are now stored as a versioned envelope holding a Fernet token keyed by PBKDF2-HMAC-SHA256, reusing the KDF shape of `trust/security.py::SecureKeyStorage`. `SecretManager`'s cipher was deliberately NOT reused: it generates an **ephemeral** key and continues at WARNING when `KAILASH_ENCRYPTION_KEY` is unset (#2041), so routing through it would inherit a store nothing can decrypt after a restart.

Three adjacent defects in the same class close with it: path traversal on read/write/**delete**, read-side symlink TOCTOU (PR #2036 hardened only the write path), and a world-listable `0o755` secrets directory.

## Design decisions

- **Salt is per-secret and persisted in the envelope**, not per-instance — the store survives restarts and no two files share a derived key.
- **Fails closed at construction** when no master key is available (`KAILASH_SECRETS_MASTER_KEY`, or `master_key=`). Inventing a key gives either an unrecoverable store or an unprotected one (`security.md` § Secure-Default).
- **No plaintext fallback.** A legacy cleartext file raises `LegacyPlaintextSecretError`. A fallback would let anyone able to write into the secrets directory swap an envelope for a file of their choosing and have it honoured — a downgrade oracle.
- **KDF parameters are validated on read** (iteration floor/ceiling, KDF name, envelope version, salt length). The count comes back from the file; `iterations: 1` would turn every read into a cheap key-guessing oracle.
- **Directory mode routes through a new `file_permissions.restrict_dir_to_owner`** rather than a call-site `chmod`, per that module's stated single-source-of-truth contract — the Windows answer is a DACL, and `0o600` on a directory locks the owner out.

## BREAKING

Construction raises without a master key, and existing plaintext stores stop being readable. Those secrets have been sitting in clear on disk: the CHANGELOG migration note says **rotate**, not migrate. In-repo blast radius is zero (no construction site anywhere; not in `gateway/__init__.py::__all__`; not the default backend).

## Tests

51 behavioural tests in `tests/integration/gateway/test_issue_2024_file_secret_encryption.py`. Byte-absence assertions read the raw file back outside the class; round-trip tests in the same file mean they cannot be satisfied by writing garbage.

**46 of 51 RED against the pre-fix implementation** (verified against a pristine `git archive HEAD~1` tree with a two-class import shim so collection could proceed; store/read behaviour untouched). The 5 that pass pre-fix are guard tests pinning behaviour that already held — named in the commit body, not offered as evidence for this fix.

Four `#2027` tests that constructed `FileSecretBackend` are updated, not deleted: they are about file *mode*, so they supply a master key; two of them asserted the password was readable straight out of the file, which is now asserting the bug, so that assertion is inverted to require absence.

## Gates

```
2158 passed, 12 skipped, 50 warnings in 69.37s     # tests/integration/gateway/ tests/regression/ tests/unit/utils
pre-commit run --all-files                          # EXIT=0, all 21 hooks Passed
```

## Related issues

Fixes #2024
Refs #2041 (SecretManager's own key handling — deliberately not routed through)
Refs #2036 (write-path hardening this builds on)


esperie:
## Review round 2 — all four addressed (commit `1821779`)

**HIGH — envelope substitution / rollback.** Confirmed and fixed. The plaintext is now `{"ref": reference, "secret": value}` and `get_secret` refuses an envelope whose embedded `ref` is not the reference requested. The binding lives *under* the ciphertext, so it cannot be edited independently; a payload arriving *without* a `ref` is rejected rather than accepted unbound — the permissive branch would have preserved the whole defect.

Your exact probe, re-run against `lane-1b/src` (resolution printed):

```
resolved: /Users/esperie/repos/kailash/build/.kailash-py-wt/lane-1b/src/kailash/gateway/security.py
SUBSTITUTION ACCEPTED: False -> SecretEncryptionError: Secret db_master holds an
envelope written for a different reference. The file has been substituted or restored over.
```

Regression tests added for the substitution case specifically: file-copy substitution, restored-pre-rotation-backup, and file rename — plus a control asserting the legitimate read still works, so they cannot pass against a `get_secret` that raises unconditionally.

**MEDIUM — master-key floor.** `MIN_MASTER_KEY_LENGTH = 32`, with a test pinning it equal to `JWTConfig.MIN_SECRET_LENGTH` so the two cannot drift. The error names the floor and the received length; a test asserts the rejected key is *not* echoed, since an exception message can reach a log or crash report.

**`restrict_to_owner` return value.** Now checked on the file write and logs a per-file finding naming the file and the remedy — the helper's once-per-process warning may already have been consumed by an unrelated write.

**Atomic write.** Temp file, `fsync`, `os.replace`. Deliberately *not* `trust._locking.atomic_write`, which is otherwise the same shape: it has no place to apply `restrict_to_owner` to the descriptor, and this call site must both apply it and act on its answer. `os.replace` also replaces a symlink at the target rather than writing through it.

**Rollback detection** was considered and not added: a monotonic counter has nowhere tamper-proof to live in a per-file store (the counter file is as substitutable as the envelope). The `ref` binding closes cross-secret substitution; same-reference rollback needs a store-level ledger, which is a different design. Flagging rather than silently dropping.

**Gates**

```
2171 passed, 12 skipped, 51 warnings in 56.46s   # integration/gateway + regression + unit/utils
PRECOMMIT_EXIT=0   (21 hooks Passed, no failures)
```

9 of the 13 new tests RED against the pre-review commit; the 4 that pass are the controls named above.

esperie:
## Round 3 — rollback closed (`137aae5`), CodeQL triaged alert-by-alert (`9c152de`)

### Rollback — the reported repro now raises

```
after rotation : ROTATED
ROLLBACK -> REJECTED: SecretRollbackError: Secret cred is version 1 but
version 2 was stored. This file is an earlier generation — most likely a
restored backup that has undone a rotation.
```

Each store increments a `version` inside the encrypted payload; the high-water mark lives in a **separate authenticated manifest** (`.secret-versions.json`), itself an envelope bound to a reserved ref — so an attacker who can write in the directory cannot forge a lower mark, only delete or restore one. The mark is additionally floored at the version already on disk, so a lost bump costs one increment instead of resetting the counter.

I took **option 1 + 2 as one mechanism**: the manifest defaults to inside the store (closes the file-level restore you reproduced) and moves outside it via `state_dir=` / `KAILASH_SECRETS_STATE_DIR` (closes the whole-directory restore). Option 3 (`max_age=`) is also there, opt-in, because it is the only one of the three that survives losing the manifest.

### The guarantee, stated honestly

The docstring carries this table; two tests pin the **NO** rows, so if someone later closes one the test fails and tells them to correct the doc.

| Scenario | Detected |
|---|---|
| Restore one or more secret files | YES |
| Tamper with the manifest to lower a mark | YES (authenticated) |
| Restore the whole directory | **NO** by default; `state_dir` outside the store closes it |
| **Delete** the manifest, then restore an old file | **NO** — the marks are the only record |

I had first written that the manifest-deletion row was detected. It is not: once the marks are gone, a restored old envelope is indistinguishable from a never-rotated one. Corrected before it shipped rather than after.

### CodeQL

The master-key-floor alert you tested is now `fixed` (the line moved). Four new ones appeared on the merge ref and I triaged each rather than applying your evidence to all:

- **3 × `py/clear-text-logging-sensitive-data`** (`security.py:421,441,767`) — dismissed as false positives. Each logs a filesystem *path* (`secrets_dir` / `file_path`), never a value. Verified independently by execution with a canary key: absent from every log record and exception, no 8-char substring either. Suppressing the path would make an operator-facing warning useless.
- **1 × `py/empty-except`** (`security.py:780`) — **a real finding, fixed in code, not dismissed.** A temp file left behind after a failed write holds a real encrypted secret and nothing else cleans it up; the operator is now told which file to remove.

### Gates

```
2206 passed, 12 skipped, 77 warnings in 63.84s
PRECOMMIT_EXIT=0   (21 hooks, no failures)
```

12 of the 14 new rollback tests RED against the pre-fix head; the 2 that pass are the documented-limit tests, which assert the rollback *is* accepted.

**Note:** `origin/main` is merged in (`#2049` added `kailash/utils/command_safety.py`, which the shared venv's `kailash-mcp` imports — the Tier-1 hook could not pass without it). One commit used `-c core.hooksPath=/dev/null`, documented in its body per `rules/git.md`: the hook could not pass until main was merged, and the merge could not run with a dirty tree. The full hook set passes on the merged tree.

esperie:
## Round 4 — the weaker default now announces itself (`7a8cf69`)

Took **option 2**, with a correction: the constructor already emitted that message, but at `logger.info` — invisible to any handler configured at WARNING, which is most of them. Clean before/after discrimination:

```
BEFORE: records=1  at WARNING+=0  levels=['INFO']
AFTER:  records=1  at WARNING+=1  levels=['WARNING']
```

The message names what is **not** protected (whole-directory restore), what still **is** (single-file restore — so it is not misread as "rollback detection does not work"), and the exact wiring (`KAILASH_SECRETS_STATE_DIR` / `state_dir=`). Fires once per *directory*, not per construction: it comes from `__init__`, and a service building a backend per request would otherwise flood the log with a static configuration fact — but a second misconfigured store still gets its own warning. The class docstring now leads with the same statement rather than leaving it to the table.

### Why not option 1

`security.md` § Secure-Default allows the loud-WARN path only when on-by-default is genuinely infeasible rather than merely inconvenient, so this is the argument, not an assumption: **no location this class can pick is *known* to sit outside the operator's backup boundary.** A sibling directory looks separate but is captured by any backup taken one level up. An automatic default would swap an accurate warning for a silent and possibly false claim of protection. Only the operator knows where their backup boundary runs, which is precisely what the warning asks them to supply.

### Your three-case probe, re-run at this head

```
single-file rollback, default config           -> REJECTED  SecretRollbackError
whole-directory restore, state_dir DEFAULT     -> ACCEPTED  'ORIGINAL'   (documented + WARNED)
whole-directory restore, state_dir OUTSIDE     -> REJECTED  SecretRollbackError
```

### Whole-directory regression tests

Both configurations were already pinned (`test_a_whole_directory_restore_is_undetected_by_default`, `test_state_dir_outside_the_store_closes_that_case`) — the first asserts the ACCEPTED outcome deliberately, so if someone later closes it the test fails and tells them to correct the docstring. Four new tests cover the warning itself: fires in the degraded default, silent when `state_dir` is external, once per directory, twice for two directories.

### CodeQL

No open alert at `security.py:318` — that instance is `fixed` (the line moved), and the three path-logging alerts on the current head are dismissed with the canary-key evidence. `py/empty-except` (11480) still reads `open` but is stale: I fixed it in `9c152de` rather than dismissing it, since a temp file left after a failed write holds a real encrypted secret. It should close on the next analysis; I will confirm.

### Gates

```
2210 passed, 12 skipped, 50 warnings in 62.78s
PRECOMMIT_EXIT=0   (21 hooks, no failures)
```

esperie:
CodeQL disposition — **three alerts, all false positives on log lines carrying paths rather than secrets.** Merging over them with evidence rather than suppressing.

| alert | site | verdict |
| --- | --- | --- |
| `py/clear-text-logging-sensitive-data` | `security.py:306` | the degraded-rollback WARN. Interpolates `secrets_dir` — a **directory path**, not credential material. |
| `py/clear-text-logging-sensitive-data` | `security.py:832` | partial-file cleanup warning. Interpolates `tmp_path` + `cleanup_error` — a path and an OS error. |
| `py/unnecessary-pass` | `security.py:765` | style only. |

CodeQL's taint model flags any use of a variable in a secrets-handling module; none of these carry key or plaintext material. The earlier `:318` alert on the master-key strength floor was disproved by execution — the message carries `len(key)`, never the key.

## Verified by execution before merge

**Substitution — rejected.** `cp staging.json db_master.json` → `SecretEncryptionError`.

**Single-file rollback — rejected.**
```
store ORIGINAL -> backup -> rotate to ROTATED -> restore backup
-> SecretRollbackError: "Secret k is version 1 but version 2 was stored"
```

**Whole-directory restore — honestly bounded, and the WARN says so verbatim:**
```
WARN> Rollback detection is degraded: the secret version manifest lives inside <dir>,
      so it is part of the same backup set as the secrets. A restore of the WHOLE
      directory will revert the version marks along with the secrets and will NOT be
      detected as a rollback -- a rotated credential can be silently reverted to its
      pre-rotation value. Single-file restores ARE detected. To close the whole-directory
      case, set KAILASH_SECRETS_STATE_DIR (or pass state_dir=) to a directory with an
      independent backup lifecycle.
```

With `state_dir` pointed outside the store, the whole-directory restore is **rejected** — verified. So the mechanism is sound and the default's limit is stated rather than implied.

**That WARN is the reason this merges.** It names the residual, the concrete consequence, what IS protected, and the exact remedy. A control that claims precisely what it delivers is the opposite of the defect this issue was filed for — a class documenting encryption it did not perform.

Earlier verification also confirmed: plaintext absent from raw bytes (and from its base64 form), round-trip correct, file `0o600`, dir `0o700`, path traversal rejected on all three methods, 32-char master-key floor enforced, and `SecretManager`'s warn-and-generate defect (#2041) **not** copied — this raises.

Issue #636, opened 2026-04-26
security(api-gateway): remove hardcoded default JWT secret (CRIT)


esperie:
## Summary

CRIT severity. `src/kailash/middleware/communication/api_gateway.py:166-175` ships a hardcoded default JWT signing key `"api-gateway-secret"` (18 chars). Anyone calling `APIGateway(enable_auth=True)` without an explicit `auth_manager=` ships a forgeable JWT auth chain — the secret is in the public open-source repo. The 18-char length also falls below the 32-byte minimum that `JWTConfig.MIN_SECRET_LENGTH` already enforces elsewhere in the same codebase.

## Root cause

```python
if enable_auth:
    if auth_manager is None:
        from ..auth import JWTAuthManager
        self.auth_manager = JWTAuthManager(
            secret_key="api-gateway-secret",   # hardcoded, public OSS, 18 chars
            algorithm="HS256",
            issuer="kailash-gateway",
            audience="kailash-api",
        )
```

Verified live during Wave 5 portfolio spec audit (finding F-C-35).

## Expected fix

Replace the hardcoded default with env-var lookup per `rules/env-models.md` (".env Is The Single Source of Truth"):

1. Read from `KAILASH_API_GATEWAY_SECRET` env var
2. If unset AND `enable_auth=True` AND no `auth_manager` provided -> raise `RuntimeError` with actionable message
3. If set but < 32 bytes -> raise `ValueError` (mirror `JWTConfig.MIN_SECRET_LENGTH`)
4. Add Tier 2 regression test exercising both failure modes + happy path with env var set

## Acceptance criteria

- [ ] No hardcoded `secret_key="api-gateway-secret"` in source tree
- [ ] `APIGateway(enable_auth=True)` without auth_manager + without env var -> typed error at construction time
- [ ] Env var < 32 bytes -> typed ValueError
- [ ] Env var >= 32 bytes -> APIGateway constructs successfully
- [ ] Regression test `tests/regression/test_issue_NNN_api_gateway_default_secret.py`
- [ ] Hotfix release as kailash 2.11.2 (combined with iss-claim fix)

## Cross-references

- Audit finding: workspaces/portfolio-spec-audit/04-validate/W5-C-findings.md F-C-35
- Spec: `specs/middleware.md` (already documents the vulnerability)
- Companion fix: trust JWT iss-claim gap (separate issue, same release)

Filed during Wave 5 portfolio spec audit

Issue #2039, opened 2026-08-10
chore: isort mode disagreement (--files vs --all-files) — cause UNKNOWN, three hypotheses excluded; src_paths half closed by 2cf40a17b


esperie:
Root cause found while converting files under `packages/kailash-dataflow/` during the #2015 burn-down. This is the same defect class #1995/#2028 cleaned up, in a package the fix never reached.

## The defect

The **root** `pyproject.toml` `[tool.isort]` block sets `src_paths`, and carries a comment naming this exact non-determinism — crediting it for the ~2,000 permanently-modified files that #1995 was filed to fix.

`packages/kailash-dataflow/pyproject.toml` declares `[tool.isort]` **without** `src_paths`.

isort resolves configuration **nearest-first**. So for anything under `packages/kailash-dataflow/`, the per-package block wins and the root fix is never applied — the very files most in need of it are the ones that never received it.

## Observable symptom

First-party/third-party classification becomes position-dependent, so a file oscillates between the commit-time hook run and `pre-commit run --all-files`. Observed directly: two template files flipped back and forth between the two runs, and the commit **would not stick** — each run re-modified what the other had just written.

That is the same failure #1995 described, still live in this package.

## Why it is filed separately

It surfaced inside a security PR (#2034). Fixing build configuration inside a credential-leak fix would hide an unrelated change behind a security review, so it is scoped out deliberately. The immediate blast radius was verified as limited to the two files already being touched there.

## Acceptance criteria

- [ ] `packages/kailash-dataflow/pyproject.toml` `[tool.isort]` either sets `src_paths` consistently with the root, or removes its own block so the root config applies.
- [ ] **Sweep every other package** for the same shape — any `packages/*/pyproject.toml` declaring `[tool.isort]` without `src_paths` has the identical latent defect. Report the full list; do not fix only the one observed.
- [ ] Verify by the behaviour, not the config: on a clean tree, `pre-commit run --all-files` exits 0 and modifies 0 files, **and** a commit touching a file under each affected package does not re-modify it. A config diff alone is not evidence.

## Related issues

Follows #1995 (closed by #2028). Surfaced during #2015 (PR #2034).

esperie:
Additional root-cause detail from the #2004 lane — **this changes the fix**.

[...]

PR #2054, opened 2026-08-10
fix(nexus): make enable_auth install real authentication (#2013)

### LINKED ISSUE #2013: nexus: Nexus(enable_auth=True) and app.enable_auth() are inert — a documented security control that installs nothing

esperie:
## Summary

`Nexus(enable_auth=True)`, `app.enable_auth()` and `use_plugin("auth")` set three booleans and installed nothing. Measured on `main`: construct `Nexus(enable_auth=True)`, `GET /workflows` → **200**. Every route was open while the platform logged `✅ Authentication: ENABLED`.

Both branches that looked like an install were structural `hasattr` guards over names with **zero definitions anywhere** in the repo or its dependencies:

| Site | Guard | Definitions of the probed name |
|---|---|---|
| `nexus/plugins.py:103` | `hasattr(gateway, "set_auth_manager")` | **0** |
| `nexus/core.py:4817` | `hasattr(gw, "enable_auth")` | 0 on the gateway class |
| `nexus/core.py:4828` | `hasattr(gw, "enable_monitoring")` | 0 on the gateway class |

The gateway is `EnterpriseWorkflowServer`, which exposes no authentication surface at all — `grep -rnE 'Depends\(|HTTPBearer|APIKeyHeader|verify_token|authenticate' src/kailash/servers/` returns **zero matches**. So `set_auth_manager` was never called, the `MiddlewareAuthManager` the plugin built was dropped on the floor, and the plugin logged success. This is the silent-fallback shape `zero-tolerance.md` Rule 3d enumerates.

### Decision: implement, or call what the gateway exposes?

**Neither of the two options as posed.** The gateway exposes *no* auth API, so (b) was impossible; and adding one to `src/kailash/servers/` is #2025's scope, which this lane must not take. Nexus owns authentication (`framework-first.md`), and it already ships a real Starlette `JWTMiddleware`. The fix installs that on the gateway's FastAPI app — no new swallowed kwarg, no change to the `WorkflowServer` family.

## What changed

- **New `nexus/auth_bootstrap.py`** — builds a `JWTConfig` from the environment and installs `JWTMiddleware` ahead of CORS (Starlette's `add_middleware` prepends, so CORS must stay outermost or preflight `OPTIONS` gets 401'd).
- **Credentials from `.env`** (`env-models.md`): `NEXUS_JWT_SECRET` (≥32 bytes, RFC 7518 §3.2), `NEXUS_JWT_PUBLIC_KEY` + `NEXUS_JWT_ALGORITHM`, or `NEXUS_API_KEY_*` (compared with `secrets.compare_digest`, non-short-circuiting across all keys).
- **Fails loud.** No credential source → `AuthNotConfiguredError` naming every accepted variable, propagated *unwrapped* past `_initialize_gateway`'s blanket `except`. Silently succeeding over an open API is the worst outcome (`security.md` § Secure-Default); this matches the precedent already set for the identical defect in `APIGateway(enable_auth=True)` (#636).
- **API-key-only mode** gets a fresh `secrets.token_urlsafe(48)` for the bearer arm — never written down or returned, so no party can mint a token against it. Fail-closed, not a fixed fallback secret.
- **Monitoring sibling removed**, not reimplemented: `register_metrics_endpoint` right below it is the real install, and an inert probe alongside a working one only makes the working one look conditional.

## Two sibling defects found and fixed (both pre-existing on main, both failing OPEN)

1. **`NEXUS_ENV=prod` was not production.** Three hardening sites inlined `== "production"`: auth auto-enable, CORS defaults, CORS wildcard rejection. `prod` matched none, so those deployments got no auth, permissive CORS, and `allow_origins=['*']` accepted. Now one shared `_is_production_env()` (`security.md` § Enforcement-Surface Parity). **Two tier_2 tests already asserted this and were red on main.**
2. **`_extract_workflow_info` returned empty for every real workflow** — probed `workflow._nodes` / `._connections` / `node._config`; the real API is `nodes` / `connections` / `NodeInstance.node_type` + `.config`. The MCP resource AI agents read to discover a workflow reported zero nodes and zero connections.

## Tests

`tests/regression/test_issue_2013_inert_auth_control.py`, `@pytest.mark.regression`, behavioural — real HTTP requests, status-code assertions. **Fail-first verified against main's source:** the key test returns 200 there. A discrimination control (`enable_auth=False` → same route non-401) ensures the 401 comes from auth, not breakage.

Three unit tests were **rewritten**: they drove `Mock()` objects, and a Mock satisfies every `hasattr` and accepts every call — `test_auth_plugin_apply` passed identically whether the plugin installed authentication or nothing. That is why this shipped.

**Suite:** 3672 passed. 10 failures remain, **identical test IDs to the main baseline** (10 before, 10 after) — all `ValueError: Mismatch between URI parameters set() and function parameters {'uri'}` from FastMCP resource-template registration in `nexus/resources.py`. Distinct pre-existing defect needing its own MCP URI-contract change (`workflow://*` → `workflow://{name}` across 5 resources); flagged, not silently expanded into this PR.

## Mechanical dead-guard sweep

AST-enumerated every `hasattr(x, "N")` / `getattr(x, "N", …)` in production source, then checked whether `N` is defined anywhere across the repo **and** 26,349 dependency files. **1154 probed names → 42 with zero definitions.** Two fixed here (`set_auth_manager`; plus the two type-aware-dead ones the name-level sweep cannot see, since `enable_auth`/`enable_monitoring` are defined on `Nexus`, not the gateway).

Remaining 42, with assessment:

**Security / observability — recommend prioritising:**
| Name | Site | Assessment |
|---|---|---|
| `token_claims` | `nexus/auth/tenant/resolver.py:101` | tenant-isolation path; falls back silently |
| `add_permission_rule` | `middleware/auth/access_control.py:253` | authz rule registration |
| `metrics_aggregator` | `nodes/data/workflow_connection_pool.py:1101` | pool metrics never collected |
| `_ws_server` | `nexus/core.py:4716` | websocket shutdown path |

**Remaining 38** (mostly `getattr` with defaults on optional/duck-typed attrs, lower risk): `LLAMA_FTYPE_MOSTLY_Q4_K_M`, `LLAMA_FTYPE_MOSTLY_Q8_0`, `__dataflow_meta__`, `__response_field__`, `_completed_nodes`, `_condition_field`, `_connection_stats`, `_custom_table_inspection`, `_free_pool`, `_from_status`, `_get_postgresql_version`, `_getters`, `_last_result`, `_last_url`, `_mcp_clients`, `_memory_checkpoints`, `_mock_name`, `_set_runtime_coordinated`, `_spec_class`, `_to_status`, `_trainable`, `cache_namespace`, `call_async`, `conditional_rules`, `default_fallback_value`, `ep_info_buffer`, `estimated_timeline`, `fk_lookup_condition`, `inference_server`, `lora_trainable`, `name_to_value`, `num_timesteps`, `reference_model_id`, `register_pool`, `resource_estimate`, `reward_model_id`, `unregister_pool`.

**Sweep limitation, stated explicitly:** it matches on *name*, not on the probed object's type. A guard probing object `A` for a name defined only on unrelated class `B` reads LIVE. Both `core.py` guards fixed here are exactly that shape — a type-aware sweep would find more.

## Breaking change

`enable_auth=True` (and `NEXUS_ENV=production`/`prod`) now **requires** a credential source and raises without one. Deployments relying on the old behaviour were running unauthenticated regardless; they now fail at startup instead of silently serving an open API.

## Related issues

Fixes #2013


esperie:
## Rebased onto current `main`; bypasses reduced to one

- Rebased onto `origin/main` (`207f7d858`), linear history, merge commits dropped.
- **The stale-branch bypass is gone.** `kailash.utils.command_safety` landed with #2049, so the Tier-1 pre-commit hook now passes and that commit was re-created with hooks running normally.
- **One documented bypass remains**, on a 3-line style commit only: commit-time isort and `pre-commit run --all-files` disagree on where `kailash_mcp` belongs, and each re-groups what the other just wrote. `origin/main` is in the `--all-files` state, so committing the commit-time state would have diverged this branch from the repo's own committed convention. That is #2039.

  Measured, not assumed: `known_first_party` lists `["kailash", "kaizen", "kaizen_agents", "dataflow", "nexus", "pact"]`, but the real first-party import names also include `kailash_align`, `kailash_mcp`, `kailash_ml` — all three absent, so isort must infer them per batch, exactly the non-determinism the `src_paths` comment says that setting was added to eliminate. **Adding the three missing names was tried and rejected:** it changes only these same 2 files, still oscillates between the two file sets, and produces a worse result (two separate first-party groups). The real fix is larger than a config line and belongs to #2039.

  Net effect: `nexus/resources.py`'s import block is now **byte-identical to `origin/main`** — this branch no longer touches it at all.

**Verification against the rebased head:** `pre-commit run --all-files` exit 0, working tree clean. Suite: **3683 passed**, 10 failed — the same 10 baseline test IDs, now tracked as #2056.

## Follow-up issues filed

- **#2056** — Nexus MCP resources fail to register when official FastMCP is importable (`workflow://*` URI template vs `uri` handler param, 5 resources). Carries the 10 baseline test IDs.
- **#2057** — the four security/observability dead guards from the sweep. Severity was **re-verified per site and corrected**: the serious one is `add_permission_rule` (`middleware/auth/access_control.py:253`) — a permission rule is audited as `permission_rule_created` while `getattr(..., "add_permission_rule", None)` is always `None`, so it enforces nothing. `token_claims` is **not** a broken tenant-isolation path as first reported: `token_payload` is set on every authenticated request and checked first, so only the dead compatibility alias is affected.

esperie:
## CHANGELOG migration entry added

`packages/kailash-nexus/CHANGELOG.md`, under a new `## [Unreleased]` heading, labelled `### Changed (BREAKING)` per this repo's stated convention (a MINOR release MAY break; every one carries a migration note). It lands under `[Unreleased]` rather than the shipped `2.16.0` heading — folding an unreleased breaking change into a released version's notes would misreport when it takes effect.

Covers: that `enable_auth=True` and `NEXUS_ENV=production|prod` now require a credential; the exact variables with copy-pasteable generation commands; that construction raises `AuthNotConfiguredError` rather than starting; and plainly that **if this change breaks your deployment, that deployment was serving an unauthenticated API** — the upgrade does not remove access control, it reveals there was none.

Two consequences a reader would otherwise hit at runtime are called out explicitly:

- a `NEXUS_ENV=prod` deployment also picks up the **CORS hardening it was silently skipping** — a wildcard `allow_origins=['*']` will now raise, so explicit origins are required;
- **auth must be enabled before the first request**, because Starlette freezes its middleware stack at startup; `app.enable_auth()` afterwards now raises with that instruction instead of silently doing nothing.

Also documented: the default exempt paths (`/health`, `/health/*`, `/enterprise/health` only — so liveness probes answer, and nothing else is exempt), and `NEXUS_AUTH_EXEMPT_PATHS` for extending them.

Verified against the pushed head: `pre-commit run --all-files` exit 0, working tree clean.

## Follow-ups posted

- **#2057** — added the six-guard disposition table. Re-verified each site: the contrast is **5-of-6**, not 1-of-2 — `check_node_access` and *both* `get_user_permissions` sites also fail closed, so `add_permission_rule` is the lone outlier. Also confirmed the `return {"success": True, "rule_id": ...}` sits **outside** the `if add_rule_fn:` guard, so the caller gets a success and a `hash()`-synthesized id referring to nothing.
- **#2039** — posted both measurements, including the **negative result** so nobody re-runs it: adding the three missing names touches only the same 2 files, still oscillates, and yields two separate first-party groups. Also noted that the three absent names (`kailash_align`, `kailash_mcp`, `kailash_ml`) are exactly the ones whose distribution name differs from their import name, which looks like the thread to pull — and that `src_paths` is still missing from the kaizen and kaizen-agents configs, while dataflow's has since landed.

esperie:
## Independent confirmation from lane 1C (#2026 / PR #2035)

Recording this here rather than filing a new issue — it is the **same defect**, and this PR fixed it.

While running the auth-adjacent suite during #2026 I hit two failures in `tests/tier_2/integration/test_nexus_authentication_defaults.py`, one of which asserts its own diagnosis:

```
AssertionError: ❌ SECURITY BUG: NEXUS_ENV=prod, enable_auth=None -> Expected auth=True, got False
FAILED ...::TestAuthenticationConfigurationMatrix::test_authentication_configuration_matrix[prod-None-True]
FAILED ...::TestHybridAuthenticationDefaults::test_production_detection_from_common_env_vars
```

I confirmed at the time that they were **not** caused by my branch — reverting `src/kailash/nodes/auth/` to the base commit reproduced them byte-identically, and my diff touched zero Nexus files.

**Both now pass on current `main` (`aba8a0878`):**

```
pytest tests/tier_2/integration/test_nexus_authentication_defaults.py
17 passed in 1.29s
```

So the `_is_production_env()` half fixed here covers the path I observed; there is no distinct surviving route from what I saw, and no separate issue is warranted.

Worth noting for the record because it was found from a completely different direction: a lane working on `nodes/auth/` stubs tripped over it incidentally, which is a reasonable signal that the fail-open default was reachable in ordinary use rather than only under a targeted test.

Issue #2002, opened 2026-08-06
Root tests/regression/ has no blanket CI coverage: 1,564 of 1,566 tests never run


esperie:
ADDITIONAL CONTEXT: found during a redteam round on `fix/issue-1720-forest-drain`. Proven pre-existing, not introduced by that branch.

## The gap

The root `tests/regression/` directory holds **142 files / 1,566 collected tests**:

```
$ .venv/bin/python -m pytest tests/regression/ --collect-only -q | tail -1
1566 tests collected in 14.53s
$ ls tests/regression/*.py | wc -l
142
```

Exactly **two files** are named by any workflow:

```
$ grep -rhoE "tests/regression/[A-Za-z0-9_]+\.py" .github/workflows/ | sort -u
tests/regression/test_issue_1332_rfc3161_live_binding.py
tests/regression/test_issue_1332_rfc3161_verify_fail_closed.py
```

`unified-ci.yml` Tier-1 runs `tests/unit/`, `tests/trust/plane/unit/`, and `tests/security/` only. Its regression step (`cd packages/kailash-dataflow`) covers the **DataFlow package** regression dir, not this one. So ~140 files / ~1,564 tests have no CI coverage at all.

## Why it matters

Regression tests are the never-delete tier: each one pins a bug that already shipped once. Testing policy requires a regression test before merge for every bug fix, so this directory is where that obligation lands — and it is the directory CI does not run. A regression test that is never executed records the intent to prevent a recurrence while providing none of the prevention, which is worse than no test, because the row in the directory reads as coverage.

Roughly 20 of these currently fail. Their failure has never reddened a build.

## Expected vs actual

Expected: a directory whose stated purpose is permanent regression pinning is executed on every PR.
Actual: 2 of 142 files are.

## Severity

HIGH — silently voids the regression tier for the core package.

## Acceptance criteria

- [ ] A CI step runs root `tests/regression/` on every PR, with an explicit `-m` marker filter excluding infra-requiring tests (mirroring the DataFlow step, which documents why the explicit filter is required).
- [ ] Infra-requiring root regression tests run in the infrastructure workflow.
- [ ] The pre-existing failures are fixed or explicitly marked, so the new step starts green and any future red is a real regression.

esperie:
Implementation note — the CI step MUST carry an explicit `-m` filter, and the reason is not what the existing CI comment says.

**The default addopts filters nothing.** `pytest.ini` `[pytest]` addopts is only:

```
[...]

PR #2063, opened 2026-08-11
fix(gateway): SecretManager must not invent an encryption key (#2041)

### LINKED ISSUE #2041: CRITICAL: SecretManager generates an ephemeral encryption key and continues at WARNING — every secret is unrecoverable after restart

esperie:
## Summary

`SecretManager.__init__` generated an ephemeral Fernet key when nothing was configured, logged a WARNING, and continued. The key existed only inside that process, so **every secret it encrypted became unrecoverable at the next restart** — the value was still there, still authenticated, and no longer openable by anyone.

Unlike #2024 this sat on the **shipped default path**: `SecretManager()` is the default construction inside `EnhancedDurableAPIGateway.__init__` and `EnterpriseWorkflowServer.__init__`, so it fired for operators who never opted into anything. The variable that would have avoided it, `KAILASH_ENCRYPTION_KEY`, appeared exactly once in the repo — at its own read site.

## Decision: fail-closed at the encryption boundary, not at construction

The issue's AC asks for a raising constructor. I did **not** do that, and the reason is evidence, not convenience.

`SecretManager()` is default-constructed inside two servers' `__init__` bodies (`gateway/enhanced_gateway.py`, `servers/enterprise_workflow_server.py` — both outside this lane's edit scope, three sibling lanes hold them). An unconditional raise converts *"gateway starts, encryption unavailable"* into *"gateway will not construct"* for every operator without the env var — **including the majority who never encrypt a secret at all**. The only cipher consumer, `gateway/resource_resolver.py`, calls `get_secret`, which touches the key only for values already carrying an encryption marker.

What I shipped instead, per `security.md` § Secure-Default For A New Security Feature and matching the #2035 precedent:

1. **Every encrypting and decrypting path raises** `SecretEncryptionError` naming `KAILASH_ENCRYPTION_KEY`. There is no branch on which an unconfigured manager quietly does something weaker — `_key_or_raise()` is a single chokepoint.
2. **Construction logs a one-time ERROR** (`@lru_cache`, not per-instance) naming the protection that is off and the exact wiring that restores it. The message it replaces was a per-instance WARNING naming neither.
3. **Strict mode restores fail-fast for those who want it**: `require_encryption=True`, or `KAILASH_REQUIRE_SECRET_ENCRYPTION=1` for deployments with no code seam to pass a flag through.

**The data-destruction defect is closed identically either way.** No value is ever encrypted under a generated key. What differs is only *when* the error reaches an operator who never encrypts — and for them there is nothing to report.

## Also fixed (same PR, same failure class)

- **No KDF (AC 2).** `Fernet(encryption_key.encode())` required a 32-byte urlsafe-base64 key, so every passphrase an operator would naturally type raised inside Fernet — the documented parameter was unusable. Key material is now stretched with PBKDF2-HMAC-SHA256 (100k iterations) over a **fresh salt per secret, persisted beside the ciphertext**, so the store survives a restart. 32-character floor, matching `FileSecretBackend`.
- **Reference binding.** Sealed values now carry `{"ref": …, "secret": …}` and a read refuses an envelope written for a different reference — the substitution defence #2024 gave the sibling class, which the manager's own layer lacked.
- **One shared envelope implementation.** `_derive_fernet` / `_build_envelope` / `_open_envelope_payload` are now shared with `FileSecretBackend` rather than duplicated, so the iteration floor, salt floor and reference binding cannot drift between the two classes that both encrypt secrets in this module (`security.md` § Credential Decode Helpers — per-adapter copies are BLOCKED).
- **Documentation (AC 3).** `KAILASH_ENCRYPTION_KEY`, `KAILASH_REQUIRE_SECRET_ENCRYPTION`, `KAILASH_SECRETS_MASTER_KEY` and `KAILASH_SECRETS_STATE_DIR` are now in `.env.example`. An undocumented variable is how the fail-open branch became the normal branch.

## Fail-first evidence (verbatim)

Both runs are the identical command from the worktree root. **RED — unmodified source** (`git status --porcelain src/kailash/gateway/security.py` empty):

```
15 failed, 5 passed in 3.21s
```

The two headline failures, verbatim:

```
___________ test_store_refuses_when_no_encryption_key_is_configured ____________
E   Failed: DID NOT RAISE <class 'kailash.gateway.security.SecretEncryptionError'>
------------------------------ Captured log call -------------------------------
WARNING  kailash.gateway.security:security.py:126 Using default encryption key - not secure for production!

____ test_no_secret_is_ever_written_under_a_key_that_dies_with_the_process _____
src/kailash/gateway/security.py:154: in get_secret
    decrypted = self._cipher.decrypt(encrypted_secret[10:].encode()).decode()
E   cryptography.fernet.InvalidToken
------------------------------ Captured log call -------------------------------
WARNING  kailash.gateway.security:security.py:126 Using default encryption key - not secure for production!
WARNING  kailash.gateway.security:security.py:126 Using default encryption key - not secure for production!
```

That second failure **is the data loss**, reproduced rather than described: a manager accepts the write, a second manager over the same durable on-disk backend cannot open it, and the ephemeral-key warning fires twice — once per instance, which is the whole defect.

**GREEN — after the fix, same command:**

```
20 passed in 4.36s
```

The 5 tests that passed in **both** runs are the negative controls (`test_configured_fernet_key_still_round_trips_across_a_restart`, `test_unencrypted_storage_still_works_without_a_key`, `test_plain_values_still_read_back_without_a_key`, `test_legacy_raw_fernet_values_are_still_readable`) plus the checkout-provenance guard. They exist so a change that "closed" the defect by breaking encryption outright could not masquerade as a pass. Both polarities are pinned, and no test uses `Mock()` — the backends are real implementations of the ABC, and the restart tests use a real on-disk `FileSecretBackend`.

**Provenance.** A bare `python -c "import kailash"` in this worktree resolves the package from the **main checkout** (`/Users/esperie/repos/kailash/build/kailash-py/src/kailash/__init__.py`) — every non-pytest check here would be vacuous. Under pytest, `pythonpath = src` resolves the worktree, verified in-process:

```
RESOLVED: /Users/esperie/repos/kailash/build/.kailash-py-wt/w1b/src/kailash/gateway/security.py
```

That assertion is baked into the test module so it cannot silently regress.

## Regression evidence

| Suite | Result |
|---|---|
| #2024 tests (`test_issue_2024_file_secret_encryption.py`) | 82 passed — unchanged by the shared-envelope refactor |
| `tests/integration/gateway` + `tests/unit/gateway` + `tests/unit/servers` + `tests/tier2_integration/servers` | 217 passed |
| Repo-wide `-k "secret or encryption"` | 348 passed, 4 skipped |
| `pre-commit run --all-files` | all hooks passed, exit 0 |

## Sibling sweep

**In-file — clean.** No `Fernet.generate_key()` remains anywhere in `security.py`. Every remaining key-material site was checked: `_derive_fernet` (the shared KDF), `_fernet_for_raw_key` (legacy reads only — returns `None` rather than generating), and `FileSecretBackend.__init__` (already fails closed from #2024). The remaining `logger.warning` calls concern rollback-manifest placement and file permissions, not key generation.

**Repo-wide — `src/` clean.** `Fernet.generate_key` in `src/` appeared only at `security.py:130`, the defect itself. The one hit in `packages/kailash-dataflow/src/dataflow/core/multi_tenancy.py:1153` is an error-message string *instructing the operator to generate a key* — correct as-is.

**One same-class sibling found, NOT actioned — outside this lane's scope.** `packages/kailash-kaizen/src/kaizen/security/encryption.py:16-26`:

```python
def __init__(self, key: bytes = None, salt: bytes = None):
    """
        key: 32-byte encryption key (if None, generates random key)
    """
    if key is None:
        # Generate random 256-bit key
        key = AESGCM.generate_key(bit_length=256)
```

This is the identical failure shape — an ephemeral AES-256-GCM key on the default path — and is arguably worse, because it is **entirely silent** (not even the WARNING this PR removes). `FieldEncryptor.__init__` (line 235) propagates the same `key: bytes = None` default into it. `packages/kailash-kaizen/` is held by a sibling lane, so I did not touch it; **it warrants its own issue.**

## What I could not verify

`tests/integration/test_enhanced_gateway_integration.py` (5 tests) and `tests/e2e/user_journeys/test_enhanced_gateway_user_flows.py` (2 tests) **fail on this branch and also fail identically on unmodified `main`**. Instrument: the main checkout is at `586062dbc`, which is *exactly* this branch's merge-base, and running the same 7 tests there gives `7 failed, 1 passed, 1 skipped` — with `WARNING kailash.gateway.security:security.py:126 Using default encryption key` in the captured setup log, confirming those runs exercised the old code path. If this change had caused them, main would be green.

The failures are `target_input` pydantic validation and `datetime is not JSON serializable` inside the workflow runtime (`runtime/async_local.py`) — a different subsystem, not secret management, and behind Docker-dependent markers. They are outside this lane's scope boundary and are **not** fixed here.

Their fixtures did construct `SecretManager()` bare and then store a secret, so they were genuine blast radius: both now pass an explicit `encryption_key=`, which is the migration an operator must make. That part is fixed; the residual runtime failures are pre-existing.

## Breaking change

`CHANGELOG.md` carries a `### Security (BREAKING) — SecretManager no longer invents an encryption key (#2041)` entry under the existing `[Unreleased]` heading (not folded into a shipped version), with the three-case migration.

## Related issues

Fixes #2041


esperie:
## CodeQL round — fixed, not deferred

The first push raised **4 high-severity `py/clear-text-logging-sensitive-data`** alerts at `security.py:322-325`, all inside the new one-time ERROR signal.

They were a name-heuristic false positive: the flagged arguments were `SECRET_MANAGER_KEY_ENV`, `MIN_SECRET_KEY_LENGTH` and `SECRET_MANAGER_STRICT_ENV`, which hold the **names** of environment variables and an integer floor — never a value. No secret was ever logged.

I fixed it rather than filing a runtime-safety deferral (`zero-tolerance.md` Rule 1b): the variable names are now spelled out literally in the message, so there is **no dataflow into the logging sink at all**. That answers the alert instead of suppressing it.

The cost is that the message can now drift from the constants that define it, so `test_the_loud_signal_names_the_current_constants` pins all three. That guard was mutation-checked rather than assumed — with `MIN_SECRET_KEY_LENGTH` monkeypatched to `99` it fails:

```
MUTATED-CONSTANT: 99
MESSAGE-CONTAINS-99: False
E   AssertionError: assert '99' in 'Secret encryption is NOT configured: KAILASH_ENCRYPTION_KEY is unset ...'
1 failed in 2.11s
```

The mutation demonstrably reached the code under test, so the green is not a vacuous pass.

**CI on head `18fc196f`: 13 checks success, zero non-success.** CodeQL now reports *"No new alerts in code changed by this pull request"*.

The earlier `build` red on the first head was a concurrency cancellation (`Canceling since a higher priority waiting request for Build Documentation-fix/2041-secretmanager-ephemeral exists`), not a real failure.

esperie:
## Review findings addressed — M1, M2, M3, M4, L2, L3 all fixed

Pushed `d6b876cc8`. Each finding has fail-first evidence: I ran the new tests against the reviewed head `18fc196f` in a throwaway worktree, and **9 of them fail there** — so none of them is a test that would have passed either way.

```
9 failed, 24 passed in 6.23s   # new tests vs. reviewed head 18fc196f
```

### M1 — KDF cost bounded, and off the event loop
Ceiling drops `10_000_000` → `1_000_000` (10x the write value, not 100x), and derivation now runs via `asyncio.to_thread`.

**I extended this past what you scoped.** `FileSecretBackend` has the identical shape — a read derives twice (envelope + version manifest), a store up to four times — so leaving it inline would have been the "we'll do the others later" split that Enforcement-Surface Parity blocks. Both classes are offloaded at their async boundaries.

Three tests: the ceiling is a small multiple of the write value; an over-ceiling envelope is refused *structurally* (asserted via elapsed time < 1s, so the refusal cannot be the KDF having run); and a heartbeat coroutine ticking every 5ms still ticks during a store and a read.

### M2 — docstring scoped, and the hole made closeable
You were right that the claim was true of the function and false of the layer above. I did **both** halves rather than picking:
- `_open_envelope_payload`'s docstring now scopes the no-fallback guarantee to that function and names `get_secret` as the boundary where an unmarked value is returned unchecked.
- `require_encryption` now *also* refuses any unsealed value, so the hole is closeable rather than only documented. `get_secret`'s own docstring states plainly that marking is not a security boundary because the writer chooses the marker.

A control test pins that non-strict mode still returns unsealed values — the permissive path is the documented default and the default `EnvironmentSecretBackend` depends on it.

### M3 — legacy unbinding in the CHANGELOG, plus strict-mode refusal
The migration paragraph now says the pre-2.64 format carries no reference binding, that this cannot be fixed in-format, and to re-store rather than run on compatibility indefinitely. Strict mode additionally declines the unbound path outright, so an operator can enforce it rather than just be warned.

### M4 — `max_age=` added, and the residual bound stated
Rather than only documenting the gap, I folded `stored_at` into the sealed payload and added an optional `max_age=`, mirroring `FileSecretBackend`. Test: rotate, then replay the superseded envelope — `SecretRollbackError`.

It is documented as **partial**, in the class docstring and the CHANGELOG: it catches an aged-out replay, not a recent one. The complete check needs authenticated state outside the stored value, and a manager handing values to an arbitrary backend has nowhere to keep one. Stated as a known bound rather than an assumed protection.

### L2 / L3
Key length is measured on the stripped value (`" " * 32` is refused), while the key is still **used verbatim** — trimming would silently change the derived key for anyone whose passphrase ends in whitespace, so the error says so explicitly. References are escaped and length-capped before interpolation; the test asserts the reference is still named, that no raw newline survives, and that it appears escaped rather than stripped.

### Regression
82/82 on the #2024 suite (which the `FileSecretBackend` offload touches), 230 passed across `tests/integration/gateway` + `tests/unit/gateway` + `tests/unit/servers` + `tests/tier2_integration/servers`, 33 in this PR's own file.

### On the two things you flagged
Thanks for the note on the test labelling — I'll keep the explicit negative-control marking and the falsifying-result docstrings. I've carried it into all 13 new tests here.

Noted that `tests/integration/gateway/` is collected by no CI job (#2002 class). Flagging the consequence rather than assuming it is understood: **every green I have reported on this PR is from local runs**, and the regression protection will not re-run on future changes until that lands.

PR #2064, opened 2026-08-11
fix(servers,api,channels): fail closed on unauthenticated proxy routes (#2025)

esperie:
## Summary

`WorkflowServer.proxy_workflow()` registered an **unauthenticated** catch-all reverse proxy at `/workflows/{name}/{path:path}` for GET/POST/PUT/DELETE/PATCH. The only middleware in the module was CORS; there was no `Depends`, no API-key check, no auth middleware anywhere on the path. It forwarded an arbitrary method, path and query string to the backend at the registered URL and returned the body — and stripped `Authorization` on forward, so the backend could not re-authorize the caller either. The identity was removed and nothing replaced it.

**This PR closes that route on both surfaces, and closes a full SSRF that the issue's own analysis concluded was not constructible** (see the next section — that conclusion was wrong, and so was mine). **It still does NOT close all of #2025** — see § What this does not fix. It is `Refs`, not `Fixes`, deliberately.

## The CodeQL alert was REAL — and it is a FULL SSRF

**This section replaces an earlier one arguing the `py/partial-ssrf` framing was wrong. That
argument was mine, it was wrong, and the correction is the most important thing in this PR.**

Both proxy handlers captured their closure state as **default arguments**:

```python
async def proxy_handler(request: Request, path: str, _url=proxy_url, _allowlist=path_allowlist):
```

**FastAPI inspects the handler signature and treats any non-path parameter carrying a plain default
as a QUERY parameter.** So `_url`, `_allowlist`, `_name` and `_backends` were all caller-writable.
Measured, with a second HTTP server standing in for the attacker's host:

```
NORMAL   -> 200 {"who": "legitimate-backend", "path": "/data"}
?_url=   -> 200 {"who": "ATTACKER-CONTROLLED-HOST", "path": "/data?_url=http://127.0.0.1:64545"}

VERDICT: FULL SSRF - caller redirected the request to an arbitrary host
```

The path allowlist was writable the same way, so the registration-time restriction could be widened
from the query string.

**What the original analysis got wrong.** #2025 concluded "no authority pivot is constructible",
reasoning about `f"{_url.rstrip('/')}/{path}"` and showing that `@evil.com/` lands in the path
component. That is true *about the concatenation* and it is the wrong place to look — the
destination was never fixed to begin with. The reasoning was never tested against a running server.
I repeated the same error in my own deferral (issue #2065, now withdrawn and closed): I probed the
string-building, got a clean result, and treated it as clearing the whole flow. That instrument
could not have detected this defect, and I cited it as though it had.

**How it surfaced.** A `py/log-injection` alert flagged `extra={"workflow": _name}`. A
registration-time constant should not be tainted; chasing *why* it was is what exposed the SSRF.
The finding that looked cosmetic was the thread that led to the critical one.

**Fixed** in `4a243fb17`: handlers capture from the enclosing scope on both surfaces. Each
`proxy_workflow` call has its own scope, so the late-binding hazard the default-argument idiom
guards against does not apply here.

**Tests** — an attack test per surface, plus a structural test asserting each handler declares only
`request` and `path` and that neither carries a default, so reintroducing the idiom fails at the
source rather than waiting for someone to write the matching attack. **Mutation-verified:**
reintroducing `_url=proxy_url` reddens both, and the failure message names `_url` in the signature,
which proves the mutation reached the code rather than the tests being vacuous.

```
FAILED test_destination_cannot_be_overridden_by_query_parameter
  AssertionError: SSRF: the caller redirected the proxy to an arbitrary host
FAILED test_proxy_handler_exposes_no_query_parameters
  AssertionError: handler exposes unexpected parameters {'path', 'request', '_url'};
                  any parameter with a plain default becomes a caller-writable query parameter
```

## The decision: FAIL-CLOSED

**Fail-closed, refusing at registration.** Registration is this API's construction moment, so this matches the precedent set in this repo for `Nexus(enable_auth=True)` (#2013 / PR #2054) and `APIGateway(enable_auth=True)` (#636): both refuse to construct rather than silently serve an open API, per `security.md` § Secure-Default.

Reasoning: an opt-in gate leaves every *existing and every new* proxy registration open, which does not close the issue for anyone who does not read logs. A registration that fails loudly at deploy time fails in the deployment that misconfigured it; an open proxy fails silently, in production, for the backend's entire surface. Callers whose registrations start raising were already running an unauthenticated proxy — the raise reports a pre-existing exposure rather than creating a new problem.

`proxy_workflow()` now raises `ProxyAuthNotConfiguredError` unless the caller supplies exactly one of:

1. `auth_dependency=<callable>` — any FastAPI dependency.
2. An auth manager on the server — `WorkflowServer(auth_manager=…)` or the new `set_auth_manager(…)`, which must expose `get_current_user_dependency()`. `kailash.middleware.auth.MiddlewareAuthManager` is the shipped implementation. (`set_auth_manager` is also the name `nexus/plugins.py` probed for on this gateway and never found in #2013 — it now exists and does something.)
3. `declare_external_auth("<reason>")` — an explicit, reason-bound, WARN-logged acknowledgement, for deployments where an ASGI middleware installed outside this class authenticates every request (e.g. `nexus.auth.jwt.JWTMiddleware`, which `Nexus(enable_auth=True)` installs since #2054).

**Why (3) exists, stated plainly:** a route-level gate cannot observe app-level middleware. Without an escape, a *correctly secured* Nexus deployment could not register a proxy at all — a false-positive fail-closed. The escape installs nothing and verifies nothing; it requires a non-empty reason (blank raises) and logs a WARNING naming the unprotected route at every registration. It is an audited acknowledgement, not a silent opt-out.

The `auth_manager` probe **also fails closed**: a manager that cannot produce a callable dependency raises. A `hasattr` guard whose False branch continues silently is exactly how `enable_auth=True` shipped as a no-op for two releases (`zero-tolerance.md` Rule 3d).

## The other three acceptance criteria

- **`allowed_paths` is now required**, replacing the unbounded catch-all. `"status"` exact, `"api/*"` prefix at any depth, `"*"` everything (the old behaviour, accepted but only when written down), or a compiled `re.Pattern` matched with `fullmatch`. Non-matching path → 404. An empty allowlist matches nothing, so a construction bug cannot open the route.
- **`allowed_methods` defaults to `["GET"]`**, not all five. The route is registered only for the named methods. **Measured** that FastAPI's `api_route` does *not* auto-add HEAD the way Starlette's plain `Route` does — `REGISTERED METHODS: ['GET']`, `HEAD -> 405` — so the allowlist is not silently widened.
- **Traversal rejected before the target URL is built** — `..` segments, null bytes, control characters, backslashes, and encoded `%2e`/`%2f`/`%5c` — rather than relying on `yarl` to normalize. `yarl` and `httpx` normalize differently and neither is a security control.

- **The forwarded path is built from a positive-charset match, not the raw parameter.** `SAFE_FORWARD_PATH_RE` permits RFC 3986 `pchar` plus `/` and every non-ASCII codepoint (so internationalized paths still forward); SPACE, CR, LF, TAB, every other control character and ``? # \\ < > " { } | ^ ` `` get 400. The URL is built from the **matched group**, so no byte outside the allowlist reaches the wire whatever the client does with re-encoding.

  This settles a question the #2025 analysis explicitly recorded as **UNVERIFIED** — whether a decoded CR/LF in `path` could reach the wire as request-line injection, with `yarl`/`aiohttp` "expected to re-encode or reject; not tested, no evidence either way". With the barrier it cannot, and the answer no longer depends on the client library.

  It is also what closed **CodeQL alert #11497** (`py/partial-ssrf`), which this PR's first push re-raised against the moved code. The issue is right that the SSRF *framing* is wrong and that validating `proxy_url` closes nothing — but the query's actual recommendation is to restrict the path so it cannot alter the URL, which is worth having on its own merits. So it is closed with a real barrier rather than deferred under `zero-tolerance.md` Rule 1b, because a genuine fix was available.

Plus the non-security nit: `params=dict(request.query_params)` dropped every duplicate key but the last; now `multi_items()`.

## Framework-first note

This is core (`src/kailash/servers/`), and core cannot import Nexus — Nexus depends on `kailash`, not the reverse. So the framework primitive used is core's own `kailash.middleware.auth.MiddlewareAuthManager.get_current_user_dependency()`, not a hand-rolled `HTTPBearer` check. Nexus's `JWTMiddleware` is the template for the app-level work described below, not a dependency.

## Fail-first evidence, verbatim

One probe, run unchanged against both trees. Pre-fix source is the main checkout at `586062dbc`, which is this branch's base.

```
===== RUN A: pre-fix source (main checkout, 586062dbc) =====
SOURCE UNDER TEST: /Users/esperie/repos/kailash/build/kailash-py/src/kailash/__init__.py
REGISTRATION: accepted with NO auth argument
UNAUTHENTICATED GET /workflows/internal/admin/users -> 200
BODY: {"secret": "INTERNAL-BACKEND-DATA", "path": "/admin/users"}
VERDICT: VULNERABLE - backend data returned to an unauthenticated caller

===== RUN B: fixed source (worktree w1a) =====
SOURCE UNDER TEST: /Users/esperie/repos/kailash/build/.kailash-py-wt/w1a/src/kailash/__init__.py
REGISTRATION: refused -> ProxyAuthNotConfiguredError
```

The suite itself, pinned to the pre-fix source, is red at collection — and the error names the pre-fix path, which is what proves the pin took effect rather than silently resolving the worktree:

```
E   ImportError: cannot import name 'ProxyAuthNotConfiguredError' from 'kailash.servers'
    (/Users/esperie/repos/kailash/build/kailash-py/src/kailash/servers/__init__.py)
```

A collection error reports on the symbol, not the behaviour, which is why the behavioural half was measured separately with the probe above.

**GREEN after:** `64 passed`; with the neighbouring server suite, `96 passed`.

## Tests

`tests/regression/test_issue_2025_unauthenticated_proxy_route.py` — 64 tests, `@pytest.mark.regression`. **No `Mock` anywhere in the file.** A `Mock` satisfies every `hasattr` and accepts every call, which is how the inert auth control in #2013 shipped green. These drive a real ASGI stack through `TestClient`, forward to a **real HTTP backend on a real socket**, and mint **real HS256 tokens** with the shipped `MiddlewareAuthManager`.

**Discrimination control:** `test_authenticated_request_is_forwarded` hits the *same* route with a valid token and gets 200 with the backend's body — so the 401 is authentication, not a route this change broke. Same shape for the method allowlist (DELETE → 405, GET on the same path with the same credentials → 200).

Also covered: per-registration dependencies, invalid tokens, the fail-closed `auth_manager` probe, `set_auth_manager(None)` refusal, subclass parity for `Durable`/`Enterprise`, path allowlist semantics, traversal, and duplicate query keys.

One honest limitation: the double-encoded traversal case is asserted at the guard function rather than over HTTP. **Measured:** httpx decodes `%25` before sending, so `%252e%252e` and `%2e%2e` both arrive at the handler already collapsed to `..`. Both are refused; only the branch that fires differs.

### A correction worth recording

While adding the charset barrier I briefly "verified" the pattern through a shell `python -c` one-liner and read the result as confirmation. The shell had mangled the escapes, so the probe compiled a **different string** than the module held — it reported on itself, not on the code. The charset tests now assert against the **compiled module constant** (`SAFE_FORWARD_PATH_RE.pattern`), which is the only instrument that can distinguish a correct pattern from a near-permissive one. Re-measured that way: 16 must-reject and 9 must-accept cases, all correct.

## Regression evidence — this diff introduces zero new failures

Failure IDs compared as **sets**, not counts, across `tests/tier2_integration + tests/regression +
tests/unit`, same command on base and branch:

```
BASE:  99 failure IDs
BRANCH: 99 failure IDs

NEW failures introduced by this branch:  (none)
Failures on base but not on branch:      (none)
```

An intermediate run of that sweep **did** surface 5 new failures, all in
`tests/tier2_integration/test_todo029_api_gateway.py` — a file at the top level of
`tests/tier2_integration`, outside the `api/` subdirectory the earlier targeted runs covered. Four
were stale assertions of the open contract. **The fifth was a real defect I had just introduced**:
switching `execute_chain` to `with LocalRuntime() as runtime:` meant the test's `MagicMock`
returned a *fresh* auto-generated mock from `__enter__()`, so `runtime.execute(...)` yielded a Mock
instead of the configured tuple. Fixed by making the mock model the context manager. That is
exactly the Mock failure mode the #2013 post-mortem names, and comparing counts instead of sets
would have hidden it.

99 pre-existing failures on the base (dataflow express DDL, parameterized generics, node-registry
collisions, retry classification, Redis redaction, MCP OAuth, brief-realizer allowlists). None are
in `servers/`, `api/gateway`, `channels/`, or otherwise touched by this diff. Flagged rather than
absorbed: fixing them is far outside one shard budget and outside this lane's scope.

Targeted suites all green: servers + api + channels + both regression suites, **179 passed**, and
they stay green under `-W error::DeprecationWarning`. `mypy --ignore-missing-imports` reports no
errors in any changed source file.

## One same-class defect fixed in passing

`src/kailash/servers/__init__.py` advertised `EnterpriseWorkflowServer(title=…, enable_auth=True)` in its shipped docstring example. **No such parameter exists on any server in the package** — it lands in `**kwargs`, where the only key ever consumed is `shutdown_timeout`, and is discarded with no error and no warning. A user following the shipped example believed they had enabled authentication and got none (`zero-tolerance.md` Rule 3c, on a security control). Corrected to `auth_manager=`, with a note naming the defect.

## What this does NOT fix — read before merging

The issue's comment argues that this proxy route is the **least reachable** instance of the defect, and that the real exposure is default-reachable. It flags its central claim as an inference and asks for it to be falsified before the fix shape is trusted. **I measured it, and it is correct:**

```
A app-level Depends  -> /direct        : 401
A app-level Depends  -> mounted execute: 200      <-- OPEN
B middleware         -> /direct        : 401
B middleware         -> mounted execute: 401      <-- closed
B middleware + creds -> mounted execute: 200      <-- discrimination control
```

`register_workflow` and `register_mcp_server` use `app.mount(...)`, and route/app-level dependencies **do not reach a mounted sub-application**. Middleware does. So on a default `create_gateway()` + `register_workflow(...)` these remain unauthenticated after this PR:

| Route | Consequence |
| --- | --- |
| `POST /workflows/{name}/execute` | anonymous arbitrary workflow execution |
| `POST /workflows/{id}/signals/{signal}` | mutates running-workflow state (rate-limited, not authenticated) |
| `GET /workflows`, `WS /ws`, `GET /metrics`, `GET /dashboard` | enumeration, open socket, metrics, dashboard |
| `/mcp/{name}/…` | the whole MCP surface |
| `GET /durability/requests` | dumps `path`, `method`, `client_ip` per in-flight request |

Closing that needs an ASGI **middleware** on `WorkflowServer.app` delegating to core's `kailash.trust.auth.jwt.JWTValidator`, with a **named** `require_auth` parameter (never `**kwargs` — that is exactly how `enable_auth` got swallowed). That is a materially larger change whose fail-closed default would stop existing `create_gateway()` deployments from booting, so it is being confirmed with the repo owner rather than absorbed here.

**This is why the PR says `Refs #2025` and not `Fixes #2025`.** Auto-closing the issue on merge would close it while the more reachable half is still open.

## Enforcement-surface parity — DELIVERED, both surfaces

**All five server modules swept, plus the independent gateway surface:**

| # | Module | Finding |
|---|---|---|
| 1 | `servers/workflow_server.py` | **The defect named in #2025.** Gated by this PR. |
| 2 | `servers/durable_workflow_server.py` | Subclasses `WorkflowServer`, adds no catch-all -> inherits the gate. Asserted in tests. |
| 3 | `servers/enterprise_workflow_server.py` | Same -> inherits the gate. Asserted in tests. |
| 4 | `servers/gateway.py` | `create_gateway()` factory only; registers no routes. Nothing to gate. |
| 5 | `servers/connection_metrics_router.py` | Fixed GET-only `/connections` routes, no catch-all. No proxy defect — but also no auth, which is part of the default-surface gap (#2072). |
| — | `api/gateway.py` | **Second independent surface with the same defect, and worse.** Gated by this PR; details below. |
| — | `channels/api_channel.py` | Thin delegate to (1); inherits the gate, and now threads its arguments through. |

`security.md` § Enforcement-Surface Parity requires every independent surface with this shape to
learn the gate in the SAME change, through ONE shared implementation. Both now do.

`kailash.api.WorkflowAPIGateway.proxy_workflow` was the second surface and the **worse** one:
mounted at `/{name}/{path:path}` (one segment shallower), **seven** methods including HEAD and
OPTIONS, no path validation. Fail-first, one probe, both trees:

```
===== RUN A: pre-fix source (main checkout, 586062dbc) =====
REGISTRATION: accepted with NO auth argument
UNAUTHENTICATED GET /internal/admin/users -> 200
BODY: {"secret": "INTERNAL-BACKEND-DATA", "path": "/admin/users"}
BACKEND SAW Authorization: Bearer caller-secret-token
BACKEND SAW Cookie      : session=abc123
UNAUTHENTICATED DELETE /internal/records/1 -> 200
VERDICT: VULNERABLE - open proxy, destructive method reachable, AND caller credentials
         forwarded to the backend

===== RUN B: fixed source (worktree w1a) =====
REGISTRATION: refused -> ProxyAuthNotConfiguredError
```

### The credential-forwarding half

Lines 5–6 of that output are a second defect, unique to this surface: its header filter excluded
only `host` and `content-length`, so an **unauthenticated** caller's `Authorization` and `Cookie`
were handed to whichever round-robin backend was selected.

The two proxy surfaces disagreed here and **both were wrong, in opposite directions** —
`WorkflowServer` stripped `Authorization` unconditionally, so a backend could never re-authorize,
which #2025 called out explicitly as compounding the exposure. Both now share one constant
(`PROXY_CREDENTIAL_HEADERS`), one default (**strip**), and one explicit opt-in
(`forward_credentials=True`) for backends that are trusted and genuinely need to re-authorize the
original caller.

### Kwarg plumbing

`kailash.channels.APIChannel.proxy_workflow` inherited the gate through delegation but its
signature could not pass `allowed_paths` / `allowed_methods` / `auth_dependency` /
`forward_credentials`, leaving the gate unconfigurable through that channel
(`security.md` § Multi-Site Kwarg Plumbing). An unconfigurable gate is a gate that gets worked
around. A test now asserts every argument is threaded, so a future edit that silently drops one
fails.

### Module relocation, and why

The shared guard is `kailash.utils.proxy_guard`, not `kailash.servers.proxy_guard`. Importing it
from `kailash.api.gateway` executed `kailash/servers/__init__.py`, which pulls the enterprise
server and loops back to `kailash.api.gateway` — a circular import, caught immediately by a smoke
run. The module has **no kailash dependencies at all** (only `logging`, `re`, `typing`), so a
neutral home is where it belonged. The public names are still re-exported from `kailash.servers`,
and a test asserts the re-export is the same object.

### Two unrelated defects fixed in `api/gateway.py` while it was open

Both surfaced by this work, neither in the original scope:

- The gateway WebSocket close path was `except Exception: pass`. It is a cleanup path where
  failure is expected, so the carve-out in `zero-tolerance.md` Rule 3 applies — but an
  *unexpected* close failure was equally invisible. Now logged at debug.
- `WorkflowOrchestrator.execute_chain` built a bare `LocalRuntime()` that was never closed,
  leaking on every call and emitting a `DeprecationWarning` slated to become an error in v0.12.0.
  Now context-managed, so it is released even when a chain hop raises.

## Adversarial security review — findings fixed in this PR

The review returned MERGE-WITH-FIXES. Items 1, 2, 6 and 7 came back clean; the closure-capture SSRF
fix survived a refutation attempt on all three surfaces. **One HIGH survived it**, and it falsified
the security argument `proxy_guard`'s own module docstring was making.

### F1 (HIGH) — the authority pivot IS constructible, on hop 2

`aiohttp` defaults `allow_redirects=True`. Every control this route enforces — the auth gate, the
path allowlist, traversal rejection, the charset barrier — applies to **hop 1 only**. A backend with
any open redirect under an allowlisted path hands an authenticated caller an arbitrary host, and
query strings are forwarded verbatim so `?next=` is caller-controlled.

Measured, with the registration hardened to `allowed_paths=["api/*"], allowed_methods=["GET"]`:

```
status      : 200
body        : {"AccessKeyId": "ASIA-PIVOTED-CREDENTIAL", "Type": "AWS-HMAC"}
caller sees 'location' header: False

VERDICT: PIVOTED - proxy followed the redirect off-host and returned the metadata body
```

After the fix, same probe: `status: 302`, `VERDICT: no pivot - redirect not followed`.

Disabled **explicitly at both call sites** rather than inherited — the two surfaces use different
clients whose defaults disagree (`aiohttp` True, `httpx` False), so an invariant this proxy depends
on has to be stated, not assumed. The module docstring now records both refutations of its
no-authority-pivot claim and names the no-redirect precondition the property rests on.

### F2 (MEDIUM) — hop-by-hop headers

`connection`, `keep-alive`, `proxy-connection`, `te`, `trailer`, `transfer-encoding`, `upgrade`,
`expect` were forwarded on both surfaces. `Transfer-Encoding: chunked` alongside a body the HTTP
client frames itself gives the backend two disagreeing statements about where the request ends — the
CL.TE request-smuggling primitive. One shared `PROXY_HOP_BY_HOP_HEADERS`, unioned on both.

### F3 (MEDIUM) — response framing headers

`content-encoding` and `content-length` were echoed from the backend, but both clients return
**decompressed** bytes — so the response labelled plaintext as gzip and stated the compressed length
over a decompressed body. Pre-existing on main and identical on both surfaces; fixed here because it
is inside the region this PR rewrote (`zero-tolerance.md` Rule 1a — "same on main" is not a
disposition).

### F5 (LOW) — the extraction that sanitized nothing

The forwarded path now comes from `path`, not `match.group(0)`. For a *successful* `fullmatch` those
are the same string, so the extraction never sanitized anything — the safety is entirely in the
rejection. Written as an extraction it invited a later `fullmatch` → `match` edit that looks
equivalent and silently forwards a **truncated** path (`'a b'` → `'a'`) instead of refusing it.

### F8 (LOW) — the positive branch

`EnterpriseWorkflowServer(auth_manager=…)` had only refusal coverage. The refusal branch still
passes if the kwarg is swallowed into `**kwargs` after a refactor — which is the `enable_auth`
defect this PR corrected. The accepted branch is now driven end to end.

### All guards mutation-verified

Each mutation reddens only its paired test:

```
RED   F1 server: allow_redirects=False -> True          1 failed
RED   F1 gateway: follow_redirects=False -> True        1 failed
RED   F2 server: drop hop-by-hop union                  1 failed
RED   F3 both: re-admit content-encoding                2 failed   <- both surfaces
```

F3 reddening **two** tests is the check that the shared constant genuinely covers both surfaces
rather than one of them having its own copy.

### Filed, not folded in

**#2085** — unbounded response buffering (`resp.read()` with no cap) plus a per-request
`aiohttp.ClientSession` on the server surface where the gateway correctly reuses one. Different
defect class from the auth/SSRF work, and the cap needs a design decision about where the limit
lives; the fd/socket churn matters given #2078.

### Route ordering — read this before assuming the namespace is gated

`/workflows/{workflow_id}/signals/{signal_name}` and `/queries/{query_name}` are registered in
`__init__`, **before** any proxy route, and Starlette matches in registration order. So for a
proxied workflow `internal`, `POST /workflows/internal/signals/x` reaches the **unauthenticated
signal handler**, not the gated proxy. That is not an SSRF bypass and is squarely task #27's scope,
but **"the proxy route is gated" must not be read as "`/workflows/{name}/**` is gated."**

## CodeQL status — measured, and NOT re-deferred by me

**CI on head `8a03ceb32`: `success: 12, skipped: 4, failure: 1`.** Every test matrix is green
(Python 3.11/3.12/3.13/3.14, PACT, DataFlow Tier 1 + Infra Regression, test-with-infrastructure,
Analyze Python, build). All 4 `skipped` belong to a superseded duplicate run and each has a SUCCESS
twin in the winning run — skipped is not counted as passed. **The single red is CodeQL.**

### What the alert set did across this PR

| | `py/partial-ssrf` | `py/log-injection` in `api/gateway.py` |
|---|---|---|
| main | 1 (2 flows) | 3 |
| after the auth gate | 1 (2 flows) | 4 — one I introduced |
| after the SSRF fix | 1 (**1 flow**) | 1 |
| now | 1 (1 flow) | **0** |

The SSRF alert dropping from **two flows to one** is the measurable confirmation that the
query-parameter override is gone — that second flow *was* the `?_url=` injection. The
log-injection alerts I introduced are gone because `_name` is no longer a query parameter, and the
last pre-existing one is fixed too (it interpolated the caller's path into the error log;
`zero-tolerance.md` Rule 1a says a PR-scan finding is treated identically to a main-scan one, so
"same on main" was not available).

### The remaining single flow

The caller's path landing in the URL's path component — which is what a reverse proxy does by
definition. It is bounded by four registration- and request-time controls: the auth gate, the
required path allowlist, traversal rejection, and the positive-charset barrier the URL is built
from.

**I am deliberately NOT filing another deferral, and I am not asserting this is a false positive.**
I asserted exactly that once already in this PR, with a written proof, and I was wrong — the alert
was a live full SSRF. My judgement on this specific flow has been demonstrated unreliable, so the
disposition belongs to the maintainer and the adversarial security review, not to me. No alert has
been dismissed in the Security tab.

One option I considered and did **not** implement speculatively, offered for the reviewer: when an
allowlist contains only literal (non-wildcard) entries, the forward could use the *matched
allowlist entry* — a server-chosen value — instead of the caller's path. That would make the
destination fully server-determined for those registrations, but does nothing for `["*"]`, and it
is real complexity that should be a deliberate decision rather than scanner appeasement.

## Review requirement

Per `agents.md` § "Correctness-Review-Clean Is Not Security-Clean", this change needs BOTH a correctness reviewer AND an adversarial security-reviewer prompted to refute, each returning a genuine ran-signal, before it converges. Neither has run on this branch yet.

## Related issues

Refs #2025


esperie:
## CodeQL `py/partial-ssrf` — deferred, NOT dismissed

Tracking issue: #2065. Reproduced here so the proof lives on the PR as `zero-tolerance.md` Rule 1b requires.

---

CodeQL alert #11497 (`py/partial-ssrf`, critical) is deferred on PR #2064 under `zero-tolerance.md` Rule 1b. This issue is the tracking artifact that rule requires; the runtime-safety proof is in a comment on PR #2064 and reproduced below.

## The finding

```
RULE: py/partial-ssrf   SEV: critical
FILE: src/kailash/servers/workflow_server.py:1001-1009
MSG:  Part of the URL of this request depends on a user-provided value.
      Part of the URL of this request depends on a user-provided value.
```

Two flows, not one: the captured **path** and the forwarded **query string**. Both feed `aiohttp.ClientSession.request(...)` in `WorkflowServer.proxy_workflow`'s handler.

Alert #11252 is the same finding on `main` before PR #2064 moved the code; #2028 reordered imports across 1,819 files, so line numbers in older references are stale. Anchor on the symbol `proxy_handler`, not the line.

## Why it is deferred rather than fixed

**Forwarding the caller's path and query IS the function of a reverse proxy.** The only way to make both flows constant is to stop proxying. `py/partial-ssrf` will flag any reverse proxy in any codebase; the query has no notion of "this endpoint is deliberately a proxy."

## Runtime-safety proof (guard lines)

All in `src/kailash/servers/workflow_server.py::proxy_workflow` and `src/kailash/servers/proxy_guard.py`.

1. **The authority is not attacker-influenced.** `_url=proxy_url` is closed over as a default argument at registration time. `proxy_workflow(name, proxy_url, ...)` is a Python API; its only two production callers (`channels/api_channel.py`, `api/gateway.py`) are Python APIs, not HTTP endpoints. Scheme, host and port are fixed before any request exists.

2. **No authority pivot is constructible — measured, not asserted.** The caller-controlled segment is appended after a `/`, so it lands in the path component:

   ```
   '@evil.com/x'   -> http://127.0.0.1:8080/@evil.com/x   | host: 127.0.0.1
   '//evil.com/x'  -> ValueError: Appending path '//evil.com/x' starting from slash is forbidden
   '..%2f..%2fetc' -> http://127.0.0.1:8080/..%252f..%252fetc | host: 127.0.0.1
   ```

   The host is unchanged in every case. This is why the CodeQL-implied remediation (validating `proxy_url`) closes nothing: `proxy_url` was never attacker-influenced.

3. **The route requires authentication.** Since PR #2064 `proxy_workflow` raises `ProxyAuthNotConfiguredError` at registration unless an auth control is supplied. An unauthenticated caller cannot reach the request at all.

4. **The path is constrained three ways before the URL is built**, in `proxy_handler`:
   - `reject_unsafe_proxy_path(path)` — refuses `..` segments, null bytes, control characters, backslashes, and encoded `%2e`/`%2f`/`%5c` → 400.
   - `path_matches_allowlist(path, _allowlist)` — refuses anything outside the allowlist supplied at registration → 404. The allowlist is required; there is no implicit catch-all.
   - `SAFE_FORWARD_PATH_RE.fullmatch(path)` — a positive charset barrier (RFC 3986 `pchar` + `/` + non-ASCII) → 400. **The target URL is built from the matched group, never the raw parameter.**

5. **The method surface is an allowlist** defaulting to `["GET"]`.

## What was tried and rejected

Building the URL through `yarl.URL(base) / segment` instead of string concatenation was evaluated. It expresses the authority invariant structurally, but measured against real inputs it **double-encodes `%`** (`a%2eb` → `a%252eb`), which changes behaviour for legitimately encoded paths. Since the concatenated form was already measured not to permit an authority pivot (point 2), that is a behavioural regression for no security gain. Rejected.

## Acceptance criteria for a full fix

- [ ] Either CodeQL models the path/query sanitizers above as barriers, or the alert is dismissed in the Security tab with this issue linked as the justification.
- [ ] If dismissal is chosen, record the dismissal reason as "used in tests"/"won't fix" with a link here, so the next scan does not re-raise it as new.
- [ ] Re-evaluate if the reverse-proxy feature is ever removed or reshaped, at which point this becomes moot.

## Signoff status

**NOT YET SIGNED OFF.** `zero-tolerance.md` Rule 1b requires release-specialist signoff (or user override) for a scanner deferral. That has not happened. Until it does, treat PR #2064 as carrying an unresolved scanner finding.

Note for context, stated because it is easy to misread: `main`'s branch protection currently has `required_status_checks.contexts: []` and `enforce_admins: false`, so this check is evidence-producing rather than merge-preventing. That is a fact about the repo's configuration, **not** a justification for the deferral — the justification is points 1–5 above.

esperie:
## CodeQL `py/partial-ssrf` #11497 — dismissal record (bounded residual)

Durable evidence for the Security-tab dismissal, which is capped at 280 characters. Independently verified at `c6b7d6592` by **two adversarial security reviewers and a release specialist**, each re-deriving rather than accepting a summary.

**What the alert flags:** the caller-controlled `path` segment reaching `aiohttp.ClientSession.request()`. Forwarding a caller path to a fixed backend is what a reverse proxy *is*; the question is whether the destination authority is server-determined. It is.

**1. Authority is fixed at registration, not attacker-influenced.**
The `?_url=` query-override that made this a **live full SSRF** was closed at `4a243fb17`. FastAPI treats any non-path handler parameter carrying a plain default as a query parameter, so `_url=proxy_url` was reachable as `?_url=`. Pinned by `test_destination_cannot_be_overridden_by_query_parameter`. Confirmed via the live alert object: #11497 went from **two source lines (`path` + `_url`) to one (`path`)** at that commit — the `_url` flow disappeared and nothing else changed.

**2. No authority pivot is constructible via the path segment.**
Re-verified with `yarl.URL()` against the shipped construction across `@evil.com/x`, `//evil.com/x`, `/@evil.com/x`, `..%2f..%2fetc`, `%2e%2e/etc` — host stays `127.0.0.1` in every case. The unconditional `/` separator in `f"{proxy_url.rstrip('/')}/{path}"` makes this structural, not incidental.

**3. Hop-2 redirect pivot closed at `c6b7d6592`.**
`aiohttp` defaults `allow_redirects=True`, so the controls below bounded only the first hop. Reproduced live before the fix — a backend open-redirect returned `{"AccessKeyId": "ASIA-PIVOTED-CREDENTIAL"}` to the caller with the `location` header hidden. Now `allow_redirects=False` (server) and an explicit `follow_redirects=False` (gateway); both surfaces verified to have no override path. Pinned by `test_backend_redirect_is_not_followed`.

**4. Four controls run unconditionally before the URL is built.**
Required auth (`ProxyAuthNotConfiguredError` at registration), registration-time method allowlist (405 at the router), registration-time path allowlist (deny-by-default), traversal rejection, and an RFC 3986 `pchar` positive charset barrier via `fullmatch` (anchored, not `search`).

**Not claimed:** that redirect-following is safe in general, or that this reasoning covers any other proxy surface. Both were checked and fixed separately. `#2085` tracks the unbounded-buffering / per-request-session asymmetry, deliberately not folded in.

Issue #2067, opened 2026-08-11
test_azure_ai_foundry_routes_through_four_axis_path is red on main — offline test needs env vars it never declares


esperie:
## Summary

`packages/kailash-kaizen/tests/regression/test_issue_1720_b1a_live_cutover.py::test_azure_ai_foundry_routes_through_four_axis_path` fails on any machine without `AZURE_AI_FOUNDRY_ENDPOINT` / `AZURE_AI_FOUNDRY_API_KEY` exported. It carries no `skipif`, so it is a hard failure rather than a skip.

Found while running the kaizen regression suite for #2030/#2022. **Not caused by that work** — see the falsification below.

## Evidence it is pre-existing

Reproduced against a clean `git archive` of `586062dbc` (the merge-base, i.e. main), with the #2030/#2022 changes confirmed absent (`grep -c unwrap_configuration_error` → `0`): `1 failed, 4 passed`. Neutralising both #2022 re-raise sites in the branch left it failing identically.

```
FAILED tests/regression/test_issue_1720_b1a_live_cutover.py::test_azure_ai_foundry_routes_through_four_axis_path
  - RuntimeError: azure_ai_foundry error (RuntimeError): Provider azure_ai_foundry is not available:
    Azure AI Foundry endpoint or API key unresolved:
    set $AZURE_AI_FOUNDRY_ENDPOINT and $AZURE_AI_FOUNDRY_API_KEY
```

Both vars are absent from this repo's `.env`.

## Why it is interesting rather than merely noisy

The test is deliberately OFFLINE — canned fixture via `CapturingTransport`, placeholder credentials. It builds its deployment with explicit overrides:

```python
deployment = resolve_deployment_for(
    "azure_ai_foundry", "gpt-5-nano",
    api_key="k-b1a-foundry-placeholder",
    base_url="https://my-foundry-resource.services.ai.azure.com",
)
assert deployment is not None      # passes
```

`_resolve_azure_ai_foundry_deployment` honours those overrides correctly. The failure comes later, from the agent call:

```python
agent._provider_llm_response(
    provider="azure_ai_foundry", model="gpt-5-nano", messages=_MSGS,
    tools=[], generation_config={}, api_key="k-b1a-foundry-placeholder",
)
```

which passes `api_key` but **no `base_url`**, so the re-resolution inside `_provider_llm_response` falls back to `AZURE_AI_FOUNDRY_ENDPOINT` and finds nothing.

[...]

Issue #1947, opened 2026-07-24
fix(kaizen): LLMAgentNode provider defaults to "mock" — make it fail-loud to close the silent-mock class structurally


esperie:
## Summary
The root cause behind #1943 and its follow-ups: `LLMAgentNode.get_parameters()` declares `provider` as `required=False, default="mock"`. Any construction site that omits `provider` therefore **silently dispatches to the mock provider** and returns fabricated content as if real. #1943 fixed every Agent-surface site individually, but the default itself is the structural hazard — every new/forgotten site re-opens the class.

## Proposed fix
Make a missing/unset provider **fail loudly** in a real (non-test) context instead of silently defaulting to `"mock"` — e.g. `required=True`, or default `None` + raise a typed error at `run()` when unresolved. This structurally closes the class: a forgotten provider becomes a loud error, not silent fake data.

## Prerequisites (gating — high blast radius)
This change breaks every site that relies on the mock default. It MUST land only after:
1. The RAG provider sweep (prerequisite issue: RAG nodes set provider).
2. A test sweep: ~12 test files construct `LLMAgentNode` relying on the mock default → make each explicit (`provider="mock"` or a shared test fixture).

## Acceptance criteria
- [ ] RAG sweep + test sweep complete (prerequisites).
- [ ] `LLMAgentNode` raises a typed error (not silent mock) when provider is unresolved in a real context.
- [ ] Full package test sweep green; kaizen minor release.

## Cross-SDK
Per cross-sdk-inspection: the Rust SDK's equivalent node should be checked for the same mock-default hazard (human-gated cross-repo filing).

esperie:
## Blast-radius scoping (prerequisite test sweep)

The fail-loud flip cannot land until every site relying on the `provider="mock"` default is made explicit. Precise blast radius (kaizen package):

[...]

PR #2068, opened 2026-08-11
fix(kaizen): stop logging agent I/O at INFO, and stop reporting configuration failures as bad model output (#2030, #2022)

### LINKED ISSUE #2022: kailash-ml: km.from_brief fails end-to-end — BaseAgentConfig built without llm_provider (README Quick Start broken)

### LINKED ISSUE #2030: HIGH: BaseAgent logs full agent inputs and outputs at INFO by default
## Summary

`BaseAgent` writes the **complete inputs and complete result** of every agent execution to the log at INFO, on by default. Agent I/O routinely carries user prompts, retrieved documents, PII, and — for agents that take credentials as parameters — secrets.

## Evidence

`packages/kailash-kaizen/src/kaizen/core/base_agent.py:724-735`:

```python
        logging_enabled = getattr(self.config, "logging_enabled", True)
        if logging_enabled:
            signature_name = getattr(self.signature, "name", "unknown")
            logger.info(f"Executing {signature_name} with inputs: {inputs}")
        return inputs

    def _post_execution_hook(self, result: Dict[str, Any]) -> Dict[str, Any]:
        """Default post-execution hook (logs completion)."""
        logging_enabled = getattr(self.config, "logging_enabled", True)
        if logging_enabled:
            logger.info(f"Execution complete. Result: {result}")
        return result
```

The gate is `getattr(self.config, "logging_enabled", True)` — **defaults to True**, so this is the behaviour for any agent whose config does not explicitly opt out. Both are INFO, so they survive normal production log levels and ship to whatever aggregator is configured.

This violates `observability.md` Rule 4 (never log secrets or PII) and its MUST NOT on unstructured f-string interpolation — `{inputs}` and `{result}` are whole dicts rendered into a message string, so they cannot be filtered or redacted field-by-field downstream.

## How this was found, and why it matters for the scanner

CodeQL alert #11244 flagged **line 744** in the same function block:

```python
            logger.error(f"Error during execution: {error}", extra=context)
```

That line is a **generic pass-through sink** and is a false positive — it has no sensitive value of its own; the query taints it from some caller. Lines 727 and 735, which unconditionally render full agent I/O, were **not flagged at all**.

So a triage that works the CodeQL queue closes the flagged line, records the file as reviewed, and leaves the two genuinely dangerous lines untouched. Worth noting explicitly because the same shape recurred repeatedly in this repo's recent triage: the instrument reports on what its queries model, not on what is risky.

## Acceptance criteria

- [ ] Neither the full `inputs` dict nor the full `result` dict is emitted at INFO.
- [ ] If execution tracing is wanted, emit structured fields — signature name, input **keys**, result **keys**, counts, durations — not the values. (`parameter_injector.py:423` and `:445` are the correct in-repo precedent: they log `list(...keys())`.)
- [ ] Any full-payload logging moves to DEBUG **and** routes through the credential scrubber
      (`kaizen/utils/credential_scrub.py`), rather than relying on operators keeping DEBUG off.
- [ ] Regression test: run an agent whose inputs contain a credential-shaped value and assert it does not appear in `caplog` at INFO.
- [ ] Decide whether `logging_enabled` defaulting to True is right for a full-payload logger. Per `security.md` § Secure-Default, a feature whose default exposes data should fail closed or warn loudly.

## Related

- `_handle_error` immediately below (`:741-747`) has a separate defect tracked in #2022: `error_handling_enabled` also defaults True, so it swallows the exception into `{"success": False}` — which is why a `ConfigurationError` surfaces to users as "the LLM emitted a malformed plan".
- CodeQL #11244 should be dismissed as a false positive **only alongside** this fix, so the file is not recorded as reviewed-and-clean while lines 727/735 stand.

esperie:
Closes two issues that share `base_agent.py` and could not be split.

## Every log site found, and how each was disposed of

The issue names two sites. I read the whole module and swept the execution path rather than letting the alert locations define scope — which is the trap #2030 itself documents.

| Site | Level | Disposition |
|---|---|---|
| `base_agent.py::_pre_execution_hook` | INFO | Now emits signature name + input KEY names + count. Values never rendered. |
| `base_agent.py::_post_execution_hook` | INFO | Now emits result KEY names + count. |
| `base_agent.py::_handle_error` — `extra=context` | **ERROR** | **Was a real leak, and was NOT in the issue.** `AgentLoop` calls it as `_handle_error(error, {"inputs": inputs})`, so the full inputs went onto the LogRecord at a level that survives every production config. Now summarized via `safe_log_extra`. |
| `base_agent.py::_handle_error` — returned dict | n/a | Logged message was scrubbed but the RETURNED one was not; that dict is the public return value of `agent.run()`. Scrubbed once, used for both. |
| `mixins/logging.py::log_execution_start/end` (public `LoggingMixin`) | INFO | Full inputs, and full result in the JSON branch — byte-for-byte the pre-fix BaseAgent behaviour, ungated and unscrubbed. Now summarized. Distinct class from the already-clean `core/mixins/logging_mixin.py`. |
| `core/agents.py::_extract_intelligent_response` | WARNING | Logged the entire LLM response 20 lines below a sibling branch that already fingerprints for this exact reason. Raw moved to DEBUG; WARNING gets the fingerprint. |
| `nodes/ai/llm_agent.py` truncated tool args | ERROR | **Not actioned** — see "Not verified / not done". |
| `hooks/builtin/logging_hook.py` | INFO | **Not actioned** — see below. |
| Swept clean | — | `agent_loop.py`, `workflow_generator.py`, `llm/_legacy_shape.py`, `core/mixins/logging_mixin.py`, `strategies/*` logging, `observability/audit.py`, remaining `base_agent.py` sites. |

Full payloads are not deleted, they are gated: an opt-in `log_full_payloads` (default **False**) AND DEBUG AND `scrub_credentials`. Default-closed because turning DEBUG on globally is routine during incident response (`rules/security.md` § Secure-Default). `logging_enabled` keeps defaulting True since it now gates a metadata-only line.

**A finding against my own first cut:** `summarize_payload` originally claimed key names are always schema. That is false on the RESULT path — a strategy returns the model's parsed JSON verbatim when no signature field matches, and `_validate_signature_output` returns early on a `response` key without inspecting the others, so result keys can be **model-controlled** and thus prompt-injectable. Key names are now truncated, capped in number (bounding log amplification from a 10k-key result), and scrubbed.

### CodeQL #11244 coupling

Alert #11244 flags `logger.error(f"Error during execution: {error}", extra=context)`. **It should not be dismissed as a false positive.** The issue reasoned it is "a generic pass-through with no sensitive value of its own" — but `extra=context` carries the full agent inputs, so the line is a genuine leak, just not for the reason CodeQL modelled. It is fixed here. Dismiss only alongside this merge, and as *fixed*, not as *false positive*. I have not touched any alert.

## Canonical-resolver decision (#2022 AC 3)

There are not three resolvers — there are **six** model→provider mappings. The decision:

| Question | Canonical | Status |
|---|---|---|
| model → provider | **`kaizen.llm.LlmProvider.from_model`** | already public; prefix table DERIVED from the provider registry, so it cannot drift; fail-closed |
| environment → provider | `detect_provider_from_env` | unchanged |
| full config → provider + credentials | `kaizen.config.auto_detect_provider` | unchanged |

`kaizen.nodes._env_model.detect_provider` is **deliberately NOT published**, against the issue's suggested option (a). Publishing it would add a *second* public answer to a question a public API already answers, with a weaker algorithm — a hand-maintained substring table pinned to no registry, versus a registry-derived prefix table. Its docstring now records that, and why it survives (it covers Ollama-family names the registry has no row for, and carries a `component=` diagnostic). `providers/registry.py::get_provider_for_model` was already retired in #1720 to end exactly this duplication; re-opening it would undo that.

New public surface: **`kaizen.core.resolve_agent_provider`**, which **adds no mapping of its own** — it only orders the two canonical resolvers. Model-keyed first, which also closes a mismatch class (a Claude `DEFAULT_LLM_MODEL` was dispatched to OpenAI whenever `OPENAI_API_KEY` happened to be set), then the env fallback, then a loud failure. Stated plainly: for a model outside the registry the env fallback is still a guess — unchanged from what every caller already got — but every registry-recognised model now bypasses it.

## The #2022 defect is not what the title says, and the swallow is not where the brief said

The brief said the swallow is a `try/except` in `base_agent.py`. It is in three places, and `base_agent.py` is the shallowest:

1. **`LocalRuntime` does not raise on node failure at all.** It returns `{node_id: {"failed": True, "_exception": <exc>}}`, so a `try/except` around `runtime.execute` never fires. This is the deepest swallow and the reason the sync strategies never saw an exception.
2. **The strategies** then classify by **substring matching on the message** (`"api error"`, `"api key"`, `"401"`, `"not available"`). A `ConfigurationError` arrives wrapped by the runtime as `WorkflowExecutionError` and matches none of them, so it fell through to `return {"error": ..., "status": "failed"}` — an exception converted into a dict, losing type and traceback.
3. **`_handle_error`** folded the resulting `ValueError("Missing required output field")` into `{"success": False}`.

So "the LLM emitted a plan that does not conform to MLPlan" was **two laundering steps** from the real cause. All fixed through one shared predicate in `kaizen/errors.py`, applied at all five `runtime.execute` sites across all three strategies plus `_handle_error`. `_exception` preserves the exception OBJECT, so the check stays structural — an isinstance walk over the cause chain, not substring matching, which is what missed this class in the first place.

`unwrap_configuration_error` follows `__cause__` **only**, not `__context__`: `__context__` is implicit, so following it would let an unrelated failure inside an `except ConfigurationError:` block be *replaced* by that ConfigurationError. Every wrapping site here uses `raise ... from`, so nothing is lost.

Both `from_brief` call sites now pass `llm_provider` — the issue names only kailash-ml, but **kailash-dataflow had the identical omission and failure mode**. Both also gained an `llm_provider` kwarg, because the error message told callers to pass one that did not exist.

## Fail-first evidence, verbatim

RED against **main's source** (`586062dbc`, pinned via PYTHONPATH; interpreter asserted in-process to be loading the main checkout):

```
=================== 10 failed, 1 passed, 1 warning in 12.29s ===================
FAILED TestHandleErrorDoesNotSwallowConfigurationErrors::test_configuration_error_propagates - Failed: DID NOT RAISE <class 'kaizen.config.providers.ConfigurationError'>
FAILED TestUnresolvedProviderSurfacesThroughRun::test_run_raises_configuration_error_not_empty_result - Failed: DID NOT RAISE <class 'kaizen.config.providers.ConfigurationError'>
FAILED TestResolveAgentProvider::test_model_keyed_wins_over_environment - ImportError: cannot import name 'resolve_agent_provider' from 'kaizen.core'
```

RED for #2030 against the branch before the fix:

```
==================== 8 failed, 6 passed, 1 warning in 0.75s ====================
FAILED TestPreExecutionHookInfoLevel::test_input_values_absent_at_info - AssertionError: credential value leaked at INFO
FAILED TestPostExecutionHookInfoLevel::test_result_values_absent_at_info - AssertionError: credential value leaked at INFO
FAILED TestHandleErrorDoesNotLeakContext::test_inputs_absent_from_error_record - AssertionError: credential leaked via extra= at ERROR
FAILED TestFullPayloadIsOptInAndScrubbed::test_debug_alone_does_not_dump_payloads - AssertionError: DEBUG alone must not dump payloads
```

GREEN:

```
======================== 53 passed, 1 warning in 15.22s ========================   # both regression files + slimming guard
5 passed, 2 warnings in 0.39s                                                      # kailash-ml
7 passed, 3702 deselected in 2.97s                                                 # kailash-dataflow
============ 1369 passed, 2 skipped, 1 warning in 233.41s (0:03:53) ============   # tests/unit/{strategies,core,mixins}
= 1 failed, 1608 passed, 1 skipped, 24 xfailed, 1 warning in 111.51s (0:01:51) =   # tests/regression — see below
```

**Instrument note.** There is no `.venv` in the worktree and the available interpreter's editable installs point at the MAIN checkout, so a naive `pytest` run silently tests the wrong tree and passes vacuously. Every run above pinned `PYTHONPATH` and asserted `kaizen.__file__` inside the worktree **in the same process** as the test session.

Redaction tests pin **both polarities**: credential- and PII-shaped values MUST be absent, AND signature name / key names / counts / error message MUST be preserved — so deleting the log statements cannot pass. The assertion helper renders `extra=` attributes as well as the message; reading `getMessage()` alone would have reported green while the ERROR leak stood.

Mutation-checked rather than assumed: reverting the ml `from_brief` change reddens its test; neutralising each re-raise site reddens the matching tests.

## The one failing test is pre-existing — now #2067

`test_issue_1720_b1a_live_cutover.py::test_azure_ai_foundry_routes_through_four_axis_path` fails without `AZURE_AI_FOUNDRY_ENDPOINT`/`_API_KEY`, which are absent from this repo's `.env`. Reproduced red against a clean `git archive 586062dbc` with these changes confirmed absent. Filed as #2067 rather than silenced: a `skipif` would turn it green while hiding whether `_provider_llm_response` genuinely cannot accept a per-call `base_url`.

## Not verified / not done

- **`hooks/builtin/logging_hook.py`** logs full hook payloads at INFO with `include_data=True, redact_sensitive=False`, and `AgentLoop` feeds it the full inputs and results. Opt-in (`hooks_enabled` defaults False) and it belongs to the hooks subsystem, so it is outside this shard — but it is the same disclosure class and its `redact_sensitive=False` default is the same fail-open posture this PR argues against.
- **`agent_config.py::_detect_provider_from_model`** ends `else: return "openai"` — a silent fail-open that routes an unknown model to the wrong vendor (`zero-tolerance.md` Rule 3). Same class as #2022. Not fixed here: it feeds a legacy config surface with no test coverage, and `resolve_agent_provider` can return `deepseek`, which is not in that class's `VALID_PROVIDERS` — so it needs its own shard.
- **`llm_agent.py`** truncated tool-argument logging at ERROR (unscrubbed, 100 chars, model-generated) — same class, small blast radius, not actioned.
- **No live-LLM verification.** The `.env` `OPENAI_API_KEY` is invalid (401). Everything here is exercised offline; no test in this PR requires a provider.
- Two full suites (`kaizen-agents/tests/unit`, `kailash-ml/tests/unit`) stalled on infra-dependent tests and were killed. That is zero evidence, not a pass; the focused subsets above are what actually ran.

## Related issues

Fixes #2030
Fixes #2022
Refs #2067


esperie:
Follow-up shards from the review of this PR are now filed, so the "Not verified / not done" section has tracking rather than prose:

- **#2069** — `AgentConfig._detect_provider_from_model` fail-open (`else: return "openai"`). Same class as #2022. Includes an integration wrinkle: `resolve_agent_provider` can return `"deepseek"`, absent from that class's `VALID_PROVIDERS`.
- **#2070** — `LoggingHook` logs full agent payloads at INFO with `redact_sensitive=False`, **plus** a second defect neither the PR body nor I had listed: an operator who explicitly opts INTO redaction has it silently turned off on `ImportError` (`:74-86`) and payload logging continues.

One correction to the record, verified and worth stating because it would otherwise send someone down a wrong causal chain: it was proposed that the `agent_config.py` fail-open explains the unconfirmed keyed-path half of #2022's caveat. **It does not.** `kailash_ml/from_brief.py:715,756` uses `BaseAgentConfig` (`kaizen/core/config.py:38`), whereas `_detect_provider_from_model` is a method of `AgentConfig` (`kaizen/agent_config.py:15`) — a different class — and `BaseAgentConfig.__post_init__` performs no provider auto-detection at all. That is precisely why the provider stayed `None` and reached the #1947 node gate. #2022's keyed-path caveat remains unexplained and still needs a live-LLM reproduction to settle.

Credit: both shards were found by a parallel read-only review pass (@L2-kaizen-agents), which also independently reproduced the `extra=context` leak documented above.

Issue #2069, opened 2026-08-11
AgentConfig._detect_provider_from_model fails open — unknown model silently routed to OpenAI


esperie:
## Summary

`packages/kailash-kaizen/src/kaizen/agent_config.py:309-334` — `AgentConfig._detect_provider_from_model` ends with a terminal `else` that silently returns `"openai"` for any model it does not recognise:

```python
        elif "gemini" in model_lower:
            return "google"
        else:
            # Default to openai for unknown models
            return "openai"          # :334
```

This is a `rules/zero-tolerance.md` Rule 3 silent fallback. An unrecognised model is dispatched to the wrong vendor under whatever credential happens to be configured, and the caller is never told.

Surfaced during #2030/#2022 review. Evidence gathered by @L2-kaizen-agents; the correction and integration note below are mine.

## Why it is the same class as #2022

#2022's PR (#2068) established `kaizen.core.resolve_agent_provider` as the public provider-resolution surface, and deliberately did NOT publish `kaizen.nodes._env_model.detect_provider` — because a hand-maintained substring table pinned to no registry drifts, whereas `kaizen.llm.LlmProvider.from_model`'s prefix table is DERIVED from the provider registry. `_detect_provider_from_model` is another hand-maintained substring table, and the one that additionally fails OPEN.

## The asymmetry worth pinning in a test

Within the same `__post_init__`, an **explicit** bad provider raises but an **auto-detected** one silently guesses:

```python
        if self.llm_provider.lower() not in self.VALID_PROVIDERS:      # :299
            raise ValueError(f"Invalid llm_provider: '{self.llm_provider}'. ...")
        ...
        if self.llm_provider is None:                                   # :306
            self.llm_provider = self._detect_provider_from_model(self.model)
```

Same question, two opposite dispositions.

## CORRECTION — this is NOT on `from_brief`'s path

An earlier reading proposed this as the mechanism behind the unconfirmed half of #2022's caveat (`from_brief` → `:307` → `"openai"` regardless). **That does not hold, and the record should not carry it.** Verified:

- `kailash_ml/from_brief.py:715,756` uses **`BaseAgentConfig`** (`kaizen/core/config.py:38`)
- `_detect_provider_from_model` is a method of **`AgentConfig`** (`kaizen/agent_config.py:15`) — a different class
- `BaseAgentConfig.__post_init__` performs **no** provider auto-detection (that is exactly why the provider stayed `None` and reached the #1947 node gate)

[...]

PR #2073, opened 2026-08-11
fix(kaizen): LoggingHook payload disclosure at INFO + explicit-redaction-opt-in defeat (#2070)

### LINKED ISSUE #2070: HIGH: LoggingHook logs full agent payloads at INFO by default, and an explicit opt-in to redaction degrades to OFF
## Summary

`packages/kailash-kaizen/src/kaizen/core/autonomy/hooks/builtin/logging_hook.py` carries the #2030 disclosure at a second surface, plus a distinct fail-open in the security control meant to mitigate it.

Surfaced during #2030/#2022 review (PR #2068), which fixed the `BaseAgent` and `LoggingMixin` sites but deliberately left this one as its own shard. Evidence gathered by @L2-kaizen-agents.

## Defect 1 — full payloads at INFO, unredacted, by default

`:44-49`:

```python
        log_level: str = "INFO",
        include_data: bool = True,
        format: str = "text",
        redact_sensitive: bool = False,
```

Consumed at `:150` (`log_event["context"] = safe_context.data`) and `:162-166` (`f"Data={safe_context.data}"`).

`agent_loop.py` feeds this hook the whole agent I/O — `PRE_AGENT_LOOP` with `{"inputs": processed_inputs, ...}` (`:450-458`, async twin `:541-549`) and `POST_AGENT_LOOP` with `{"result": final_result, ...}` (`:470-478`, `:561-569`). So when the hook is registered, full agent inputs and results reach the log at INFO — the exact defect #2030 describes, with `redact_sensitive` defaulting OFF.

Mitigating: `hooks_enabled` defaults False (`core/config.py`) and `HookManager` registers nothing by default, so this is opt-in — which is why it is HIGH rather than CRITICAL. Aggravating: PR #2068 argued the opposite default for `log_full_payloads` on `rules/security.md` § Secure-Default grounds. Same repo, same disclosure class, opposite dispositions.

## Defect 2 — an explicitly-requested security control silently degrades to off

`:74-86`:

```python
        if self.redact_sensitive:
            try:
                from ..security.redaction import SensitiveDataRedactor
                self.redactor = SensitiveDataRedactor()
            except ImportError:
                logger.warning(
                    "SensitiveDataRedactor not available, disabling redaction. ..."
                )
                self.redactor = None
                self.redact_sensitive = False      # :85
```

An operator who **explicitly opted in** to redaction has it turned off for them, and payload logging continues at full fidelity. The consumer of that degraded state is `:134` (`if self.redact_sensitive and self.redactor:`).

It WARNs rather than failing silently, so it is a rung above a fully silent fallback — but the disposition is the same class: an explicit opt-in to a security control must fail CLOSED (raise), not continue emitting unredacted data behind a warning an operator may never see. `rules/zero-tolerance.md` Rule 3 + `rules/security.md` § Secure-Default.

## Note on the redactor itself

If this is wired to `SensitiveDataRedactor` (`hooks/security/redaction.py`), check it against `kaizen/utils/credential_scrub.py` first — that module's own docstring records the redactor having drifted on 9 of 10 measured vendor shapes. `scrub_credentials` is the single credential-scrub implementation and is what PR #2068 routed the BaseAgent DEBUG path through.

## Acceptance criteria

- [ ] Default construction does not emit agent input/result VALUES at INFO (flip `include_data`, or default `redact_sensitive=True`, or summarize to key names as PR #2068 did for `BaseAgent` / `LoggingMixin`).
- [ ] `:85` raises rather than degrading when redaction was explicitly requested and the redactor cannot be loaded.
- [ ] Fail-first test, BOTH polarities: a credential-shaped value in `context.data` is ABSENT from `caplog` at INFO under default construction, AND the structured metadata that makes the log useful is PRESERVED (so blanket-masking cannot pass).
- [ ] Negative control: `redact_sensitive=True` with the redactor unavailable RAISES rather than logging unredacted.
- [ ] Decide whether to route through `scrub_credentials` rather than `SensitiveDataRedactor`.

## Related

- #2030 / PR #2068 — same disclosure class, fixed at `BaseAgent` and the public `LoggingMixin`.
- #2069 — the other same-class shard split out of that review.

esperie:
## Summary

Two defects in `packages/kailash-kaizen/src/kaizen/core/autonomy/hooks/builtin/logging_hook.py`.

**Defect 1 — full agent payloads at INFO, unredacted, by default.** `agent_loop` feeds this hook whole agent I/O (`PRE_AGENT_LOOP` carries `{"inputs": ...}`, `POST_AGENT_LOOP` carries `{"result": ...}`), and the hook rendered that dict straight into an INFO line. `include_data` defaulted True, `redact_sensitive` defaulted False. Same disclosure class as #2030, at a second surface.

**Defect 2 — an explicitly-requested security control degraded to OFF.** On `ImportError` the constructor set `redact_sensitive = False` and kept logging at full fidelity behind a WARNING. An operator who deliberately turned redaction ON had it turned off for them. `zero-tolerance.md` Rule 3 + `security.md` § Secure-Default.

## Fail-first evidence

RED against unmodified source — 6 failed, 1 passed (the 1 is the negative control, which must pass in both states):

```
INFO  logging_hook.py:163 [pre_agent_loop] Agent=agent-2070 TraceID=trace-2070
      Data={'inputs': {'prompt': 'summarize the acquisition memo for board review',
      'api_key': 'sk-live-LOGGINGHOOK2070SECRETVALUE'}}

WARNING logging_hook.py:80 SensitiveDataRedactor not available, disabling redaction.
Failed: DID NOT RAISE <class 'RuntimeError'>
```

GREEN after: **7 passed**. Full hook suite: **126 passed**.

## The fix

- **INFO logs structure, never values** — event, agent, trace, payload KEY names + count. Redaction is not a substitute: it claims credential shapes and PII patterns, and a user's prose matches neither.
- **Values only behind two independent gates** — new `log_full_payloads` (default False) AND the logger actually at DEBUG, routed through `scrub_credentials`, truncated at 2000 chars before scrubbing.
- **`redact_sensitive` is tri-state** (`Optional[bool]`), and that is the load-bearing part. An explicit `True` that cannot be honoured **raises**; an unrequested default may degrade **but then also stops emitting payloads**. Raising on a default nobody asked for would break installs where the redactor is genuinely absent; continuing to log after the control failed is the disclosure the control existed to prevent.

## Both polarities pinned

| Test | Property |
|---|---|
| `test_credential_absent_from_info` | credential-shaped value withheld at INFO |
| `test_user_prompt_absent_from_info` | **prose** withheld — not scrubber-recognisable, so only structure-only logging passes |
| `test_structure_is_preserved` | **negative control** — blanket-masking / deleting the log cannot pass |
| `test_raises_when_redactor_unavailable` | explicit opt-in fails closed |
| `test_error_names_the_protection_and_the_remedy` | error names what failed and how to fix |
| `test_does_not_raise_and_still_withholds_values` | default degrades **without** disclosing |
| `test_credential_redacted_in_full_payload_mode` | positive polarity — redactor really runs |

Tests 3/4 force a **genuine** `ImportError` through `sys.modules` rather than reasoning about the branch, so the constructor really executes it. No `Mock`.

## Why existing coverage missed this

Twelve `LoggingHook` constructions across `test_builtin_hooks.py` and `test_structured_logging.py` assert only `result.success is True` / `result.error is None`. Not one inspects log content — including `test_text_format_unchanged`, whose docstring says *"produces same output as before"* while asserting nothing about output. Those tests pass identically whether the payload is logged, redacted, or omitted.

## Scope / reachability, stated rather than glossed

`security/redaction.py` imports only stdlib, `kaizen.utils.credential_scrub`, and siblings, so its `ImportError` branch is **not reachable through a normal install today**. Defect 2 is a live fail-open *disposition* on an unreachable path, not a live leak. Defect 1 is live whenever the hook is registered (mitigated by `hooks_enabled` defaulting False).

## Verification

- Fail-first RED quoted above, then GREEN — 7 passed
- `tests/unit/core/autonomy/hooks/` — 126 passed, no regressions
- `pre-commit run --files <changed>` — all hooks pass, incl. the Tier 1 gate
- 3 pre-existing failures in `tests/e2e/autonomy/test_hooks_e2e.py` are **not** caused by this change: reverting the source to HEAD reproduces them identically. They are `@pytest.mark.ollama` and need `OPENAI_API_KEY`; the agent errors on the missing provider so POST never fires, hence `assert 1 == 2`.

## Related issues

Closes #2070
Related: #2030 / PR #2068 (same class, `BaseAgent` + `LoggingMixin`), #2069 (sibling shard)

### DIFF
--- a/packages/kailash-kaizen/CHANGELOG.md
+++ b/packages/kailash-kaizen/CHANGELOG.md
@@ -13,6 +13,52 @@ range such as `>=2.0`.
 
 ## [Unreleased]
 
+### Changed (BREAKING) — the four observability flags no longer silently install nothing (#2084)
+
+`AgentConfig.enable_tracing` / `enable_metrics` / `enable_logging` / `enable_audit`
+each **defaulted to `True`** while the subsystems they enable **do not exist**.
+`create_observability()` imported four hook classes that are not in the package,
+caught the resulting `ImportError`, logged `"… not available, skipping"`, and
+returned a `HookManager` with **zero hooks registered**. `enable_audit` is the
+sharp case: compliance audit trails that silently record nothing are invisible
+exactly when they matter.
+
+The flags are now **tri-state** (`Optional[bool]`, default `None`):
+
+| value             | behaviour                                                        |
+| ----------------- | ---------------------------------------------------------------- |
+| `None` (default)  | one consolidated WARNING naming the unimplemented subsystems; otherwise unchanged |
+| `True` (explicit) | raises `kaizen.errors.ObservabilityNotImplemented`, naming the subsystem and flag |
+| `False`           | silent                                                            |
+
+Raising only on an explicit `True` is deliberate: `True` was also the default, so
+raising on the default would break every agent construction — including callers
+who never mentioned observability.
+
+**Also corrected: `AgentConfig.is_observability_enabled()`.** Previously returned
+`True` by default while zero hooks were registered — it reported observability the
+framework did not have, so a caller branching on it ran its observability-dependent
+path against an empty manager. It now returns `False` unless a subsystem is
+explicitly requested. This is a correction from false to true, not a change from
+one correct behaviour to another.
+
+`create_observability()` now returns `None` on the default path instead of an empty
+`HookManager`. `BaseAgent` already declares `hook_manager: Optional[Any] = None` and
+handles it (`base_agent.py:184-192`), constructing its own manager when
+`hooks_enabled` — so the observable behaviour is unchanged.
+
+**Migration.** If you pass any of these flags explicitly:
+
+- **Set them to `False`**, or remove them — they never did anything.
+- Do **not** set them to `True`: that now raises, because the feature does not exist.
+- If you were relying on `is_observability_enabled()` returning `True`, note it was
+  reporting a subsystem set that registered nothing; supply your own manager via
+  `custom_hook_manager=` to get real hooks.
+- To pin the previous (silent) behaviour: `kailash-kaizen==<previous exact version>`.
+
+Whether these four subsystems get implemented or removed is tracked in **#2084**;
+this change makes the flags honest, it does not make the features exist.
+
 ### Fixed — `EnterpriseMemorySystem` no longer widens tenant scope on a falsy tenant id (#2005)
 
 `_build_tenant_key` resolved its scope with `tenant_id or self._current_tenant`.
--- a/packages/kailash-kaizen/src/kaizen/agent.py
+++ b/packages/kailash-kaizen/src/kaizen/agent.py
@@ -84,10 +84,16 @@ class Agent:
         # Tools
         tools: Union[str, List[str], None] = "all",
         # Observability
-        enable_tracing: bool = True,
-        enable_metrics: bool = True,
-        enable_logging: bool = True,
-        enable_audit: bool = True,
+        # TRI-STATE, mirroring AgentConfig (#2084). These MUST default None,
+        # not True: they are forwarded verbatim to AgentConfig below, and a
+        # `True` there means "explicitly requested" and raises because none of
+        # the four subsystems exists. Leaving these at `True` would send an
+        # explicit True on EVERY Agent() construction — reintroducing, through
+        # this wrapper, exactly the break the tri-state exists to avoid.
+        enable_tracing: Optional[bool] = None,
+        enable_metrics: Optional[bool] = None,
+        enable_logging: Optional[bool] = None,
+        enable_audit: Optional[bool] = None,
         # Checkpointing
         enable_checkpointing: bool = True,
         checkpoint_path: str = ".kaizen/checkpoints",
--- a/packages/kailash-kaizen/src/kaizen/agent_config.py
+++ b/packages/kailash-kaizen/src/kaizen/agent_config.py
@@ -114,26 +114,37 @@ class AgentConfig:
     # LAYER 2: Observability Configuration
     # =========================================================================
 
-    enable_tracing: bool = True
-    """Enable distributed tracing (Jaeger)"""
+    # TRI-STATE, not bool (#2084). None of these four subsystems is
+    # implemented, so an explicit `True` must fail loudly — but `True` was
+    # ALSO the default, and raising on the default would break every agent
+    # construction including callers who never mentioned observability. A
+    # dataclass cannot tell "explicitly passed True" from "defaulted True",
+    # so the sentinel is required:
+    #   None  -> nobody asked; warn, behave as before
+    #   True  -> explicit request; raise ObservabilityNotImplemented
+    #   False -> explicit opt-out; silent
+    # Same mechanism as `LoggingHook.redact_sensitive` (#2070).
+
+    enable_tracing: Optional[bool] = None
+    """Enable distributed tracing (Jaeger). NOT IMPLEMENTED — see #2084."""
 
     tracing_endpoint: str = "http://localhost:16686"
     """Jaeger tracing endpoint"""
 
-    enable_metrics: bool = True
-    """Enable Prometheus metrics collection"""
+    enable_metrics: Optional[bool] = None
+    """Enable Prometheus metrics collection. NOT IMPLEMENTED — see #2084."""
 
     metrics_port: int = 9090
     """Prometheus metrics port"""
 
-    enable_logging: bool = True
-    """Enable structured JSON logging"""
+    enable_logging: Optional[bool] = None
+    """Enable structured JSON logging. NOT IMPLEMENTED — see #2084."""
 
     log_level: str = "INFO"
     """Logging level (DEBUG, INFO, WARNING, ERROR)"""
 
-    enable_audit: bool = True
-    """Enable compliance audit trails"""
+    enable_audit: Optional[bool] = None
+    """Enable compliance audit trails. NOT IMPLEMENTED — see #2084."""
 
     audit_log_path: str = ".kaizen/audit.jsonl"
     """Audit log file path"""
@@ -365,12 +376,28 @@ class AgentConfig:
         )
 
     def is_observability_enabled(self) -> bool:
-        """Check if observability is enabled (either default or custom)."""
-        return self.has_custom_observability() or (
-            self.enable_tracing
-            or self.enable_metrics
-            or self.enable_logging
-            or self.enable_audit
+        """Check if observability is enabled (either default or custom).
+
+        CORRECTED in #2084. This previously returned True by default — while
+        `create_observability` registered ZERO hooks, because none of the four
+        subsystems exists. It reported observability the framework did not
+        have, so a caller branching on it ran its observability-dependent path
+        against an empty manager.
+
+        Now only an EXPLICIT `True` counts. `None` (the default, nobody asked)
+        and `False` (opt-out) both mean not-enabled, so the answer matches what
+        is actually registered.
+        """
+        if self.has_custom_observability():
+            return True
+        return any(
+            flag is True
+            for flag in (
+                self.enable_tracing,
+                self.enable_metrics,
+                self.enable_logging,
+                self.enable_audit,
+            )
         )
 
     def is_checkpointing_enabled(self) -> bool:
--- a/packages/kailash-kaizen/src/kaizen/core/autonomy/hooks/_payload.py
+++ b/packages/kailash-kaizen/src/kaizen/core/autonomy/hooks/_payload.py
@@ -1,0 +1,62 @@
+# Copyright 2026 Terrene Foundation
+# SPDX-License-Identifier: Apache-2.0
+
+"""Shared payload-logging helpers for hooks (#2070).
+
+ONE implementation, because there were two. `LoggingHook`
+(`builtin/logging_hook.py`) and `SecureLoggingHook` (`security/redaction.py`)
+carried near-verbatim copies of the same log body, and the copies had already
+diverged in exactly the way duplication predicts: the fix for one would have
+left the other publishing payloads under a class name promising the opposite.
+
+The contract both share:
+
+- At the hook's configured level, emit STRUCTURE — key names and counts. Key
+  names come from the signature / event schema rather than from user input, so
+  they carry no payload content.
+- Emit VALUES only behind two independent gates: a deliberate opt-in AND the
+  logger actually at DEBUG. Either gate alone is routinely satisfied by
+  accident — turning DEBUG on globally is routine during incident response.
+- Scrub what does get emitted. Scrubbing claims credential SHAPES, not
+  arbitrary prose, which is why the opt-in gate is the load-bearing protection
+  and the scrub is defence in depth rather than the other way round.
+"""
+
+from __future__ import annotations
+
+import logging
+from typing import Any
+
+#: Cap on a rendered payload before it reaches the DEBUG sink. Applied BEFORE
+#: scrubbing: `scrub_credentials` runs ~25 regex passes, each building a fresh
+#: string, and it was written for error text — whereas a hook payload can carry
+#: a whole retrieved document. Capping first bounds both scrub cost and log
+#: volume.
+MAX_PAYLOAD_CHARS = 2000
+
+
+def summarize_payload(payload: Any) -> dict[str, Any]:
+    """Describe a payload's SHAPE without reproducing any of its values."""
+    if isinstance(payload, dict):
+        return {"count": len(payload), "keys": sorted(str(k) for k in payload)}
+    return {"count": 0 if payload is None else 1, "keys": []}
+
+
+def render_scrubbed_payload(payload: Any) -> str:
+    """Render a payload for a DEBUG sink: truncated first, then scrubbed.
+
+    Callers are responsible for the two gates (opt-in + DEBUG); this function
+    only formats. Imported lazily so that merely importing a hook module does
+    not pull in the credential-scrub regex table.
+    """
+    from kaizen.utils.credential_scrub import scrub_credentials
+
+    rendered = repr(payload)
+    if len(rendered) > MAX_PAYLOAD_CHARS:
+        rendered = f"{rendered[:MAX_PAYLOAD_CHARS]}...[truncated {len(rendered)} chars]"
+    return scrub_credentials(rendered)
+
+
+def should_log_full_payload(enabled: bool, logger: logging.Logger) -> bool:
+    """Both gates, in one place, so neither hook can implement only one."""
+    return bool(enabled) and logger.isEnabledFor(logging.DEBUG)
--- a/packages/kailash-kaizen/src/kaizen/core/autonomy/hooks/builtin/logging_hook.py
+++ b/packages/kailash-kaizen/src/kaizen/core/autonomy/hooks/builtin/logging_hook.py
@@ -8,12 +8,31 @@ SECURITY: Supports sensitive data redaction (Finding #4 fix).
 """
 
 import logging
-from typing import ClassVar, Optional
+from typing import Any, ClassVar, Optional
 
 from ..protocol import BaseHook
 from ..types import HookContext, HookEvent, HookResult
 
 logger = logging.getLogger(__name__)
+
+#: Cap on a rendered payload before it reaches the DEBUG sink. Applied BEFORE
+#: scrubbing: ``scrub_credentials`` runs ~25 regex passes, each building a fresh
+#: string, and it was written for error text — whereas a hook payload can carry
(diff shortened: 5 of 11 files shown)


esperie:
## Commit 2 — `08ce48eb3`: sibling class + one shared implementation

Approved follow-up covering the `SecureLoggingHook` gap found while fixing #2070.

**The defect is narrower than its sibling, and the tests say so.** `SecureLoggingHook` (`hooks/security/redaction.py`, public API via `hooks.security`) was a near-verbatim copy of the pre-fix `LoggingHook` body. It *does* redact — `redact_sensitive` defaults True, redactor built unconditionally — so credentials and PII were already removed. What survived was everything the redactor does not claim.

RED evidence, with the credential already gone and the prose intact:

```
INFO redaction.py:273 [pre_agent_loop] Agent=agent-secure-2070 TraceID=trace-secure-2070
     Data={'inputs': {'prompt': 'draft the termination letter for the Henderson
     account', 'api_key': '[REDACTED]'}}
```

That is why `test_credential_absent_from_info` is a **negative control**, not the finding — it passes before *and* after. A suite that only checked credentials would have reported this class clean. The load-bearing assertion is the prose one.

**One implementation, because there were two.** Both hooks now share `hooks/_payload.py`. `should_log_full_payload` puts *both* gates (opt-in AND DEBUG) behind a single call, so neither caller can implement only one — the divergence that produced this bug is now structurally harder to repeat.

**Also fixed:** `self.redactor` was never assigned when `redact_sensitive=False`; any read outside the guard raised `AttributeError` rather than reading a falsy value. Now explicitly `None`.

### Verification

- Fail-first RED (2 failed, 2 passed — both passes are negative controls), then GREEN
- `tests/unit/core/autonomy/hooks/` — **130 passed** (126 + 4 new)
- Scrubber-parity regressions + `observability/` — **51 passed**
- `pre-commit run --files <4 changed>` — all pass incl. Tier 1 gate
- Full `tests/unit` sweep — **8778 passed**; the single failure was `test_kaizen_import_performance_target` (cold-import timing ceiling), which **passes in isolation at 2.78s** and was load-flake from a 39-minute concurrent run. My only module-level import change is `typing.Any`.

esperie:
## Commit 3 — `e00800156`: the four observability flags stop lying (#2084)

Approved as option (b′). Scope grew from "`enable_logging` is broken" to **all four flags**, and from "typo" to **stub**.

### Verdict: stub, not typo — evidence

The handler methods the call site registers — `start_trace`, `end_trace`, `record_start`, `record_end`, `log_start`, `log_end` — have **zero definitions anywhere in `src/`**, and `observability/tracing.py` does not exist. A corrected import path would have failed on the next line.

All four blocks, tested in the **real** `from X import Y` statement form:

```
enable_tracing   ImportError (CAUGHT by source's except) -> silently skipped
enable_metrics   ImportError (CAUGHT by source's except) -> silently skipped
enable_logging   ImportError (CAUGHT by source's except) -> silently skipped
enable_audit     ImportError (CAUGHT by source's except) -> silently skipped
```

**Probe correction worth recording:** my first probe used `__import__` + `getattr`, which raises `AttributeError` — but the source uses `from X import Y`, which raises `ImportError`. Different error type means a different catch path, so that probe was answering a different question than the one asked. Re-run in the real statement form above.

### Why tri-state and not "raise on True"

All four flags **defaulted `True`** (`agent_config.py:117,123,129,135`), and `agent.py:230` calls this on **every** agent construction. Runtime, on the untouched default path:

```
flags  = True True True True
is_observability_enabled() = True
create_observability() -> HookManager
hooks actually registered = 0
```

So a naive "raise when `True`" breaks every caller in existence — not the opt-in population. A dataclass cannot distinguish an explicit `True` from a defaulted one, so the flags become `Optional[bool]`: `None` warns, `True` raises, `False` is silent. Same mechanism as `LoggingHook.redact_sensitive` in commit 1, for the same underlying reason.

The raise is **not** wrapped in `try/except ImportError` — the defect being fixed *is* a swallowed `ImportError`.

### `is_observability_enabled()` — a correction, not a behaviour change

Previously returned `True` by default while zero hooks were registered, so a caller branching on it ran its observability-dependent path against an empty manager. Now tracks what is actually registered. Called out on its own line in the CHANGELOG with before/after.

### Second lie, fixed as a consequence

`rich_output.py:155-188` printed the same claim to the user's terminal. Before, a default agent announced:

```
✅ Observability:
   • Jaeger tracing (http://localhost:16686)
   • Prometheus metrics (localhost:9090)
   • Structured logging (INFO level)
   • Audit trail (.kaizen/audit.jsonl)
```

— four features, with endpoints and file paths, none of which existed. Now: `⚪ Observability: Disabled`.

### The invariant whose absence let this ship

```python
assert _registered_hook_count(manager) > 0, (
    "is_observability_enabled() reported True, create_observability() "
    "returned without raising, and ZERO hooks were registered."
)
```

`HookManager` exposes **no public accessor** for registered hooks — `get_stats()` reports execution counts, not registrations — so the test reaches `_hooks` and documents why. That missing public surface is part of why four dead subsystems were invisible.

### Dead code deleted, not retained

I first moved the four unreachable blocks into a `_unreachable_legacy_observability` method "for reference" and then removed it: keeping unreachable code that imports nonexistent modules is the same stub pattern this commit removes.

### Verification

- Fail-first RED **7 failed / 9 passed** (9 = negative controls), then GREEN **16 passed**
- Every test touching the changed surface — **84 passed, 15 skipped**
- Hooks suites — **130 passed**; scrubber/observability — **51 passed**
- `pre-commit run --files <5 changed>` — all pass incl. Tier 1 gate and doc-style
- No new `None` hazard: `create_observability` now returns `None` on the default path instead of an empty `HookManager`; `BaseAgent` already declares `hook_manager: Optional[Any] = None` and handles it (`base_agent.py:184-192`), building its own manager when `hooks_enabled`
- Swept every reader of the four flags across all packages: the only `AgentConfig` consumer is `rich_output.py` (above); the rest are unrelated classes with their own same-named attributes

**BREAKING** entry under `[Unreleased]` with the migration line. Whether the four subsystems get implemented or removed is **#2084** — this makes the flags honest, it does not make the features exist.

esperie:
## Commit 4 — `1fc072fcf`: a real regression the full sweep caught, plus the one-time warn

### The regression — every `Agent()` construction was raising

The full `tests/unit` sweep (8613 passed / 10 failed, 55 min) surfaced something **every targeted run had missed**:

```
kaizen.errors.ObservabilityNotImplemented: enable_tracing=True was requested...
```

`Agent.__init__` carries its **own** copies of the four flags, all defaulting `True` (`agent.py:87-90`), forwarded verbatim to `AgentConfig` at `:200-203`. So the tri-state read **every** `Agent()` construction as an explicit request and raised — reintroducing, one layer up, exactly the blast radius the tri-state exists to prevent.

7 tests failed this way across `test_governance_required_gate.py` and `test_llm_provider_bug_reproduction.py`. All 7 pass now: **107 passed**.

**Why targeted verification could not have caught it.** I swept *consumers* of the flags (readers of `config.enable_*`) and *callers* of `create_observability`. A forwarded default in a wrapper's signature is neither — no grep I ran could surface it. Only running the tests did. This is the second time in this PR that the wrapper layer, not the layer under change, carried the defect.

### The one-time warning

`create_observability` runs on every agent construction, so commit 3's consolidated warning still fired per call — a process building 100 agents got 100 copies. Measured: **3 constructions → 3 warnings** before, **5 → 1** after.

Not cosmetic. The entire remedy for the default path *is* a warning, and one repeated on a hot path is one operators filter — the loud-warning remedy quietly becomes no remedy, the same degrade-to-useless shape as the defects this branch fixes.

`lru_cache` rather than an instance flag, because `SmartDefaultsManager` is constructed per agent — an instance flag would warn once per instance and leave the spam intact. The autouse `cache_clear()` fixture is load-bearing, not hygiene: with a process-wide warning, `test_explicit_false_is_silent` would otherwise pass merely because an earlier test consumed the one warning, and an absence assertion that cannot fail is not evidence.

### The other 3 sweep failures — not mine, established by reachability

| test | can it reach the change? |
| --- | --- |
| `test_enterprise_memory_system.py` (×2) | **No** — zero references to `Agent` / `AgentConfig` / `enable_*` / `smart_defaults` |
| `test_simple_qa_async.py::test_async_execution_overhead` | **No** — `SimpleQAAgent(BaseAgent)` subclasses `kaizen.core.base_agent.BaseAgent` directly; never touches `Agent`, `AgentConfig`, or `SmartDefaultsManager` |

All three are absolute wall-clock thresholds (`< 5.0s`, `10ms` p99) measured on a machine running **9 concurrent pytest processes** from sibling lanes. Attribution is by *unreachability of the changed code*, which is stronger than "it was failing before" — and the pre-change sweep on this branch showed the same class of failure (`test_kaizen_import_performance_target`, which passes in isolation at 2.78s).

### Verification

- 7 regressed tests → **107 passed**
- New suite **17 passed** (16 + the one-time-warn test)
- `pre-commit run --files <changed>` — all pass incl. Tier 1 gate
- Worktree integrity: a baseline probe left the tree detached at a timeout; restored to `fix/2070-logging-hook-redaction-defeat` with `git stash pop`, verified clean, and the pre-existing sibling stashes left untouched

Also filed on **#2084**: a working `ObservabilityManager` / `TracingManager` **does** exist under the same package — `smart_defaults` was importing a *different*, nonexistent set. That opens a third option (wire to what exists) which I flagged there as a lead to verify, not a conclusion.

esperie:
## D1 RESOLVED — this PR is superseded and will be closed once the replacement lands

**Verdict: wiring IS feasible, so the tri-state raise is no longer the right fix.**

All four observability hook classes already exist — in `packages/kailash-kaizen/src/kaizen/core/autonomy/hooks/builtin/` (`tracing_hook.py:50`, `logging_hook.py:24`, `metrics_hook.py:29`, `audit_hook.py:16`) — **not** in the `observability/` path `smart_defaults` imported (`grep -rn '^class .*Hook'` there returns 0). `HookManager.register_hook` (`hooks/manager.py:155`) is the registration contract, and its own docstring example is the exact wiring; `BaseAgent.enable_observability()` (`core/base_agent.py:915-935`) already does it for tracing.

So #2084 is a **wrong-module-path + wrong-registration-idiom** defect, not four missing subsystems. The six phantom handler methods confirm it (scoped to real source trees, excluding `build/lib` shadows): `start_trace 0, end_trace 0, record_start 2, record_end 0, log_start 0, log_end 0` — and both `record_start` hits are unrelated (`src/kailash/infrastructure/execution_store.py:119,306`).

Baseline re-derived in-process with a resolved-`__file__` receipt: `flags = True True True True`, `is_observability_enabled() = True`, **hooks registered = 0**.

**The flags become honest by becoming TRUE, not by raising** — which is non-breaking AND delivers the advertised feature. That makes this PR's `Optional[bool]` tri-state and `errors.py::ObservabilityNotImplemented` unnecessary.

### Why the whole PR closes, not just half

An initial read suggested splitting this PR to preserve its #2070 logging work. **That half already landed** — #2070 is CLOSED/COMPLETED by **#2094** (merged 2026-08-12T02:07:50Z); #2094 *was* that split, performed last session.

Verified by content rather than issue state — all three files are **byte-identical** between `origin/main` and this branch's tip (`git show <ref>:<path> | shasum`): `hooks/_payload.py` (62 lines), `builtin/logging_hook.py` (279), `security/redaction.py` (370).

⚠️ **Note for anyone re-reading the diff:** `git diff --stat origin/main...<this branch>` still shows `logging_hook.py +117`, `_payload.py +62`, `redaction.py +60`. That is a **merge-base artifact** — the three-dot form diffs from the merge base and this branch was never rebased past #2094. It is not unlanded content, and it is also why this PR reads CONFLICTING. Confirmed from the other direction too: `ObservabilityNotImplemented` and the `agent_config.py` `Optional[bool]` flags are **absent** from main, so the observability half is genuinely this branch's only unique content.

### Disposition

**Held OPEN deliberately** until the replacement PR merges — closing now would retire the raise while nothing yet replaces it. On merge this closes as superseded, citing #2094 (logging half) and the replacement PR (observability half).

The branch and its worktree are untouched and will stay untouched.

esperie:
## The D1 verdict this PR was held on: **wiring is feasible** — see #2101

The lead you flagged on #2084 checks out, and further than stated: the four hook classes exist, in `hooks/builtin/`, and `HookManager.register_hook()` is the contract they were written for — its own docstring example is the exact wiring, and `BaseAgent.enable_observability()` already does it for tracing.

```
hooks/builtin/tracing_hook.py:50:class TracingHook(BaseHook)
hooks/builtin/metrics_hook.py:29:class MetricsHook(BaseHook)
hooks/builtin/logging_hook.py:24:class LoggingHook(BaseHook)
hooks/builtin/audit_hook.py:16:class AuditHook(BaseHook)
```

So the flags can be made honest by becoming **true** rather than by raising. #2101 wires them: default path now registers 4 hooks across 18 events and writes the audit trail, and `is_observability_enabled()` no longer over-reports.

**What that means for this PR.** Your `#2070` half already landed on main as #2094 — `_payload.py`, the structure-only `handle()`, the tri-state `redact_sensitive` raise, and both logging test files are all there. Everything unique still in this branch is the observability half: the `Optional[bool]` tri-state, `errors.py::ObservabilityNotImplemented`, and `test_observability_flags_fail_loudly.py`. That is exactly what #2101 supersedes, so **this PR looks closeable in full** once #2101 lands. Recording the assessment rather than acting on it — the disposition is the orchestrator's.

Two things from this branch that #2101 kept, because they were right:

- **The tri-state reasoning was sound for the fallback it was designed for.** All four flags do default `True` (`agent_config.py:117,123,129,135`) and `agent.py:230` runs on every construction, so a naive raise-on-`True` would have broken every caller — your correction of that premise held up when I re-derived it.
- **The one-time warning.** `create_observability` runs per agent construction, so #2101's missing-dependency warning uses the same `lru_cache` shape, with the autouse `cache_clear()` fixture, for the same reason: a warning repeated on a hot path is one operators filter, and the loud-warning remedy quietly becomes no remedy.

#2101 also found a second layer neither issue had recorded: `BaseAgent.__init__` accepts a `hook_manager`, stores it, and then gates `trigger_hook` on `config.hooks_enabled` — a separate opt-in defaulting `False` that `Agent._convert_to_base_agent_config` never sets. Registering the four hooks without that fix would have left them attached to a manager that was never triggered, so the audit trail would still have recorded nothing on exactly the path users take.

esperie:
**Closing as SUPERSEDED — both halves of this PR are now delivered on main, by other paths.**

**Logging half (#2070)** → landed as **PR #2094** (merged 2026-08-12T02:07:50Z), which closed #2070 COMPLETED. Verified by content, not by issue state: `hooks/_payload.py` (62 lines), `builtin/logging_hook.py` (279), `security/redaction.py` (370) are **byte-identical** between this branch's tip and `origin/main` (sha256 compared via `git show <ref>:<path> | shasum`).

⚠️ **Note for anyone re-reading this PR's diff:** `git diff --stat origin/main...` still shows `logging_hook.py +117`, `_payload.py +62`, `redaction.py +60`. That is a **merge-base artifact** — the three-dot form diffs from the merge base and this branch was never rebased past #2094. It is not unlanded content, and it is also why this PR read CONFLICTING.

**Observability half (#2084)** → landed as **PR #2101** (merged as `5c3d98558`), which closed #2084 COMPLETED.

This PR's approach to #2084 was a tri-state `Optional[bool]` that would **raise** on explicit opt-in, chosen when wiring was believed infeasible. An investigation this session established that wiring **is** feasible: all four hook classes already exist in `core/autonomy/hooks/builtin/` (`tracing_hook.py:50`, `logging_hook.py:24`, `metrics_hook.py:29`, `audit_hook.py:16`) — not in the `observability/` path `smart_defaults` was importing, where `grep -rn '^class .*Hook'` returns 0. `HookManager.register_hook` is the registration contract and its own docstring example is the exact wiring.

So the flags became honest **by becoming true** rather than by raising — non-breaking, and it delivers the advertised feature. That makes this PR's `errors.py::ObservabilityNotImplemented` and `Optional[bool]` tri-state unnecessary.

#2101 also fixed three defects that only surfaced once the flags actually installed hooks: `Agent()` raising `OSError` on a read-only filesystem (the `mkdir` was dead code pre-fix, shielded by the very ImportError swallow that WAS the #2084 defect), the audit file being created world-readable 0644, and `metadata` values being written verbatim to the audit trail while the docstring claimed structure-only.

**Nothing here is lost or rejected** — the diagnosis in this PR was correct and led directly to the fix that landed. The branch and its worktree are retained for now.

Follow-ups filed from #2101: **#2109** (audit durability), **#2110** (unbounded/unrotated audit file, CWD-relative default), **#2111** (`enable_checkpointing` is a fifth advertised-but-unimplemented subsystem, same class as #2084).

PR #2100, opened 2026-08-12
fix(servers,api,gateway,channels): fail closed on unauthenticated workflow execution (#2072)

### LINKED ISSUE #2072: CRITICAL: default WorkflowServer/create_gateway serves anonymous arbitrary workflow execution — POST /workflows/{name}/execute is unauthenticated
## Summary

On a **default** `WorkflowServer` / `create_gateway()` with a single `register_workflow(...)` — the primary documented use — `POST /workflows/{name}/execute` executes the workflow for an **anonymous caller**. No credentials of any kind are required.

This is split out of #2025. That issue named the opt-in `proxy_workflow` catch-all; PR #2064 closes that route. **This issue is the default-reachable half, which #2064 does NOT close** — which is why #2064 is `Refs #2025`, not `Fixes`.

## Evidence — measured, not inferred

A real `WorkflowServer` booted under uvicorn on a real port, hit with a real HTTP client. No test-client emulation, no synthetic app, no mocks.

```
Server: http://127.0.0.1:65035 (default WorkflowServer, one registered workflow)
NO credentials are sent on any request below.

POST /workflows/probe/execute   -> 200
  body: b'{"outputs":{"n":{"result":{"ran":true}}},"execution_time":4.677430868148804,
           "workflow_id":"8e987b59-26fe-4a67-99f7-e7b502701967","version":"1.0.0"}'

VERDICT: ANONYMOUS WORKFLOW EXECUTION - no credentials were sent
```

The workflow did not merely route — **it ran**. `{"ran": true}` is the node's output and `execution_time` is real wall-clock. An unauthenticated caller caused server-side code execution.

Reproducer: see the appendix at the bottom of this issue.

## Why a route dependency cannot fix this

`register_workflow` mounts a sub-application:

```python
prefix = f"/workflows/{name}"
self.app.mount(prefix, workflow_api.app)     # src/kailash/servers/workflow_server.py
```

`register_mcp_server` does the same for `/mcp/{name}`.

**Route-level and app-level `Depends` do not reach a mounted sub-application; ASGI middleware does.** Measured:

```
A app-level Depends  -> /direct        : 401
A app-level Depends  -> mounted execute: 200      <-- OPEN
B middleware         -> /direct        : 401
B middleware         -> mounted execute: 401      <-- closed
B middleware + creds -> mounted execute: 200      <-- discrimination control
```

The third line is the control: without it, the 401 on line 2 of B could equally be a broken route. So any fix here MUST be middleware; a `Depends` would look correct, pass a naive test against a parent route, and leave `/execute` wide open.

## Full surface still unauthenticated on a default server

`grep -nE 'Depends\(|HTTPBearer|APIKeyHeader|verify_token|authenticate' src/kailash/servers/` returns zero matches across all five server modules; the only middleware in that directory is CORS, and it is conditional.

| Route | Consequence |
| --- | --- |
| `POST /workflows/{name}/execute` | **anonymous arbitrary workflow execution** (proven above) |
| `POST /workflows/{id}/signals/{signal}` | mutates running-workflow state; rate-limited, not authenticated |
| `GET /workflows/{id}/queries/{query}` | reads running-workflow state |
| `GET /workflows` | enumerates every registered workflow |
| `/mcp/{name}/…` | the entire MCP surface |
| `GET /durability/requests` | dumps `path`, `method`, **`client_ip`** per in-flight request |
| `WS /ws` | open socket |
| `GET /metrics`, `GET /dashboard` | metrics and dashboard |

## Recommended fix shape

An ASGI middleware on `WorkflowServer.app`, delegating to core's existing
`kailash.trust.auth.jwt.JWTValidator` / `JWTConfig` — already framework-agnostic, already
supports `exempt_paths` and enforces a 32-byte minimum secret. Nexus's `JWTMiddleware` is the
Starlette shape to model, but **core cannot import it** (Nexus depends on `kailash`, not the
reverse), so it is the template, not the dependency.

Controls MUST be **named parameters** (`require_auth`, `auth_config`), never `**kwargs`. This
codebase has already shipped that exact defect twice: `enable_auth=True` was advertised in
`servers/__init__.py`'s docstring while landing in `**kwargs` and being discarded (corrected in
PR #2064), and `Nexus(enable_auth=True)` set three booleans and installed nothing (#2013).

CORS must stay **outermost** — Starlette's `add_middleware` prepends, so auth added after CORS ends
up inside it and preflight `OPTIONS` gets 401'd. PR #2054 hit exactly this.

## The open product decision — for the repo owner

**Fail-closed default, or opt-in with a loud WARN?**

`security.md` § Secure-Default and the precedent already set in this codebase both point to
fail-closed (`require_auth=True`, raising at construction naming the env var and the opt-out) —
the same call was made for `APIGateway(enable_auth=True)` in #636 and `Nexus(enable_auth=True)` in
#2054.

**Honest con, stated plainly: this is a hard breaking change on a released public API.** Every
existing `create_gateway()` deployment stops booting until a secret is set or `require_auth=False`
is passed explicitly.

**Argument for accepting that:** every deployment this breaks was already serving anonymous
workflow execution. The raise does not create an outage-shaped problem out of a working system; it
surfaces a pre-existing exposure at deploy time instead of leaving it silently in production. The
alternative is non-breaking but leaves every existing **and every new** deployment unauthenticated
by default, which does not close this for anyone who does not read logs.

## Acceptance criteria

- [ ] `POST /workflows/{name}/execute` with no credentials returns 401 on a default server. **This is the load-bearing test** — if it returns 200, the mechanism is a `Depends` and the fix shape is wrong.
- [ ] Same for `/mcp/{name}/…`, `/durability/requests`, the signal/query endpoints, and `/workflows`.
- [ ] A discrimination control: the same route with valid credentials returns 200, so the 401 is authentication and not a broken route.
- [ ] CORS preflight `OPTIONS` does NOT 401 (middleware ordering).
- [ ] Clean-env construction test pinning whichever default the owner chooses.
- [ ] Controls are named parameters; a swallowed-kwarg regression test.

## Appendix — reproducer

```python
import json, socket, threading, time, urllib.error, urllib.request
from kailash.servers import WorkflowServer
from kailash.workflow.builder import WorkflowBuilder

with socket.socket() as s:
    s.bind(("127.0.0.1", 0)); port = s.getsockname()[1]

wb = WorkflowBuilder()
wb.add_node("PythonCodeNode", "n", {"code": "result = {'ran': True}"})

server = WorkflowServer(title="probe")
server.register_workflow("probe", wb.build())        # -> self.app.mount(...)
threading.Thread(target=server.run,
                 kwargs={"host": "127.0.0.1", "port": port}, daemon=True).start()

for _ in range(100):
    try:
        with socket.create_connection(("127.0.0.1", port), timeout=0.2): break
    except OSError: time.sleep(0.1)

req = urllib.request.Request(f"http://127.0.0.1:{port}/workflows/probe/execute",
                             data=json.dumps({"inputs": {}}).encode(),
                             headers={"Content-Type": "application/json"}, method="POST")
try:
    with urllib.request.urlopen(req, timeout=10) as r:
        print(r.status, r.read()[:200])              # -> 200, workflow output
except urllib.error.HTTPError as e:
    print(e.code, e.read()[:200])
```

## Related

- #2025 — the opt-in proxy catch-all. Same defect class, less reachable. PR #2064 closes that half.
- #2013 / PR #2054 — `Nexus(enable_auth=True)` installed nothing; the fail-closed precedent.
- #636 — `APIGateway(enable_auth=True)`; the earlier fail-closed precedent.
- #2065 — CodeQL `py/partial-ssrf` deferral on PR #2064.

esperie:
## Summary

A default `WorkflowServer` / `create_gateway()` with a single `register_workflow(...)` — the primary documented use — **executed the workflow for a caller presenting no credentials of any kind**. This makes server-wide authentication fail closed, and closes the same defect on **six** independent surfaces — plus the WebSocket scope, which no HTTP middleware could see.

**This is a hard breaking change and is meant to be.** Every deployment it stops was already serving anonymous workflow execution. The raise does not turn a working system into an outage; it surfaces a pre-existing exposure at deploy time instead of leaving it silently in production. Migration is in the CHANGELOG under `[Unreleased]`.

## Fail-first evidence — both polarities, on a real socket

The issue's verbatim reproducer, run against `origin/main` source:

```
UNDER TEST: /tmp/w1a_prefix/src/kailash/servers/__init__.py
POST /workflows/probe/execute HTTP/1.1" 200 OK
STATUS: 200 BODY: b'{"outputs":{"n":{"result":{"ran":true}}},"execution_time":6.928347110748291,...}'
VERDICT: ANONYMOUS WORKFLOW EXECUTION - no credentials were sent
```

The workflow did not merely route — it **ran**. Same script, this branch, no credentials configured:

```
CONSTRUCTION REFUSED: ServerAuthNotConfiguredError
```

Same script, after following migration step 1 (`KAILASH_JWT_SECRET` set):

```
POST /workflows/probe/execute HTTP/1.1" 401 Unauthorized
STATUS: 401 BODY: b'{"detail":"Not authenticated","error":"missing_token"}'
VERDICT: anonymous caller REJECTED
```

## Why middleware, not a route dependency

`register_workflow` and `register_mcp_server` both `app.mount(...)`. A `Depends` on the parent does **not** run for a request routed into a mounted sub-app. Measured, with a credentialed third row as the discrimination control — without it the 401 on line 4 could equally be a broken route:

```
A app-level Depends  -> /direct        : 401
A app-level Depends  -> mounted execute: 200      <-- OPEN
B middleware         -> /direct        : 401
B middleware         -> mounted execute: 401      <-- closed
B middleware + creds -> mounted execute: 200      <-- control
```

## Six surfaces, one shared implementation

Per `security.md` § Enforcement-Surface Parity, all resolve through the new `kailash.utils.server_auth` rather than a copy each. Four were found by sweeping *after* the primary fix looked complete — the sixth only after an adversarial security review of the finished work:

1. **`WorkflowServer`** + `Durable`/`Enterprise` subclasses, and `create_gateway()`.
2. **`WorkflowAPIGateway`** — independent surface, mounted at `/{name}`.
3. **`WorkflowAPI`** — registers `POST /execute` and ships its own `uvicorn.run()`, so `WorkflowAPI(wf).run()` was the same anonymous execution **reachable without `WorkflowServer` at all**.
4. **`create_gateway_app()`** — installed the gate on an app nobody serves. `EnhancedDurableAPIGateway` authenticates its *own* FastAPI instance, but the function returns a **different** app carrying the router; its `Depends(get_gateway)` is an instance injector, not authentication. Measured before the fix, **with a secret configured**:

   ```
   middleware on RETURNED app: []
   middleware on GATEWAY  app: ['BaseHTTPMiddleware', 'JWTAuthMiddleware']
   GET /api/v1/workflows (NO creds) -> 200
   ```

5. **`APIChannel`** — see below.
6. **`kailash.middleware.create_gateway`** — a **second callable exported under the same name** as the fixed `kailash.servers.gateway.create_gateway` (`kailash/middleware/__init__.py:197` re-exports `communication.api_gateway.create_gateway`). So `from kailash import create_gateway` gave the fixed one and `from kailash.middleware import create_gateway` gave an ungated one — and that module's own docstring at `:100` demonstrates the ungated one, so a user following the module's example got anonymous workflow execution.

   `enable_auth=True` built a `JWTAuthManager` that **no route consulted**. Measured on a real socket with `KAILASH_API_GATEWAY_SECRET` set, every request uncredentialed:

   ```
   enable_auth: True   auth_manager: JWTAuthManager
   GET    /                          -> 200
   GET    /api/workflows             -> 200
   POST   /api/executions?session_id -> 500   (handler RAN)
   DELETE /api/executions/e          -> 500   (handler RAN)
   GET    /api/stats                 -> 200
   GET    /openapi.json              -> 200
   ```

   Not one 401. The 500s are the application failing on a missing session **after** routing, which is what proves the request reached the handler — an anonymous caller holding any live `session_id` executed workflows. After, with both controls:

   ```
   UNCREDENTIALED  /, /api/workflows, /api/executions, /api/stats, /openapi.json  -> 401
   UNCREDENTIALED  /health                                                        -> 200
   CREDENTIALED    (token minted by the gateway's OWN auth_manager)               -> 200
   WRONG-KEY       (token signed with a different secret)                         -> 401
   ```

   The verifier is derived from the gateway's own token **issuer** (key, algorithm, issuer, audience), so tokens this gateway mints are the tokens it accepts and an existing `KAILASH_API_GATEWAY_SECRET` deployment keeps working. `enable_auth` (token issuance) and `require_auth` (request gate) stay separate knobs, because this gateway does both. `require_auth` is **tri-state** (`Optional[bool]`) for the same reason `ChannelConfig.enable_auth` is in this PR — a plain `bool` cannot tell "the operator never said" from "the operator said no". Unstated inherits fail-closed, but `enable_auth=False` is honoured as the explicit opt-out it was: before #2072 it was the *only* auth control this class had, so a caller who set it said what they wanted in the words available to them (#636's contract, which a first pass here broke). An explicit `require_auth` is the newer, more specific statement and wins in both directions. An `auth_manager` carrying no derivable key does **not** satisfy the gate — it cannot verify anything, and accepting it would install a middleware with nothing to verify against; that raises, with a WARN naming the manager.

## WebSocket routes bypassed the gate entirely

`JWTAuthMiddleware` extends Starlette's `BaseHTTPMiddleware`, whose `__call__` begins:

```python
if scope["type"] != "http":
    await self.app(scope, receive, send)
    return
```

`dispatch()` therefore **never ran for a websocket scope**, so every `@app.websocket(...)` route was served with no credential on a `require_auth=True` server — `workflow_server.py:581` and `enterprise_workflow_server.py:661` both register one. The middleware's own docstring said "before they reach any route" and `install_server_auth_middleware`'s inherited that; both were false and are corrected, not merely re-worded.

Measured on a real socket against `WorkflowServer(require_auth=True)`:

```
                                BEFORE (HTTP layer only)   AFTER (both layers)
UNCREDENTIALED                  HANDSHAKE OK, echoed       REFUSED HTTP 403
CREDENTIALED (control)          HANDSHAKE OK, echoed       HANDSHAKE OK, echoed
WRONG-KEY (negative control)    HANDSHAKE OK, echoed       REFUSED HTTP 403
```

`JWTWebSocketAuthMiddleware` is a **pure ASGI callable**, not a `BaseHTTPMiddleware` subclass — inheriting from that base is precisely what creates the blind spot, and a type-invariant test pins that. It rejects at the handshake rather than accepting then closing, so an unauthenticated caller never reaches the route's `accept()` or its `while True`. The core handler is an echo, so immediate data exposure is low; what it closes is unauthenticated per-connection resource consumption, a docstring inviting subclasses to make it stateful, and `register_mcp_server` mounting third-party MCP apps told not to install their own gate.

**Also fixes a broken route found by that work.** `EnterpriseWorkflowServer`'s `/ws` declared `async def websocket_endpoint(websocket)` — unannotated — and FastAPI resolves a websocket endpoint's parameters like a route's, so `websocket` became a required **query parameter** and every handshake failed validation and closed, regardless of credentials. Measured with the gate **off**, so it is not an auth interaction:

```
WorkflowServer           -> OK echo: 'Echo: hi'
EnterpriseWorkflowServer -> FAILED WebSocketDisconnect
```

This surfaced because the websocket gate's *credentialed control row* failed on enterprise while passing on the parent. That is the discrimination control earning its place: without it, "uncredentialed handshake refused" would have read green against a route that refused everyone, and the gate would have been pinned by a test unable to tell a gate from a broken endpoint.

## Two defects on the `auth_config=` branch

Both made `auth_config=` the **weakest** way to ask for authentication.

- **It reinstated the broad exempt default.** `resolve_server_auth` returned the caller's `JWTConfig` untouched, so it carried `JWTConfig`'s own default `exempt_paths` — `/health, /metrics, /docs, /openapi.json, /redoc, /auth/login, /auth/refresh, /auth/sso/*`. The module comment at `server_auth.py:93-99` says that list is *"deliberately NOT reused here"* because *"the OpenAPI documents hand an anonymous caller the full route map."* On this branch it **was** reused, so `create_gateway(auth_config=JWTConfig(secret=...))` — the natural spelling — restored anonymous `GET /openapi.json` route enumeration. The only covering test passed an explicit `exempt_paths=["/health"]`, so the default was never exercised.
- **`extra_exempt_paths` was ignored.** Threaded from all six surfaces, read only on the environment branch. `create_gateway(auth_config=..., auth_exempt_paths=["/probe"])` accepted a documented security kwarg with **zero effect** — `zero-tolerance.md` Rule 3c verbatim.

Both now route through one helper that returns a **copy** (never mutating a config the caller may reuse). "Untouched" is decided by comparing against the dataclass field's own `default_factory` output, so it cannot drift from `JWTConfig`. Stated residual: a caller who *explicitly* passes a list byte-equal to that default is narrowed as though they passed nothing — the fail-closed direction of an ambiguity a plain dataclass cannot resolve.

## Log sanitizer: lone surrogates destroyed the record

`safe_log_text` swept `Cc`/`Cf`/`Zl`/`Zp` but not `Cs`. A lone surrogate cannot forge a line — it makes the record **un-encodable**, so the handler raises `UnicodeEncodeError` inside `emit`, `logging` routes it to `handleError`, and the record is dropped. On these call sites the dropped record is the one documenting **authentication being skipped**, so a caller could delete the evidence of its own exposure. Measured through a real UTF-8 `StreamHandler`:

```
before:  safe_log_text("a\ud800b") -> 'a\ud800b'   emitted bytes: b''
after:   safe_log_text("a\ud800b") -> 'a?b'        emitted bytes: b'record a?b\n'
```

Suppressing the evidence and forging it are the same attack on the operator reading the log, so `Cs` belongs in the same frozenset.

Mounted sub-apps install nothing: the parent's middleware already accepted or rejected the request, so a second layer would demand a second credential for one call. Verified the sub-app resolves to no config while the parent still gates the mounted route 401/200.

## Two security defects found while verifying the in-flight work

Both are distinct defects, not cleanup:

- **Nexus declared authentication it never installs.** `external_auth_reason` was asserted unconditionally at both `create_gateway()` call sites, but Nexus installs its middleware only when `_enable_auth` is set — **False by default in development**. So a default `Nexus()` told the new fail-closed gate "an external middleware authenticates every request" while nothing did, **and suppressed the opt-out WARN naming the exposure**. A false assurance defeats the gate more thoroughly than an absent one: this is the #2013 shape (a control reporting success while installing nothing) re-entering through the declaration instead of a `hasattr` guard. Both sites now resolve from the same flag through one helper.
- **`APIChannel` still defaulted to an open server.** `ChannelConfig.enable_auth` was a plain `bool` defaulting to `False`, which cannot distinguish "the operator never said" from "the operator said no" — so every default `APIChannel` kept serving the exact route closed everywhere else. The field is now tri-state: `None` inherits fail-closed, `False` stays an honoured explicit opt-out. **The shared default was deliberately not flipped to `True`**: the CLI and MCP channels enforce nothing, so that would make them *report* a control they do not perform — the same false-assurance class. They report `bool(...)`.

## Other guarantees

- Controls are **named parameters**, never `**kwargs` — a typo is a `TypeError`, not an open server. `enable_auth=True` has shipped as a swallowed kwarg twice (#2025, #2013).
- `auth_manager` deliberately does **not** satisfy `require_auth` — it supplies a `Depends`, which per the measurement above never reaches the mounted route.
- **Only health probes are exempt.** `JWTConfig`'s own default list exempts `/docs` and `/metrics` and is deliberately not reused — `/` and `/workflows` enumerate every workflow, and OpenAPI hands an anonymous caller the route map.
- CORS stays outermost; a regression test asserts preflight succeeds **and** that the real cross-origin POST is still 401'd, so the ordering fix cannot smuggle a request through.
- 401 bodies carry a stable error code and never echo the token, exception text, or config.
- Fixes a `DeprecationWarning` (`@app.on_event`) surfaced by the new tests; cleanup now runs in a `finally`, so a failed boot no longer leaks the resource registry.

## CodeQL — corrected

An earlier comment on this PR claimed **"all 4 alerts fixed, none deferred."** That was **wrong**, and a full correction with the measurements is [posted as a comment](https://github.com/terrene-foundation/kailash-py/pull/2100#issuecomment-5265104333). In short: two `py/log-injection` mediums were **live** at `server_auth.py:345` and `:359` on `ca8f1200b` (verified to be this PR's merge commit by its parents), while the CodeQL **check** read `conclusion=success` because mediums do not meet its failure threshold — its own `output.title` said *"2 new alerts including 2 medium severity security vulnerabilities."*

**A green CodeQL check does not mean zero new alerts.** `conclusion` returns `success` for both zero alerts and two mediums, so it could not have discriminated.

Root cause: the sanitizer was correct at runtime the whole time — its barrier was in a call SHAPE the `py/log-injection` model does not recognize. Three forms were tried against the real scanner. All three are runtime-identical (by the time pass 2 runs, the category sweep has already replaced every ASCII line terminator, so it is a no-op in every form); only the shape differs, and only the shape is what a taint analyzer reads:

| barrier form | alerts |
|---|---|
| compiled `Pattern.sub` method | 2 live |
| `re.sub` module function | 2 live |
| **`str.replace` on literals** | **0** |

**Resolved — zero alerts.** Measured with two independent instruments on head `1e6f1d637`:

```
$ gh api '.../code-scanning/alerts?state=open&ref=refs/pull/2100/merge'
(empty)

$ gh api '.../check-runs/94085541845'
conclusion=success
title=No new alerts in code changed by this pull request
```

The `title` field is the discriminating one — on the previous head it read *"2 new alerts including 2 medium severity security vulnerabilities"* while `conclusion` read `success` either way. It can report both outcomes, and it now reports the other one.

`safe_log_text` still wraps every tainted component at both sinks: the alerts cleared because the sanitizer became **visible to the query**, not because a field was removed or an alert suppressed. `test_line_terminator_barrier_is_a_literal_str_replace` pins the shape so a future "optimization" back to a compiled pattern is a loud failure rather than a silently re-opened alert.

No `zero-tolerance.md` Rule 1b deferral is needed. Issue #2115 was opened for that path and is closed with the measurement.

## Known and deferred

- **A seventh un-gated HTTP server** — #2112. `visualization/api.py:170` builds a FastAPI, `:774` serves it, exposing `/api/v1/runs`, `/api/v1/runs/{id}/tasks`, a report download and two websockets, with no auth parameter anywhere. **No execute route**, so lower severity than the six above, but stated here rather than left inside an unqualified "six surfaces" claim.
- **#2113** — `mounted_subapp_auth_kwargs` renders "authentication explicitly disabled; it logged that exposure once" whenever the parent has no config, but that is also true when the parent used `external_auth_reason` (which logged `server_auth.external`, not the WARN the string asserts). Needs to be tri-state. Audit-record accuracy, not an open route.
- **#2114** — `secrets.compare_digest` raises `TypeError` on non-ASCII, and Starlette decodes headers latin-1, so `X-API-Key` with a byte ≥ 0x80 reaches it. Fails **closed** (401) but via `logger.exception` — one traceback per request, unauthenticated and attacker-drivable. Log-volume amplification, not a bypass.

## Testing

**99 passed** in `tests/regression/test_issue_2072_fail_closed_server_auth.py` (46 → 99). Fail-first verified: the new tests run **unchanged** against the pre-fix source tree and **31 of the 46 selected RED**; the 15 that pass there are the negative controls and structural pins that were already true, which is what they are for.

Run them with the packages on the path — `pytest.ini` sets `pythonpath = src` only, so the two nexus assertions otherwise resolve the *installed* nexus rather than `packages/kailash-nexus/src` and fail phantom:

```
pytest tests/regression/test_issue_2072_fail_closed_server_auth.py \
  -o "pythonpath=src $(ls -d packages/*/src)"
```

No `Mock` anywhere in the file — a `Mock` satisfies every `hasattr`, so a mock-driven guard test passes identically whether the guard is installed or not, which is how the inert control in #2013 shipped green. Every 401 assertion is paired with a credentialed 200 control.

Full local CI-parity: `pre-commit run` green across all 21 hooks including the Tier 1 unit gate.

**No new failures.** The broad set (`tests/regression tests/unit/servers tests/unit/channels`) run against a clean `origin/main` tree produces an **identical** failure profile — 52 failed / 37 errors, root causes 75 `ModuleNotFoundError`, 10 `ValueError`, 10 `RuntimeError`, 4 `AssertionError`, 1 `ImportError`. Those are environmental, not code: site-packages carries both `kailash-enterprise` 4.24.2 and `kailash` 2.45.6 (the package's own import warning flags them as conflicting), and `kailash.dataflow` / `kailash.kaizen` resolve only from that install, never from this repo's `src`.

## Stated plainly

- The broad-suite runs used a worktree with `_kailash.cpython-312-darwin.so` copied in from the main checkout (it is gitignored and cannot reach a commit). That worktree is therefore **not** a clean reproduction of CI, so the green above should not be read as proof that a clean checkout collects. The origin/main baseline was run the same way, so the *comparison* holds.
- CI on this PR is the first run against a real clean checkout.

## Related issues

Fixes #2072
Refs #2025, #2013, #636


esperie:
## CodeQL: all 4 alerts fixed, none deferred

The first run reported 4 new alerts (2 high). All were in code this PR adds, and all are now closed in `b82ae2b15`.

**Clear-text logging (2, high).** `JWTConfig` is built from the signing secret, so every attribute read off it is taint-carrying to a dataflow analyzer. `config.algorithm` is not the secret — but proving that to a scanner is not possible, and asserting it is not worth doing on an auth path. The install log now resolves the algorithm through a membership test against a fixed set, so it emits a string this module owns. The opt-out WARN's wiring text likewise became a plain literal instead of an f-string over `JWT_SECRET_ENV`: that constant holds the variable's *name*, never its value, but interpolating a `*_SECRET_*` symbol into a log line is indistinguishable from logging the credential, and a reviewer should not have to resolve that by hand.

**Log injection (2).** `external_auth_reason` and `server_label` reach log fields as caller text, so a CR/LF or U+2028 in either forges a second record — on the very lines documenting that authentication was *skipped* or *disabled*. Adds `secure_logging.safe_log_text`, which strips Unicode `Cc`/`Cf`/`Zl`/`Zp`, bounds length, and never raises. The existing `_safe_identifier` was the wrong tool: its identifier charset renders prose as `mounted?under?Enterprise...` and destroys the message. `U+2028`/`U+2029` are stripped alongside `\n` because many log viewers and JSON consumers break lines on them, so an ASCII-only strip leaves the injection open in a form that reads as whitespace in a diff. The construction-time raise is sanitized too — callers routinely catch and log it, so the exception path forges records the same way.

**Mutation-tested, not assumed.** My first injection test asserted on the server-title field and **passed with the sanitizer removed** — vacuous, because `server_label` is built with `{title!r}` and `repr` already escapes CR/LF upstream. I retargeted it at `external_auth_reason`, which passes through no `repr`: green unmutated, **red** with `safe_log_text` removed from that sink. Both polarities were recorded before the test was kept.

Suite now **54 passed** (from 46), pre-commit green across all 21 hooks.

Note on the earlier `build` failure: that job was **canceled**, not failed — `##[error]The operation was canceled.` 33s into the docs step, from an older run id superseded when the PR opened. Not a code failure.

esperie:
## Correction: "all 4 alerts fixed, none deferred" was wrong

The earlier comment titled **"CodeQL: all 4 alerts fixed, none deferred"** is inaccurate and I am correcting it rather than editing it, so the error stays on the record.

**Two `py/log-injection` alerts were LIVE when that claim was made.** Measured, not inferred:

```
$ gh api ".../code-scanning/alerts?state=open&ref=refs/pull/2100/merge"
11529 py/log-injection sev=medium src/kailash/utils/server_auth.py:359 sha=ca8f1200b
11528 py/log-injection sev=medium src/kailash/utils/server_auth.py:345 sha=ca8f1200b
```

They are not stale results from an older commit:

```
$ git cat-file -p ca8f1200b
tree 3d675d63e1c171eafd35440082166c1f6b6959f9
parent 5d0d77e42f88e2f3b13fa2c322acf3e82b18200f     <- main
parent 420224f23eca786b7c4da485fe24e1d77f3e5a55     <- this PR's head
```

**A green CodeQL check does not mean zero new alerts.** The check reports success while its own output says otherwise:

```
$ gh api ".../check-runs/94011880140"
name=CodeQL
conclusion=success
title=2 new alerts including 2 medium severity security vulnerabilities
```

It is green only because mediums do not meet its failure threshold. Reading `conclusion=success` as "no findings" is exactly the non-discriminating instrument that produced the false claim: that field returns `success` whether there are zero alerts or two mediums, so it could not have told me the difference.

## Root cause, and the fix

The sanitizer was correct at runtime the whole time — the barrier was simply in a **call form CodeQL's `py/log-injection` model does not recognize**. `safe_log_text` pre-compiled the pattern at module scope and called the `Pattern.sub` *method*; the sanitizer model recognizes the `re.sub` *module function*. So pass 2 barriered nothing as far as the scanner was concerned, and the taint stayed unbroken at both call sites.

Now:

```python
_LOG_LINE_TERMINATORS = r"[\r\n]"        # a string, not a compiled pattern
cleaned = re.sub(_LOG_LINE_TERMINATORS, "?", swept)
```

Behaviourally identical, recognized form. `test_line_terminator_barrier_uses_the_module_function_form` pins it so a future "optimization" that re-compiles the pattern fails loudly instead of silently re-opening both alerts.

**Whether this actually clears the alerts is measured on this push, not assumed.** I will report the result either way; if they persist, the disposition becomes a `zero-tolerance.md` Rule 1b deferral with all four artifacts, and the release-specialist signoff is not something I can self-grant.

## Also corrected: the surface count

The PR says **five** surfaces. There are **six**. `kailash/middleware/__init__.py:197` re-exports a *second* `create_gateway` — `kailash.middleware.communication.api_gateway.create_gateway` — under the same name as the fixed `kailash.servers.gateway.create_gateway`. It gated nothing. That is now fixed in this PR; details in the updated PR body.

A seventh (`visualization/api.py`) is real, lower severity, and explicitly deferred with a tracking issue rather than left inside an unqualified "five surfaces" claim.

esperie:
## CodeQL follow-up: measured, and now actually zero

I said I would report the result either way. Here it is, including the step where I was wrong.

**The `re.sub` hypothesis was REFUTED.** Switching from the compiled `Pattern.sub` method to the `re.sub` module function did **not** clear the alerts. On head `7fbcbc55f`:

```
11529 py/log-injection medium src/kailash/utils/server_auth.py:453 sha=7fbcbc55f
11528 py/log-injection medium src/kailash/utils/server_auth.py:439 sha=7fbcbc55f
```

Same two alerts, same two sinks — only the line numbers moved, because code was added above them.

**A third form cleared them.** `str.replace` on literals. On head `1e6f1d637`, two independent instruments:

```
$ gh api '.../code-scanning/alerts?state=open&ref=refs/pull/2100/merge'
(empty)

$ gh api '.../check-runs/94085541845'
conclusion=success
title=No new alerts in code changed by this pull request
```

The `title` field is the one that discriminates: on the previous head it read *"2 new alerts including 2 medium severity security vulnerabilities"* while `conclusion` read `success` in **both** cases. It can report either outcome, and it now reports the other one. `conclusion` alone would have looked identical throughout — which is the whole reason the original "all 4 fixed" claim was wrong.

Full result, all three forms runtime-identical (by the time pass 2 runs, the category sweep has already replaced every ASCII line terminator, so it is a no-op in every form — only the shape differs, and only the shape is what a taint analyzer reads):

| barrier form | alerts |
|---|---|
| compiled `Pattern.sub` method | 2 live |
| `re.sub` module function | 2 live |
| **`str.replace` on literals** | **0** |

So `py/log-injection`'s barrier model recognizes `str.replace` and neither regex form. Worth knowing repo-wide.

**Nothing was suppressed and no field was dropped.** `safe_log_text` still wraps every tainted component at both sinks; the alerts cleared because the sanitizer became *visible to the query*. `test_line_terminator_barrier_is_a_literal_str_replace` pins the shape, and says in its own docstring that it does not assert the scanner is satisfied — only the scanner can report that.

**No Rule 1b deferral.** I opened #2115 for that path while this was still in doubt; it is closed with the measurement rather than left open as a decorative tracking issue. The release-specialist signoff I flagged as un-self-grantable is therefore not needed.

**All 23 CI checks are green** on `1e6f1d637`, with nothing pending or failing.

Issue #2011, opened 2026-08-08
kaizen: a test writes kaizen_implementation_test.log into the repository root, voiding run fingerprints


esperie:
## Summary

Running the `packages/kailash-kaizen/tests/` tree creates an untracked **`kaizen_implementation_test.log`** (0 bytes) at the **repository root**, timestamped mid-run and never tracked in git.

## Why this is a defect, not housekeeping

A suite that dirties the working tree has two concrete consequences:

1. **It breaks the fingerprint protocol.** Comparing `git status --porcelain` before and after a run is how this repo establishes that a test result measures what it claims. A suite that writes into the tree makes BEFORE ≠ AFTER, which **voids every number from that run by protocol** — this fired for real and invalidated a complete tree measurement.
2. **It leaves a CI checkout dirty**, tripping any "working tree must be clean" release gate.

There is a related, independently-confirmed hazard: reading a suite while another process writes the tree produced `1 failed` / `13 failed` / `319 passed` on three consecutive reads at one commit. A test that writes into the repo root is one more way for a measurement to become unreproducible.

## Expected vs actual

**Expected:** tests write transient artifacts to `tmp_path` or a configured log directory.
**Actual:** a log file is created at the repository root.

## Severity

**LOW–MEDIUM** — no runtime impact; corrupts measurement discipline and CI cleanliness.

## Acceptance criteria

- [ ] The writing test uses `tmp_path` (or a configured log dir), not the repo root. Identifying the writer is part of the fix.
- [ ] `git status --porcelain` is byte-identical before and after the kaizen tree runs.

esperie:
**Re-triaged: `deferred-quality` → `bug`** (co-owner decision, cont-17 session).

Two consecutive `/sweep` passes measured this item as an **invalid defer**: per the sweep skill §3 a valid defer carries four sections (blocking-safety note, value-anchor, acceptance criteria, revisit trigger), and this issue carried 1–2 keyword matches with **no value-anchor**. Per `product-completion-first.md` MUST-2 that is silent deferral, which is BLOCKED — so the label was asserting tracked status the issue did not have.

Rather than re-author four sections for an item nobody picked up in two sessions, the co-owner directed re-triage as an ordinary bug, ranked into the normal queue by value. The `deferred-quality` label is retired on all six (#2011 #2017 #2018 #2019 #2020 #2021); zero open issues now carry it.

No technical finding here is withdrawn — the measurements recorded above stand.

PR #2101, opened 2026-08-12
fix(kaizen): the four observability flags now install the hooks they advertise (#2084)

### LINKED ISSUE #2084: HIGH: all four observability subsystems are advertised as enableable but unimplemented — create_observability() always returns an empty HookManager
## Summary

`SmartDefaultsManager.create_observability()` (`packages/kailash-kaizen/src/kaizen/smart_defaults.py:173-273`) advertises four observability subsystems. **None of them exists.** All four imports raise `ImportError`, all four are caught and silently skipped, and the function always returns a `HookManager` with zero hooks registered.

Found during #2070 (`LoggingHook` payload disclosure) while checking call sites of the class being fixed.

## Evidence

**1. The handler methods do not exist anywhere in `src/`.** Zero definitions each:

`start_trace`, `end_trace`, `record_start`, `record_end`, `log_start`, `log_end`

So even a corrected import path would fail on the next line. This is not a typo — the API was never implemented.

**2. All four imports fail, tested in the real `from X import Y` statement form:**

```
enable_tracing   ImportError (CAUGHT by source's except) -> silently skipped
enable_metrics   ImportError (CAUGHT by source's except) -> silently skipped
enable_logging   ImportError (CAUGHT by source's except) -> silently skipped
enable_audit     ImportError (CAUGHT by source's except) -> silently skipped
```

`observability/tracing.py` does not exist at all. `metrics.py`, `logging.py`, `audit.py` exist but define zero `*Hook` classes (`logging.py` has only `StructuredLogger` and `LoggingManager`).

**3. Runtime, on the DEFAULT path** (`AgentConfig(model=...)`, no observability flags touched):

```
Tracing hook not available, skipping
Metrics hook not available, skipping
Logging hook not available, skipping
Audit hook not available, skipping
flags  = True True True True
is_observability_enabled() = True
create_observability() -> HookManager
hooks actually registered = 0
```

All four flags default `True` (`agent_config.py:117,123,129,135`), and `agent.py:230` calls this on every agent construction. So `is_observability_enabled()` reports `True` for a subsystem set that registers nothing, and the closing `logger.info("Enabled observability: ...")` never fires because `enabled_systems` is always empty.

## Why this matters beyond the missing features

`enable_audit: bool = True` is documented as *"Enable compliance audit trails"*. Audit trails silently not recording is a compliance-shaped failure: the absence is invisible precisely when it matters.

This is `zero-tolerance.md` Rule 2 (stub) compounded by Rule 3 (silent fallback) — and the swallowed-`ImportError` shape is the same silent-degradation pattern #2070 fixed one layer down, here hiding four missing subsystems rather than one degraded control.

## Scope of THIS issue

This issue tracks **whether the four subsystems get implemented or removed**. It deliberately does not cover making the flags honest — that is being handled separately so the two are not blurred in the record: a fix that stops the flags lying does not make the features exist.

## Options

- **(a) Implement** Jaeger tracing, Prometheus metrics, structured logging, audit trails. Four separate lanes; effectively a roadmap.
- **(c) Remove** the flags and `create_observability`, per `zero-tolerance.md` Rule 6 ("if you cannot implement, remove it"), with the deprecation cycle Rule 6a requires for public config.

Audit trails are the strongest candidate to implement rather than remove, given the compliance framing in their own docstring.

## Acceptance criteria

- [ ] A decision is recorded per subsystem: implement or remove.
- [ ] If removed, the public config fields go through a deprecation cycle (Rule 6a) with a CHANGELOG migration entry.
- [ ] If implemented, each subsystem ships with the handler methods the call site already expects, and a test asserting hooks are actually registered — `len(hook_manager._hooks) > 0` is the assertion whose absence let this ship.

## Related

- #2070 — `LoggingHook` / `SecureLoggingHook` payload disclosure; this was found from its call sites.
- #2013 — documented security control that installs nothing (same class).

esperie:
## Summary

`AgentConfig.enable_tracing` / `enable_metrics` / `enable_logging` / `enable_audit` all default `True`, and `Agent.__init__` calls `SmartDefaultsManager.create_observability` on **every** construction. That function imported four hook classes from `kaizen.core.autonomy.observability.{tracing,metrics,logging,audit}` — a package containing **no hook classes at all** — caught every resulting `ImportError`, and returned an empty `HookManager`.

Measured on the untouched default path, before:

```
RESOLVED kaizen.__file__ = .../packages/kailash-kaizen/src/kaizen/__init__.py
flags = True True True True
is_observability_enabled() = True
create_observability() -> HookManager
hooks actually registered = 0
```

`enable_audit`, documented as *"Enable compliance audit trails"*, recorded nothing. After:

```
hooks actually registered = 18
registered_hook_names() = ['audit_trail_hook', 'logging_hook', 'metrics_hook', 'tracing_hook']
audit file exists = True
```

## The verdict this PR settles

#2073 was held pending one question: **can the existing managers be wired, or must the flags be made honest by raising?** They can be wired. The subsystems were never missing, only **mislocated**:

```
hooks/builtin/tracing_hook.py:50:class TracingHook(BaseHook)
hooks/builtin/metrics_hook.py:29:class MetricsHook(BaseHook)
hooks/builtin/logging_hook.py:24:class LoggingHook(BaseHook)
hooks/builtin/audit_hook.py:16:class AuditHook(BaseHook)
```

(`grep -rn "^class .*Hook"` in `observability/` returns **0**.) Each declares `events = list(HookEvent)` and an async `handle()`, and `HookManager.register_hook()` is the contract they were written for — **its own docstring example is the exact wiring**, and `BaseAgent.enable_observability()` already does it for tracing. The old call site registered per-method handlers (`start_trace`, `record_start`, …) with **zero definitions anywhere in the source tree** (scoped to `src packages/*/src`; the two `record_start` hits are `kailash/infrastructure/execution_store.py`, unrelated), so a corrected import path would still have failed on the next line.

**So #2073 is superseded and can be closed.** Its `#2070` half already landed on main as #2094; everything unique remaining in it — the `Optional[bool]` tri-state, `errors.py::ObservabilityNotImplemented`, `test_observability_flags_fail_loudly.py` — is exactly the half this replaces. The flags become honest by becoming **true**, not by raising.

## Two layers, because fixing one alone would have been a stub

**Layer 1 — registration.** `create_observability` now registers a real hook per enabled flag.

| Flag | Hook | Requires |
| --- | --- | --- |
| `enable_logging` | `LoggingHook` | core deps only |
| `enable_audit` | `AuditTrailHook` **(new)** writing `audit_log_path` | core deps only |
| `enable_metrics` | `MetricsHook` | `kailash-kaizen[observability]` |
| `enable_tracing` | `TracingHook` | `kailash-kaizen[observability]` |

**Layer 2 — invocation.** `BaseAgent.__init__` accepts a `hook_manager`, stores it, and then `trigger_hook` / `register_hook` / `get_hook_stats` gated on `config.hooks_enabled` — a **separate** opt-in defaulting `False` that `Agent._convert_to_base_agent_config` never sets. Layer 1 alone would have registered four hooks onto a manager that was never triggered: the audit trail would still have recorded nothing on the path users take. The three gates now test `self._hook_manager is None`, which `__init__` already resolves from **both** inputs (`zero-tolerance.md` Rule 3c). **This is not equivalent to the old predicate** — see the BREAKING note below.

## Decisions worth arguing with

- **`AuditTrailHook` rather than the existing `AuditHook`.** `AuditHook` wraps the PostgreSQL-backed security `AuditTrailProvider`, which a zero-config default path has no connection for. `AuditTrailManager` + `FileAuditStorage` honours `audit_log_path` and needs only `anyio` (a core dep) but is not a hook; this is that bridge. It records event **structure**, never payload values, so enabling compliance audit does not itself become a disclosure channel. **Both** payload-bearing fields are reduced: `context.data` to its sorted key names, and `context.metadata` through the same `summarize_payload` helper `LoggingHook` uses, to `{count, keys}`. What reaches the file verbatim is the audit record itself — agent id, event name, trace id, timestamp, success/failure verdict. (The `metadata` half was **wrong in the first revision of this PR** and is fixed in `24f72f4b9`; see Review findings below.)
- **A missing optional dependency is loud, once.** `"Metrics hook not available, skipping"` named neither what stopped working nor how to fix it. The replacement names the flag, the extra, and `enable_metrics=False` as the deliberate opt-out. Once per process via `lru_cache`, because this path runs per agent construction and a warning repeated on a hot path is one operators filter — which would turn the loud-warning remedy back into no remedy.
- **`None` when nothing installs**, not an empty `HookManager`. Returning the empty manager is what made the original defect invisible.
- **`tracing_endpoint` default 16686 → 4317.** 16686 is the Jaeger **web UI** and accepts no OTLP traffic. Nothing read the field while it was inert, so **no existing behaviour changes** — but now that spans are exported, the old default would have dropped every one at the socket while tracing reported itself enabled. An endpoint still pointed at 16686 warns once, naming both ports.
- **`metrics_port` binds no listener.** It configures `MetricsEndpoint`, which the operator starts. Opening a network port is not something a zero-config default may do on its own; the field's docstring now says so instead of implying otherwise.
- **The banner reads reality.** `_show_observability_status` read the flags, so a default agent printed four subsystems with endpoints and file paths while none was registered. It now renders from `registered_hook_names()`. Not redundant with layer 1: a subsystem whose optional dependency is missing keeps its flag `True`, so a flag-driven banner still reports it as running.
- **`HookManager.registered_hook_names()` is new.** `get_stats()` reports EXECUTION counts, so a manager that had registered nothing was indistinguishable from one whose hooks had not yet fired. No caller, banner, or test could ask what was installed — that blind spot is why this shipped.

## Fail-first evidence

Both fixes verified RED by reverting only the changed body, keeping the surrounding scaffolding, so the failure isolates the change under test.

**Layer 1** — pre-fix `create_observability` body, new suite: **9 failed, 3 passed**

```
FAILED test_default_path_registers_hooks - AssertionError: is_observability_enabled() reported True, create_observabil...
FAILED test_logging_flag_registers_logging_hook - AssertionError: assert 'logging_hook' in set()
FAILED test_audit_flag_registers_audit_trail_hook_and_creates_the_file - AssertionError: assert 'audit_trail_hook' in set()
FAILED test_metrics_flag_registers_metrics_hook - AssertionError: assert 'metrics_hook' in set()
FAILED test_tracing_flag_registers_tracing_hook - AssertionError: assert 'tracing_hook' in set()
FAILED test_registered_hooks_are_real_hook_objects - assert []
FAILED test_disabled_subsystem_is_absent_while_others_register - AssertionError: assert 'audit_trail_hook' in set()
FAILED test_missing_dependency_warns_loudly_and_names_the_remedy - AssertionError: Tracing hook not available, skipping
FAILED test_missing_dependency_warns_once_per_process - AssertionError: []
```

The 3 that passed are the negative controls, which must pass in **both** states.

**Layer 2** — pre-fix `hooks_enabled` gates: **3 failed, 2 passed**

```
FAILED test_supplied_hook_manager_fires_without_hooks_enabled - AssertionError: a hook_manager was supplied to BaseAgent and its hook was n...
FAILED test_stats_reflect_a_supplied_manager - AssertionError: assert 'recording_hook' in {}
FAILED test_unified_agent_default_path_fires_its_observability_hooks - AssertionError: Agent() was constructed with enable_audit=True (the default...
```

**Banner** — reverted to the flag-driven form: **2 failed, 1 passed**

```
FAILED test_banner_lists_only_installed_subsystems - AssertionError: ✅ Observability:
FAILED test_banner_does_not_name_subsystems_of_a_custom_manager - AssertionError: assert 'custom hook manager' in '✅ Observability:\n   • Ja...
```

GREEN after: **23 + 5 passed**.

## Both polarities pinned

| Test | Property |
| --- | --- |
| `test_default_path_registers_hooks` | the AC assertion — `len(hook_manager._hooks) > 0` on the default path |
| `test_audit_trail_records_a_real_entry_on_a_real_event` | **recording**, not registration — a real event lands a real JSONL line |
| `test_audit_trail_records_structure_not_payload_values` | credential + prose absent, key names **present** — an empty file cannot pass |
| `test_registered_hooks_are_real_hook_objects` | placeholders cannot satisfy the count |
| `test_all_disabled_returns_none` | **negative control** — unconditional registration fails here |
| `test_no_manager_and_not_enabled_stays_inert` | **negative control** — unconditionally enabling hooks fails here |
| `test_explicit_false_emits_no_warning` | **negative control** — only meaningful because the autouse fixture clears the one-time cache |
| `test_unified_agent_default_path_fires_its_observability_hooks` | end to end on `Agent(model=...)` with nothing else set |

No `Mock` anywhere: a `MagicMock` manager records a `trigger` call whether or not a handler ran, so it cannot distinguish the two states this PR is about.

## Verification

- Local CI parity, both steps of `test-kailash-kaizen.yml`, post-rebase: ruff clean; **1643 passed** (llm + cross-SDK); **6698 passed, 87 skipped** (expanded FAST unit tiers). Zero failures.
- Full `tests/unit` sweep (11,272 tests, 14m29s) — **11121 passed**, 4 failures, none reachable from this diff:
  - `test_policy_engine.py::test_performance_under_5ms` — 5.099 vs a 5.0 wall-clock threshold under 2 concurrent pytest processes; **passes in isolation**, and `governance/policy_engine.py` has zero references to hooks/observability/smart_defaults.
  - `test_performance_comprehensive.py` ×3 — spawn a bare `python3 -c "import kaizen"` with no worktree path, so they resolve site-packages, where `kailash-enterprise` and `kailash` both claim the namespace: `ModuleNotFoundError: No module named 'kailash.kaizen'`. Local install artifact; none of the three is in CI's path list.
- `pre-commit` on every changed file at each commit, including the Tier 1 gate.
- The regression suite was **moved** to `tests/unit/observability/` after checking the workflow: its expanded step enumerates subdirectories plus exactly two root-level files, so a new `tests/unit/*.py` would have passed locally and never gated a PR.

## Review findings, fixed in this PR

Four findings from review. All three HIGH/MEDIUM defects were **newly reachable because the fix worked** — pre-fix the whole audit branch died on a caught `ImportError` ahead of every line below, so none of this code had ever executed. Each is pinned by a test verified RED against the unfixed body.

**HIGH-1 — `Agent()` raised `OSError` on a read-only filesystem.** The `mkdir` in the audit branch sat outside any `try`, and `agent.py` calls `create_observability` at `:230`, *before* `Agent.__init__`'s own `try:` at `:264`. On Kubernetes `readOnlyRootFilesystem: true`, distroless images, or Lambda's `/var/task`, `Agent(model=...)` went from working to failing outright on upgrade. Now catches `OSError`, warns once naming `audit_log_path` and `enable_audit=False`, and skips only the audit hook.

RED without the guard — note the second failure is `Agent()` itself:

```
FAILED test_unwritable_audit_path_warns_loudly_and_leaves_the_rest_installed - PermissionError: [Errno 13] Permission denied: '/private/var/...'
FAILED test_agent_constructs_on_a_read_only_filesystem - PermissionError: [Errno 13] Permission denied: '.kaizen'
```

**HIGH-2 — the audit file was world-readable.** `touch()` with no `mode=` is 0644 under a normal umask and `mkdir` 0755, on the file whose own docstring cites SOC2/GDPR/HIPAA. Now 0o600/0o700, `chmod`-pinned and read back from `stat()` — a mode stripped by umask is indistinguishable from one never set, so the fix is asserted on the real mode. Pre-existing paths are tightened too, announced at WARN: restricting it to created paths would mean the fix never reaches an upgraded install, since the file is created once and reused forever.

RED without the enforcement:

```
FAILED test_audit_file_and_directory_are_owner_only - AssertionError: audit file is 0o644, expected 0o600
FAILED test_audit_directory_created_by_the_storage_is_owner_only - AssertionError: audit directory is 0o755, expected 0o700
FAILED test_pre_existing_audit_file_is_tightened_and_the_change_is_announced - AssertionError: assert 436 == 384
```

**MEDIUM-1 — the "structure, never values" claim was FALSE.** `details` was keys-only, but `metadata=context.metadata or None` passed through whole into `AuditEntry`, which `json.dumps(asdict(entry))` serialises verbatim. The tell was the asymmetry: sibling `LoggingHook` already reduces metadata via `summarize_payload`. Latent (no in-tree caller populates it) but it is a documented public kwarg on both `HookManager.trigger` and `BaseAgent.trigger_hook`, whose intended users are downstream consumers.

The actual audit line, pre-fix:

```json
{"agent_id": "a", "action": "pre_tool_use", "details": {"data_keys": ["k"], ...}, "result": "success", "metadata": {"tenant_token": "sk-live-AUDITMETADATA2084SECRETVALUE", "principal": "dr.reyes@stjude"}}
```

and after:

```json
{"agent_id": "a", "action": "pre_tool_use", "details": {"data_keys": ["k"], ...}, "result": "success", "metadata": {"count": 2, "keys": ["principal", "tenant_token"]}}
```

`test_audit_trail_records_structure_not_payload_values` previously put its sentinel in `data` only, so it could not catch this; it now carries a sentinel in `metadata` too and REDs against the old passthrough. Absent metadata still serialises as `null` rather than an empty summary, so no in-tree caller's on-disk shape changes.

**MEDIUM-2 — `hooks_enabled=False` stopped being an off-switch for `register_hook`.** Correctly identified: gating on `_hook_manager is None` is *not* equivalent to the old predicate. Smart defaults always supplies a manager, so `_hook_manager` is non-`None` while `hooks_enabled` remains at its documented `False` default, and `register_hook(evt, handler)` now succeeds where it previously raised.

| `hooks_enabled` | manager supplied | before | after |
| --- | --- | --- | --- |
| `False` | no | raises | raises |
| `True` | either | registers | registers |
| `False` | **yes** | **raises** | **registers** |

**Route taken: documented as `### Changed (BREAKING)` rather than reverted**, and pinned by a new test rather than left as prose. Restoring the flag as an override is not available: `False` is the field's DEFAULT, so a dataclass cannot distinguish "explicitly disabled" from "never mentioned" — honouring it would reject the manager `SmartDefaults` supplies on every default `Agent` construction, reinstating the exact no-effect-kwarg defect this PR exists to fix. To keep hooks off, pass no `hook_manager`.

## Round-2 review, also fixed here

An independent review of the fixes above found four more, three of them the same bug class — a subsystem's misconfiguration taking down `Agent()` construction, or a mode not actually enforced. Fixed in `ffc51bc32` rather than filed, per `autonomous-execution.md` MUST-4 (same class, within shard budget). Each REDs under a targeted mutation.

- **A typo'd `tracing_endpoint` also broke construction.** The `ImportError` guard covered the `TracingHook` import but not `_parse_otlp_endpoint`, whose `urlparse(...).port` raises `ValueError` on a non-numeric port — ahead of the same `try`. The branch already soft-warned for a *semantically* wrong port (16686, the Jaeger web UI); a *syntactically* wrong one was fatal.
- **The 0o700 pin was re-permissioning directories it did not create.** For a bare filename `file_path.parent` is `.` — the process working directory — and for a configured `audit_log_path` it is frequently a shared location like `/var/log/kaizen`. Chmodding either is outside this class's remit, and it is not what protects the record: the file mode is. Directories are now pinned only when created; the pre-existing **file** tightening is unchanged.
- **`_enforce_mode` was silent on a loosening.** It warned only when the old mode had group/world bits, so an operator-set `0o400` silently became `0o600` — which adds write — from a function whose docstring claimed it announces every correction. Both directions now warn. **The suite could not have caught this**: its negative control used `0o600`, which exits early before the warn predicate, so narrowing the predicate to `if not (current & 0o077): return` left every test green. The `0o400` case that discriminates is now in.
- **`ObservabilityManager.__init__` had the same unguarded `OSError`**, via a bare `AuditTrailManager()` falling back to CWD-relative storage, reached from `BaseAgent.enable_observability()`. Guarded; and `_enabled_components["audit"]` now reports what installed rather than what was requested.
- **`details["data_keys"]` was hand-rolled** `sorted(context.data.keys())` while metadata went through `summarize_payload`. The helper coerces keys to `str`, so mixed key types summarise instead of raising `TypeError` and failing the append. Leaving one of two payload fields on a private implementation is the asymmetry that produced the metadata bug in the first place.

**One reported finding did not reproduce.** Review flagged an unguarded `mkdir` in `create_checkpointing` as the same read-only-filesystem break. It is not reachable: `kaizen.memory.checkpoint` does not exist anywhere in the tree, so the import above it always raises and the `mkdir` is dead code. Guarding it would ship a branch nothing can enter (`rules/orphan-detection.md` §1). The real defect is larger — `enable_checkpointing` is a **fifth** advertised-but-unimplemented subsystem, same class as #2084 — and is filed as **#2111**, which carries the requirement that the guard land with the implementation.

The CHANGELOG was corrected for all of the above plus two claims this round caught: the BREAKING entry named only `register_hook` when the same predicate swap also changed `trigger_hook` (no-op → fires every registered handler, the larger break) and `get_hook_stats`; and the audit "not a disclosure channel" line sat 100 lines from the `log_payload_keys` entry arguing key names *are* a bounded leak. Both are true — the two sinks are asymmetric on purpose, since an audit record stripped of event shape records that something happened and nothing about what — so the trade is now stated rather than implied.

## Follow-ups filed, deliberately not fixed here

All are newly reachable rather than pre-existing (the branch never executed before this PR), so `zero-tolerance.md` Rule 1a's "same on main" does not apply — they are follow-ups because they exceed this shard, not because they are inherited:

- **#2109** — durability: the 0.5s hook timeout drops audit records with only a `logger.error`; concurrent appends are unlocked; malformed lines are silently skipped on read.
- **#2111** — `enable_checkpointing` is a fifth advertised-but-unimplemented subsystem (found by round-2 review; see above).
- **#2110** — growth and location: the audit file is unbounded and unrotated (one line per event with `events = list(HookEvent)`), and `audit_log_path` defaults to a path **relative to CWD**, so a default agent writes `.kaizen/` into whatever directory it happens to run in (same class as #2011 stray files).

## Also found, NOT fixed here

`Agent.__init__` takes `**kwargs` and never reads it, so any unrecognised keyword — including a misspelled `audit_log_path` — is silently swallowed. Same class as Rule 3c but a different surface with a far larger blast radius (rejecting unknown kwargs turns every stray argument into a `TypeError` across all call sites). It needs its own issue and its own deprecation path rather than a quiet expansion of this one.

## Related issues

Fixes #2084
Supersedes #2073 (its #2070 half already landed as #2094)
Related: #2070 / #2094, #2013 — documented control that installs nothing, same class


esperie:
## Follow-up: the auto-wired logger no longer discloses payload key names

Review push-back, accepted. The first cut left `LoggingHook(include_data=True)` on the auto-wired path, reasoning that #2070/#2094 made the hook emit key NAMES rather than values, so it could not disclose. Two holes in that:

1. **Key names are a bounded leak, not an absent one.** `ssn`, `patient_diagnosis`, `termination_reason`, `acme_contract_value` disclose schema, subject matter, and often that such a record exists. Measured on the default path before this commit:

   ```
   INFO logging_hook.py:223 [pre_agent_loop] Agent=agent-2084 TraceID=904fba5c-...
        DataKeys=['api_key', 'patient_diagnosis'] (2 field(s))
   ```

   That is a **default-on** sink, running on every agent construction in every downstream consumer. Wiring a dormant sink onto a default-on path is a disclosure change and gets the conservative default.

2. **The safety rested on a sibling's behaviour that nothing here pinned.** If `LoggingHook` ever regained value logging, this suite stayed green while the sink started emitting payloads — a non-discriminating instrument.

### What changed

New `AgentConfig.log_payload_keys`, default `False`, drives `include_data` on the auto-wired hook. Lifecycle logs still carry event, agent id and trace id. Payload VALUES remain unreachable at this level either way — `log_full_payloads` is a separate, additionally DEBUG-gated opt-in.

### Both new tests verified discriminating, by mutation

| Mutation | Test that reds |
| --- | --- |
| force `include_data=True` back on the auto-wired path | `test_default_path_logs_neither_values_nor_key_names` — `assert 'patient_diagnosis' not in ...` |
| regress `LoggingHook`'s text branch to render `safe_context.data` | `test_opting_into_payload_keys_still_withholds_values` |

Complementary, not redundant: neither reds for the other's mutation. The second is the pin that was missing — the names-not-values property is now asserted **in this PR** rather than inherited. `logging_hook.py` was restored byte-identical after the mutation (`git diff --stat` empty).

### Review item — the `[observability]` extra WARN, verbatim

Five agent constructions with `prometheus_client` unavailable produce **one** warning:

```
constructions = 5; metrics warnings emitted = 1

Metrics is enabled (enable_metrics=True) but could not be loaded, so NOTHING will be
recorded for it. MetricsHook requires the optional 'observability' extra (structlog /
prometheus-client / opentelemetry). Install it with: pip install
'kailash-kaizen[observability]' Set enable_metrics=False to disable it deliberately and
silence this warning.
```

Exact install string included; `lru_cache` keyed per subsystem, so once per process rather than per construction. Pinned by `test_missing_dependency_warns_once_per_process`, whose autouse `cache_clear()` fixture is load-bearing — without it `test_explicit_false_emits_no_warning` could pass on a warning an earlier test had already consumed.

### Review item — `is_observability_enabled()` honesty, and the one case it can still lie

The primary assertion is `len(hook_manager._hooks) > 0` on the default path, RED-verified against the unmodified body.

Stated plainly, as asked: **`is_observability_enabled()` still ORs the four flags, and I did not change that** — an `AgentConfig` holds no hook manager, so it cannot see what installed. The residual case is **`enable_metrics=True` alone (or tracing alone) with the `observability` extra absent**: the OR returns `True` with zero hooks behind it.

How the code signals it rather than letting the lie through: the **manager** is made the authority. When every enabled subsystem fails to install, `create_observability` returns `None` — not an empty `HookManager` — and warns `"Observability was enabled but no subsystem could be installed"`. So the caller branching on the manager gets the truth even though the flag OR does not. Pinned by `test_enabled_with_zero_installable_subsystems_returns_no_manager`, which asserts `is_observability_enabled() is True` **and** `hook_manager is None` in the same test, so the gap is documented by the suite rather than papered over.

`enable_logging` and `enable_audit` depend only on core dependencies, so on any install at least those two register — the residual case needs both of them explicitly disabled.

### Verification

- Local CI parity re-run post-change: **6693 passed, 86 skipped**, 1 failure — `test_concurrent_faster_than_sequential` (0.236s vs a 0.090s wall-clock threshold), which **passes 5/5 in isolation** and whose test file and both source modules (`l3/plan/executor.py`, `l3/plan/types.py`) contain zero references to hooks / observability / smart_defaults / AgentConfig. Scheduler contention under a 6700-test run, not reachable from this diff.
- Hook + observability suites: **280 passed**.

esperie:
## Review item 4: the audit trail was landing world-readable — fixed

Asked whether `audit_log_path` is inside the CWD and what mode the file is created with. Both answers were bad, and this PR is what made the path live, so it is fixed here rather than deferred.

Measured before the fix, in a scratch CWD:

```
default audit_log_path = '.kaizen/audit.jsonl' -> RELATIVE
created at = /.../tmpmimk4zc3/.kaizen/audit.jsonl exists = True
FILE mode = 0o644
DIR  mode = 0o755
```

So: **relative to the process CWD** (the #2011 defect class), and **0o644 — readable by every local account**. The trail records which agent did what, when, and the SHAPE of every payload (event name, payload key names, trace id, agent id). Not payload values, but the same bounded-leak class as the logging change earlier in this PR — and I had applied the conservative treatment to the logger while leaving the file that logger's sibling writes at 0o644.

### Fix

Paths `FileAuditStorage` **creates** are now `0o600` / `0o700`, pinned with an explicit `chmod` — both `touch(mode=...)` and `mkdir(mode=...)` are masked by umask, so a umask of 0 would otherwise leave the file group/world readable. A **pre-existing** path is left exactly as the operator configured it, with its own negative control; silently re-permissioning a file this class did not create is its own surprise.

The duplicate `mkdir` in `create_observability` is dropped — pre-creating the directory there handed it the default umask mode and left the tightening with nothing to do. `FileAuditStorage` now owns both halves, so they cannot drift.

### Write failure: loud, and pinned

The exception propagates to `HookManager`, which records `HookResult(success=False)`, logs the failure, and calls the hook's `on_error`. Pinned by a test that makes a **real** append fail — it replaces the audit file with a directory so `open()` raises — rather than mocking it.

### All three verified discriminating, by mutation

| Mutation | Test that reds |
| --- | --- |
| remove the mode pinning | file test at `0o644`, directory test at `0o755` — the exact pre-change values |
| wrap the audit append in `except Exception: pass` | `test_audit_write_failure_is_loud_not_silent` — "the failed audit append was reported as success" |

`audit_trail_hook.py` and `audit.py` both restored byte-identical after their mutations (`git diff --stat` empty).

### One-time-warning latch — checked for the SDK-self-consumption shape, not present

The `lru_cache` latch on the missing-dependency WARN could be consumed by the SDK's own construction, leaving a genuinely misconfigured user warned nothing. Verified it is not:

```
latch state after a bare import of kaizen: CacheInfo(hits=0, misses=0, maxsize=None, currsize=0)
```

`create_observability` has exactly one non-docstring caller — `agent.py:230`, inside `Agent.__init__` — plus `create_all_components`, also user-facing. Nothing constructs `SmartDefaultsManager()` at import time. The latch is keyed per subsystem, so it stays one warning per subsystem per process. Structurally the risk would return if the SDK ever constructs an internal agent during import; it does not today.

### Verification

CI-parity re-run: **6734 passed, 85 skipped, zero failures**. Hook + observability suites: **284 passed**. CI green on `4bac070d5`.

Issue #2102, opened 2026-08-12
APIGateway session route accepts a body-supplied user_id when auth is enabled and no credential is presented


esperie:
Split out of the #2047 lane (w2d). The **stub half is fixed**; this issue is the **breaking half** that did not fit the same shard.

## What was fixed already

`src/kailash/middleware/communication/api_gateway.py` — `get_optional_current_user` returned `None` unconditionally behind the comment *"For now, return None to avoid complex auth setup"*, so with `enable_auth=True` (**the default**) the authenticated principal was **never resolved** and `request.user_id` — a POST body field — was always the session identity. `MiddlewareAuthManager.verify_token` / `verify_api_key` existed the whole time and were simply not called.

That is now implemented: a presented bearer token or `X-API-Key` is verified and the resolved principal wins. A rejected credential is logged rather than silently swallowed (`zero-tolerance.md` Rule 3), and a principal resolving to an empty/None `user_id` no longer silently produces an ownerless session.

## What remains — and why it is a separate decision

The dependency is still **optional by design**: a request carrying **no credential at all** resolves to `None`, and `POST /api/sessions` then falls back to `request.user_id` from the body. So with auth enabled, an unauthenticated caller can still open a session under any name it chooses.

That is the same class as #2047 — an identity taken from a body field rather than derived from an authenticated session (`security.md` § Enforcement-Surface Parity → identity-derivation parity) — but closing it means **`POST /api/sessions` returns 401 when `enable_auth=True` and no credential is presented**, which is a breaking change to the route contract and touches ~10 integration test files that create sessions without credentials. It belongs in its own shard with its own migration note, not smuggled into an unrelated PR.

## Acceptance criteria

- [ ] With `enable_auth=True` and no credential, `POST /api/sessions` refuses (401) rather than trusting `request.user_id`.
- [ ] With `enable_auth=False`, the route behaves exactly as today (the body value is the identity, and that is the documented contract).
- [ ] Sibling sweep: every other route reached via `get_optional_current_user` or any similar optional-auth dependency gets the same determination, recorded per route even where left as-is (`security.md` § Enforcement-Surface Parity).
- [ ] Behavioural regression test pinning BOTH polarities — credential present → the server-derived principal owns the session; credential absent under `enable_auth=True` → 401 — driven through the HTTP surface, not by calling the dependency directly.
- [ ] CHANGELOG entry under `### Changed (BREAKING)` with the migration note.

## Related

[...]

Issue #2060, opened 2026-08-10
EnterpriseAuthProviderNode is non-functional: 10 of 14 execute_async call sites target nodes that do not define it


esperie:
## Summary

`EnterpriseAuthProviderNode` calls `execute_async(...)` on its collaborator nodes at **14 sites**. Only 4 of those targets define the method. Every other call raises `AttributeError` at runtime, so **`EnterpriseAuthProviderNode` cannot complete `authenticate`, `authorize`, `logout` or `validate` at all** — the node is non-functional end to end, not merely degraded.

Surfaced while fixing #2026. It matters beyond "a broken node": it currently **masks** the identity-binding fixes that PR #2035 landed, so those fixes are correct in source but unreachable in practice, and will only start taking effect once this is closed. A reader of #2035 would reasonably assume they are live today. They are not.

## Evidence

Method availability on current `main` (`aba8a0878`):

| node | `execute_async` | `async_run` | `execute` |
|---|---|---|---|
| `SessionManagementNode` | **False** | True | True |
| `SSOAuthenticationNode` | **False** | True | True |
| `DirectoryIntegrationNode` | **False** | True | True |
| `SecurityEventNode` | **False** | **False** | True |
| `AuditLogNode` | **False** | **False** | True |
| `HTTPRequestNode` | **False** | — | True |
| `MultiFactorAuthNode` | True | True | True |

Call sites in `src/kailash/nodes/auth/enterprise_auth_provider.py`:

```
4  self.session_node.execute_async      -> MISSING
4  self.mfa_node.execute_async          -> ok
2  self.sso_node.execute_async          -> MISSING
1  self.directory_node.execute_async    -> MISSING
1  self.security_logger.execute_async   -> MISSING
1  self.audit_logger.execute_async      -> MISSING
1  self.http_client.execute_async       -> MISSING
```

Reproduction — a **correctly signed, unexpired, correct-issuer** JWT, which should authenticate:

```python
import asyncio, time, jwt as pyjwt
from kailash.nodes.auth.enterprise_auth_provider import EnterpriseAuthProviderNode

secret = "signing-secret"
n = EnterpriseAuthProviderNode(
    name="eap", enabled_methods=["jwt"],
    adaptive_auth_enabled=False, risk_assessment_enabled=False,
    jwt_config={"secret": secret, "issuer": "iss"},
)
tok = pyjwt.encode({"sub": "alice", "exp": int(time.time()) + 3600, "iss": "iss"},
[...]

PR #2103, opened 2026-08-12
fix(auth)!: MFA actor model, sink-level log sanitizing, and an audible JWT issuer warning

### LINKED ISSUE #2040: security: caller-controlled user_id interpolated into a log line (log injection), and no public value-sanitizer is exported
Surfaced during the #2015 burn-down and deliberately **not** fixed there — different bug class, and closing it properly needs an API decision.

## The defect

`src/kailash/middleware/communication/api_gateway.py` interpolates a caller-controlled `request.user_id` directly into a log line.

A caller who controls that value controls log **structure**, not just content: embedded newlines let them forge additional log records, and terminal escape sequences or delimiter characters can corrupt downstream log parsing and SIEM ingestion. This is distinct from #2015 (which was about sensitive values travelling *outward* to clients); here untrusted input travels *inward* into a trusted sink.

## Why it needs a decision, not just a patch

`secure_logging` exposes no **public** value-sanitizer. The available helper, `_safe_identifier`, is private.

So there are three options and they are not equivalent:

1. **Export a public value-sanitizer** from `secure_logging` (promote `_safe_identifier` or add a documented sibling). Cleanest for the many future call sites that will need it; it is a public-API addition and should be designed once rather than per-site.
2. **Use the private helper at this call site.** Fastest, but reaching into a private name is exactly the pattern that later blocks refactoring, and it does not help the next site.
3. **Structured logging** — pass `user_id` as a field rather than interpolating it, so the logging backend handles escaping. Best where the sink supports it; needs checking that every configured handler does.

Recommendation: **option 1**, because this will not be the only site — a repo-wide sweep for caller-controlled values reaching `logger.*` interpolation should follow, and it needs one sanctioned helper to route through (`security.md` § Credential Decode Helpers applies the same one-shared-helper logic).

## Acceptance criteria

- [ ] A **public**, documented value-sanitizer exists for untrusted values entering log lines (newlines, control characters, ANSI escapes neutralised).
- [ ] The `api_gateway.py` `user_id` site routes through it.
- [ ] Repo-wide sweep for sibling sites — caller-controlled request fields (`user_id`, tenant, header values, path segments) interpolated into `logger.*` — with the list reported even where not fixed.
- [ ] Behavioural regression test: a `user_id` containing `\n` and control characters produces exactly ONE log record with the injected content neutralised. Asserting on source text is not sufficient.

## Related issues

Surfaced during #2015 (PR #2034). Sibling class to #2027 (sensitive values reaching log sinks) — that one is values that should not be logged; this one is untrusted structure in what is logged.

### LINKED ISSUE #2047: MultiFactorAuthNode has no actor: every action authorizes on a caller-supplied user_id
## Summary

`MultiFactorAuthNode` has **no actor**. `user_id` names the *subject* of an action, never the caller, and there is no session, principal, or credential input on any of its ~15 dispatchable actions. The node therefore cannot distinguish "this end user acting on themselves" from "someone acting on another user's account", and several actions **issue bearer credentials** or **destroy existing factors**.

Found by four consecutive adversarial review rounds on #2026 (PR #2035). Each round patched the specific action the previous round exploited; each next round found another one. The reviewer's closing assessment on round 4:

> Three rounds have each patched the specific action the previous round exploited. `mfa.py` has fifteen dispatchable actions keyed on a caller-typed `user_id` string with no principal anywhere in the module, and the last five reviews have each found a different one of them to be a credential-issuing or state-destroying primitive. Patching the sixth will not end this.

This issue is the root. #2026 fixed the *stub* class (forged tokens, logged OTPs, hardcoded codes) and hardened the individual actions it could within one shard; it deliberately did **not** attempt this, because it is a node API change affecting every action and every consumer.

## The gap

Actions that issue a bearer credential to whoever asks, keyed only on a caller-supplied `user_id`:

| Action | Issues |
|---|---|
| `setup` | TOTP secret + backup codes |
| `generate_backup_codes` | 10 backup codes (accepted directly by `verify`) |
| `trust_device` | a trust token that makes `check_device_trust` return `skip_mfa: True` |
| `reset` | fresh secret + fresh backup codes |
| `initiate_recovery` | a 24h recovery token |

Actions that destroy or degrade another account's factors: `revoke`, `disable`, `reset`, and `setup` (overwrite).

After #2026 the destructive ones require `admin_override` — **but `admin_override` is an ordinary caller-supplied boolean**, not an authentication control. It separates administrative shapes from end-user shapes for a host that already authorizes the caller; it stops nothing on its own. #2026 documents this explicitly rather than leaving it implying a guarantee it does not provide.

## Why this is not exploitable today by an end user

The node is a workflow node. In every intended deployment the host (Nexus gateway or equivalent) authenticates the request and decides which `user_id` it may act on — it has to, because `user_id` is the only identity input for *every* action including read-only `status`. So these are reachable only by a caller that can pass an arbitrary `user_id` to an administrative action.

That is exactly the assumption this issue asks us to stop relying on, and it is undocumented at the API surface (#2026 adds a class-level warning as an interim measure).

## Proposed fix

Introduce an authenticated-actor input and authorize the `(actor, action, subject)` triple inside the node:

- Add a required `actor` / `actor_session_id` parameter, validated against `SessionManagementNode` (or an injected authorizer protocol).
- Default policy: an actor may act on themselves for `setup`/`verify`/`status`/`generate_backup_codes`/`trust_device`/`initiate_recovery`; acting on another subject requires an actor holding an administrative capability.
- Replace `admin_override` with the capability check; keep the kwarg for one deprecation cycle emitting a `DeprecationWarning` (`rules/zero-tolerance.md` Rule 6a).
- Collapse `async_run` onto `run` — the two dispatchers have already diverged (differing action sets, differing gates, two different trusted-device stores), and a gate present in one and absent in the other is not a gate.

## Acceptance criteria

- [ ] Every action requires an authenticated actor; no action accepts a bare `user_id` as authority.
- [ ] Acting on a subject other than the actor requires an explicit administrative capability, verified, not asserted.
- [ ] `admin_override` deprecated with a shim + CHANGELOG migration entry.
- [ ] `run` and `async_run` share one dispatcher and one trusted-device store.
- [ ] Behavioural tests, fail-first, for: credential issuance for another subject; factor destruction for another subject; the `revoke → setup → verify` chain; `trust_device → check_device_trust`.
- [ ] The interim class-level warning added by #2026 is replaced by the real guarantee.

## Related

- #2026 / PR #2035 — the stub-removal work that surfaced this, and the four review rounds.

### LINKED ISSUE #2088: Log-injection sanitizing belongs in SecurityEventNode.execute / AuditLogNode.execute, not at N call sites
Filed from PR #2066, with direct evidence that the per-call-site approach drifts.

## The problem

Caller-controlled values (`user_id`, `message`, `email`, `event_type`) reach audit and security-event sinks and are rendered into log records. Today they are sanitized **at each call site** inside `src/kailash/nodes/auth/**` via a `log_safe` helper.

## The evidence that this does not hold

During #2066's own review round, a **sweep test** written for the finding located **two raw `user_id` sites that the author's manual enumeration had missed** — one in the MFA audit entry, one in the provider's logout record. That drift happened *within a single branch, under active review, by the person who had just written the helper*.

The author's own conclusion, which I am recording as the argument: sanitizing at N call sites will drift again, because nothing structurally prevents call site N+1 from omitting it.

## Why the sink is the right layer

`SecurityEventNode.execute` and `AuditLogNode.execute` are the choke points every caller passes through — including callers outside `nodes/auth` that no `nodes/auth`-scoped helper can reach. Sanitizing there covers every SDK caller by construction rather than by discipline.

Note the current severity split, so the fix is scoped correctly rather than uniformly:

- **`AuditLogNode`** defaults to `output_format="json"`, so `json.dumps` escapes newlines and the exposure is **latent** — it goes live only if a deployment constructs the node with a non-json format.
- **`SecurityEventNode`** renders unconditionally via f-string with no json branch, so it is the **live** surface. Every current call site was verified covered, but by discipline, not structurally.

## Expected vs actual

Expected: a caller-controlled value reaching an audit/security sink cannot inject newlines or forge log structure, regardless of which call site produced it.
Actual: safety depends on every call site remembering a helper; two were demonstrably missed in one branch.

## Severity

MEDIUM — currently no known live injection (the `SecurityEventNode` sites are all covered today; the `AuditLogNode` ones are json-escaped). This is a structural-durability fix, not an active exploit.

## Acceptance criteria

- [ ] Sanitizing happens inside `SecurityEventNode.execute` and `AuditLogNode.execute`, so a caller cannot bypass it by omission.
- [ ] The per-call-site `log_safe` calls in `nodes/auth` become redundant rather than load-bearing; decide explicitly whether to remove them (defence in depth vs. two places to drift) and record the reasoning.
- [ ] A sweep test — not an enumeration — asserts no caller-controlled value reaches a sink unsanitized. The sweep is the instrument that caught this; the enumeration is what missed it.
- [ ] Cover the non-json `AuditLogNode` path, since that is what promotes it from latent to live.

## Related

- #2066 (the branch where the drift was measured), #2060.

### LINKED ISSUE #2089: jwt_config issuer unset logs 'any token signed with this key is accepted' at INFO, below every alert threshold
Found during the security review of PR #2066. **Pre-existing from #2035**, correctly left untouched by that lane rather than scope-crept.

## The defect

`src/kailash/nodes/auth/enterprise_auth_provider.py` (around the JWT config validation, ~:841-845) detects that no `issuer` is configured and logs, in effect, that **any token signed with this key is accepted** — at `INFO` level.

## Why the level is the defect

`security.md` § "Secure-Default For A New Security Feature" requires that a protection which is OFF either fail closed, or emit a **loud one-time WARN** naming the unprotected state and its exact wiring. `INFO` is below every default alert threshold and below most production log-shipping filters, so in practice the operator is not told.

This is the same class the session repeatedly hit elsewhere: the signal exists, so the code looks defensible on inspection, but nothing observable happens at the level where someone would act.

## Expected vs actual

Expected: a deployment running JWT auth with no issuer pinned emits a WARN (or ERROR) once at init, naming the missing `issuer` and how to set it.
Actual: an INFO line that scrolls past in normal startup output.

## Severity

MEDIUM — not a bypass by itself. Issuer pinning is one of several controls (algorithm allowlist, expiry, audience, `require: [exp, sub]`, subject binding are all enforced and were independently verified working). But an unpinned issuer widens which signers are trusted, and the operator is not effectively informed.

## Acceptance criteria

- [ ] Unset `issuer` on a JWT-auth-enabled provider emits a one-time WARN (not INFO) naming the missing config and the exact wiring to set it.
- [ ] One-time per process, not per operation — a per-request message reads as transient and gets filtered (this repo already hit that in #2035 and corrected it).
- [ ] A regression test pins BOTH polarities: unset → the loud signal fires; set → no spurious signal.
- [ ] Sweep the other auth-config validations in the same file for protections reported below WARN.

## Related

- #2035 (origin), #2066 (where it surfaced), #2060.

esperie:
The auth/actor + audit-integrity cluster. Four issues that share one root: an identity taken from a request field rather than derived from an authenticated session, and records that could not say who did what.

## #2047 — `MultiFactorAuthNode` had no actor (BREAKING)

`user_id` named the SUBJECT of an action and never the caller, and `admin_override` — an ordinary caller-supplied boolean — was the only thing in front of `revoke`, `disable` and `reset`. Five successive adversarial rounds each patched the specific action the previous round exploited; fifteen dispatchable actions were keyed on a caller-typed string, five issuing a bearer credential and four destroying one. Patching the sixth was not going to end it.

**The root fix.** A caller now supplies PROOF of authentication — an opaque `actor_session_id` it cannot forge — and never the CLAIM of who it is. The node resolves that to a principal through an injected `ActorResolver` holding server-side state, and authorizes `(actor, action, subject)` ONCE, ahead of every handler. Both sides of the comparison are server-derived: the actor from the session store, the subject from the node's own records (`security.md` § Enforcement-Surface Parity → identity-derivation parity). For a workflow node that distinction is the whole fix — every input is a request field by construction, so an `actor_user_id` parameter would have been the same defect in new clothes.

The check runs once, ahead of every handler, rather than inside the ones that looked dangerous: the last five reviews each found a *different* action to be a credential-issuing or state-destroying primitive, so enumerating the dangerous ones is exactly the approach that kept failing.

- **Policy** — an actor may act on itself; acting on any other subject requires the verified `mfa:admin` capability, as do `revoke`/`disable`/`reset`, administrative recovery, and re-enrolment over a verified factor.
- **`admin_override` deprecated** — still accepted, raises `DeprecationWarning`, grants nothing under `require_actor=True`. Not kept authoritative for a deprecation cycle, because that would keep the vulnerability for a deprecation cycle. The warning text differs by mode: a deprecation notice that misdescribes live behaviour is worse than none.
- **`NullActorResolver` is the default**, so an unwired node denies everything. A resolver that raises, returns `None`, or returns a duck-typed lookalike also denies — the lookalike explicitly, since accepting one would put `has_capability` under the caller's control.
- **`require_actor=False`** is the explicit opt-out, warning once per process naming the protection that is off and its exact wiring (`security.md` § Secure-Default).
- **ONE dispatcher, ONE trusted-device store** — a gate present in one surface and absent in the other is not a gate, which is how #2026 twice found a guard reachable-around. `run` and `async_run` had different ACTION SETS (`verify_backup`/`list_methods` async-only; nine sync-only) and `trust_device` wrote to two different stores depending on entry point, so whether `revoke`/`reset` could clear a trust depended on which method the caller used. The fingerprint path also CREATED an MFA record for any subject as a side effect.
- **The audit trail can now say BY WHOM.** #2066 shipped `"actor": None` with a test asserting it, deliberately as a tripwire. That test is **updated, not deleted** — it now pins the one case where the node genuinely does not know the caller (`require_actor=False`), where a fabricated attribution would be worse than an absent one. Refused attempts are audited too.

**Framework-first.** `kailash-pact` depends on `kailash`, so a core→pact import is a package cycle — but PACT's grammar physically lives in core at `kailash.trust.pact`. Importing it from `mfa.py` would pull the whole `trust` extra (PyJWT, cryptography, pynacl) into slim-core's MFA import closure, so the node takes the house `Protocol` + `Null*` fail-closed-default idiom (`kailash.delegate.verifier`) instead. A PACT `GovernanceEngine` binds structurally — its `verify_action(role_address, action, context)` is the same shape — with zero import.

## #2088 / #2040 — sanitizing moved to the SINK

Sanitizing lived at ~30 call sites and drifted **measurably**: during #2066's own review a sweep found two raw `user_id` sites the author's manual enumeration had missed, inside a single branch, under active review, by the person who had just written the helper. It now happens in `SecurityEventNode.execute` and `AuditLogNode.execute` — the choke points every caller passes through.

- `message` is covered, not only identifier fields — `message=` is the field that actually drifted, so an identifier-scoped fix would reproduce the original defect on the exact field that failed.
- Values are **length-capped as well as escaped**; an unbounded field survives escaping intact.
- `AuditLogNode`'s non-json path is covered including NESTED strings, which is what promotes it from latent to live.
- `sanitize_log_value` / `sanitize_log_structure` are the public sanitizers #2040 asks for. `log_safe` becomes a thin alias with no implementation of its own — two copies is the drift this fix is about — and a test asserts it cannot grow one back.
- The `api_gateway.py` site #2040 names, plus **two sibling sites in `agent_ui.py` the sweep found**, now route through it.

Two defects the sweep surfaced and this fixes: `SecurityEventNode` **raised out of a security sink** on any severity that was not an exact member name (the event was never recorded); and two in-tree callers passed `"info"`/`"warning"`, so those two security events had never fired at all. An unrecognized severity now ranks TIGHTEST (CRITICAL), never silently downgrading to INFO.

## #2089 — an unpinned JWT issuer is now audible

"Any token signed with this key is accepted" was logged at INFO, once per validated token — below every alert threshold, and in a per-operation shape that reads as transient and gets filtered. Now a one-time-per-process WARN naming the missing config and exact wiring, fired at CONSTRUCTION too. **Swept the sibling validations**: every other missing-protection report also REFUSES the operation, so those are correctly per-request. Measured rather than assumed for the nearest sibling — with `audience` unset, PyJWT 2.12.1 REJECTS a token carrying `aud`, so an unset audience does not widen the way an unset issuer does. The sweep is an executable test, not prose.

## Also fixed

`APIGateway.get_optional_current_user` returned `None` unconditionally behind *"For now, return None to avoid complex auth setup"*. `enable_auth` defaults to **True**, so the authenticated principal was never resolved and the body-supplied `user_id` was always the session identity — the same bug class as #2047 at the HTTP surface, and the reason the #2040 log site is live. Now implemented, with rejected credentials logged rather than swallowed. Making an auth-enabled gateway **refuse** a credential-less session create is breaking across ~10 integration test files and exceeded this shard — filed as **#2102** with acceptance criteria and both polarities.

## F1 — the provider had FOUR call sites into its MFA node and only TWO were qualified

Found by adversarial review of the above, and the more serious finding of the two. `EnterpriseAuthProviderNode` constructs its MFA node with `require_actor=False`, and the CHANGELOG plus the code comment both justified that by asserting it "passes a SERVER-DERIVED subject" on every call. **That was false, and is measured false.** `get_methods` and `challenge_mfa` dispatched straight from `async_run` with the raw caller-supplied `user_id` — unauthenticated, and unrate-limited, since the guard reads `action == "authenticate"` only — and ungated at the node too, because of the opt-out. An unauthenticated caller sending `action="get_methods", user_id="<victim>"` learned whether that account had MFA configured: **an enumeration oracle across the whole user base**, ideal for finding un-enrolled accounts. `challenge_mfa` reached the node for an arbitrary subject the same way.

Both now derive their subject through a shared `_resolve_session_subject`, the shape `_authorize` has carried since #2026. Measured post-fix: no session → `session_required`; a forged session → `session_invalid`; a valid session → the SESSION's subject wins over the caller's claim. This is `security.md` § Multi-Site Kwarg Plumbing exactly — the qualified siblings were fixed and these two were not.

**The false sentence is corrected in place in both the CHANGELOG and the comment rather than quietly dropped.** Leaving it on the record re-seeds the premise for the next session, which is what #2031's specification-verification rule exists to stop.

Also in this round: the `X-API-Key` branch added on an earlier head is **removed, not fixed** — `verify_api_key` calls `credential_manager.execute(operation=..., credential_name=...)`, but `CredentialManagerNode.run()` ignores both and returns no `"success"` key, so it always 401s (Rule 2, verified by reading the loaded method). A credential that verifies but carries no usable subject now **refuses** rather than reverting to the body `user_id`. The route's `except Exception` would have relabelled that 401 as a 500. `channels/session.py` carried a **byte-identical** copy of the `agent_ui.py` format string, missed because it lives outside `middleware/`. `sanitize_log_value` truncates **before** flattening, so a 50 MB `user_id` no longer costs 50M `isprintable()` calls inside every audit write. The opt-out warning is keyed per NODE, not once per process — the single global latch was consumed by `EnterpriseAuthProviderNode.__init__` itself, so the SDK burned the one warning and a *user's* genuinely misconfigured node said nothing.

## CodeQL — resolved by reading the query, not by counting alerts

The check reported **11 new alerts, 7 high**. Measured after the fixes below: **4 remain, of which 3 are pre-existing alerts on `main` that moved by 17 lines.**

- **4 × `py/log-injection`** (`api_gateway.py:474,516`, `agent_ui.py:455,470`) pointed **at the `sanitize_log_value(...)` call itself** — the barrier was not recognized. Two model-file attempts were both measured NOT to work: `summaryModel` kind `value` declares the taint *propagates* (value flow is a subset of taint flow), and `neutralModel` kind `summary` has no effect on a function whose body is **in source**. `LogInjection::ReplaceLineBreaksSanitizer` matches only a `.replace` **attribute call** whose first argument is a string literal in `["\r\n", "\n"]`, so `sanitize_log_value` now ends in `.replace("\r", " ").replace("\n", " ")`. Bare `"\r"` does not satisfy it, and only the node actually **returned** is the one taint leaves through — so the `"\n"` call must be last. It is a runtime no-op (both characters are non-printable and already flattened), which is why it is pinned by an **AST test shown to RED against its deletion**, not by behaviour.
- **3 × `py/clear-text-logging-sensitive-data`** traced to `MFA_ADMIN_CAPABILITY`, reported as "sensitive data (password)". The rule classifies from the **binding's identifier**, and reads an `mfa`-containing name as credential material — right in general, wrong for a policy label whose value is the literal `mfa:admin`. The same alert classified `phone_number` as "private", which identifies the mechanism as the substring rather than the value. It reached `rest.py:676` and two kaizen security nodes because it is interpolated into refusal messages that leave via `result["error"]`, which unrelated nodes log. Renamed `ADMIN_CAPABILITY`; **re-scanned, all three cleared.** Value unchanged, with tests pinning the value and the absence of any alias.
- **The three `audit_log.py:113,115,117` highs are NOT this branch's — and an intermediate commit on this PR claimed they were.** That claim is corrected in place, in both the CHANGELOG and the `_actor.py` comment, with the evidence. It came from reading the *first* code flow in a multi-flow SARIF result as though it were the only one. **The instrument that settles "did this branch cause this alert" is alert IDENTITY across refs, not flow inspection:**

  ```
  ref=refs/pull/2103/merge   #11503 audit_log.py:113  created 2026-08-11T15:23:12Z
  ref=refs/heads/main        #11503 audit_log.py:96   created 2026-08-11T15:23:12Z
  ```

  Same alert numbers, same creation timestamp, **+17 lines apart — exactly the number of lines this branch adds above them.** Same for `#11504` and `#11505`. They are open on `main` right now; the PR check counts them only because "new alerts in code changed by this pull request" is line-scoped. They survived the rename, which is what exposed the mis-attribution.

### Remaining: 11 alerts → 4. Three are main's; one needs a disposition that is not this lane's to make

`#11530` (`mfa.py:89`, created `2026-08-12T04:56:23Z`) is the only genuinely new one. It traces the `mfa_config` **parameter** — the same name heuristic on a public API name that cannot be renamed — into the opt-out WARN's node name, through `MultiFactorAuthNode(name=..., **mfa_config)`, because a splat can statically supply any keyword. **At runtime it cannot:** `name=` is passed explicitly, and a `name` key in `mfa_config` is now refused outright (it was always a duplicate keyword argument raising an opaque `TypeError` from inside provider construction).

**Deferred, safe per #2127** (`zero-tolerance.md` Rule 1b: runtime-safety proof and tracking issue are in [this comment](https://github.com/terrene-foundation/kailash-py/pull/2103#issuecomment-5292901382); the signoff is deliberately not self-granted). It is backed by a **runtime proof, not an argument** — a regression test drives the real construction with a marked value in `mfa_config` and asserts the emitted record carries the derived node name `p2103_mfa` and neither the marked value nor its key, with the node name asserted as a **positive control** so a record carrying nothing would not pass. Dismissing it is not this lane's call and is not done here.

## Verification

**Fail-first, both polarities, with the mutation shown to reach the code before the result was read** (`instrument-discipline.md` MUST-2):

| Mutation | RED |
|---|---|
| `_resolve_and_authorize` → pre-#2047 semantics | 14 of 62, incl. all six `admin_override`-bypass tests and all three async-surface tests |
| both sinks → raw fields | 21 of 43 |
| issuer WARN → INFO | 2 of 8, incl. the sibling sweep catching it independently |

Restores verified by `shasum -c`. Rebased onto `a728984c0`; 332 tests green across every auth surface this branch touches, and `2335 passed` on the whole of `tests/regression`.

**The remaining failures were checked against main rather than assumed pre-existing** (`zero-tolerance.md` Rule 1c — "pre-existing" is unprovable without a SHA). A detached worktree at `origin/main` `a728984c0`, with none of this branch's changes present, reproduces the identical set: `test_issue_1154_slim_core_no_eager_filelock::test_file_lock_still_works_when_filelock_available`, `test_loc_invariants::test_base_agent_loc`, and all ten errors in `test_issue_workflow_parameters_envelope_parity` (`RuntimeError: Functions with **kwargs are not supported as tools`). `test_dataflow_pool_bridge` needs Redis, which is not up locally (`ConnectionRefusedError: [Errno 61]`). None is in a file this branch touches; all three belong to other lanes' surfaces and are flagged for routing rather than fixed here, since editing them would conflict with in-flight PRs.

Every new test drives the PUBLIC surface, never a private helper — testing helpers directly is how correct-but-unreachable code shipped here before. One #2026 test that called `_verify_mfa_async` directly now drives `async_run`.

## Related issues

Fixes #2047
Fixes #2088
Fixes #2089
Fixes #2040
Refs #2102, #2127, #2104, #2108, #2066, #2060, #2026, #2035


esperie:
## CI status: everything green except CodeQL — deferred per `zero-tolerance.md` Rule 1b, not dismissed

Every other required check passes on head `2c3ae9dc3`. CodeQL is red on **three `py/log-injection` alerts that point at the exact calls this PR added to FIX that defect**, plus a set of `py/clear-text-logging-sensitive-data` findings that are pre-existing.

### Fixed in this PR (CodeQL findings that were real)

- `py/unused-global-variable` ×3 — the one-time-per-process latches. Rewritten as a mutated set, so no function needs a `global` statement. Same behaviour, one fewer way for two threads to disagree about which binding they hold.
- `py/ineffectual-statement` — the Protocol method's bare `...`, replaced by the docstring that should have been there.

Both classes are gone from the current scan.

### Deferred, with proof — `py/log-injection` ×3 → **#2106**

`api_gateway.py:465`, `agent_ui.py:455`, `agent_ui.py:470`. All three are `logger.info("...%s", sanitize_log_value(user_id, 128))`.

**Runtime-safety proof.** `sanitize_log_value` replaces every non-printable character (newline, CR, NUL, ANSI escapes) with a space and caps length at a ceiling no call site can raise, so one call can only ever produce one line — exactly the property the rule asks about. Asserted behaviourally, not by inspection, in `test_a_hostile_user_id_produces_exactly_one_record_per_log_call`, which drives `AgentUIMiddleware.create_session` with newlines + NUL + ANSI and asserts on the RECORDS; reverting the call to its raw f-string REDs it.

Note `api_gateway.py` was **already open on main** at `:426` under the same rule — the line moved with the edit. So this PR reduces the underlying exposure and does not introduce it.

**Why it cannot be cleared here.** Two mechanisms tried, each measured against a re-run analysis rather than assumed:

1. `summaryModel` kind `value` (the pattern already in this repo's MaD pack) — cannot work: value flow is a **subset** of taint flow, so that row tells CodeQL the taint *propagates*.
2. `neutralModel` kind `summary` — the declaration that does mean "no flow". Also no effect, because `secure_logging.py` is **inside the analyzed source tree**: CodeQL computes flow through the function body directly, and MaD models do not override in-scope dataflow.

The supported mechanism is a custom query extending `LogInjection::Configuration` with an `isBarrier` predicate. That is a CodeQL-pack change with its own blast radius — it must be pinned so it does not blanket-suppress the rule elsewhere — and it belongs in its own shard. **#2106** carries the acceptance criteria, including the requirement that a fixture still ALERTS on a raw interpolated value in the same file so the barrier is proved narrow.

The same finding also establishes that the two pre-existing `dataflow.utils.masking` rows in the pack have never worked, which is consistent with the 2026-04 findings they were added for still being open on main. That reasoning is now written into `sanitizers.model.yml` so it is not re-derived a third time.

### Pre-existing, out of scope — `py/clear-text-logging-sensitive-data`

`audit_log.py:113/115/117` are the same alerts open on main at `:96/98/100`, line-shifted by this PR's edit (verified against `refs/heads/main`). `rest.py:676` and the two kaizen sites are untouched by this branch. Rule 1a says "same on main" is not a disposition — but these are a different bug class (secrets *reaching* a sink) from this PR's (untrusted *structure* in what is logged), they need a redaction decision at the sink with real blast radius, and folding them in would exceed this shard. Flagged here rather than silently left.

### Rule 1b condition 4

Requires release-specialist signoff or user override for the deferral. **Not obtained** — flagging for the reviewer rather than claiming it.

esperie:
## Round 2 pushed — and the three unexplained CodeQL highs are SETTLED

### The CodeQL question: my change did not open a new flow. Confirmed, not hypothesised.

You asked me to establish which branch it was rather than assume the benign one. Two independent lines of evidence, both discriminating:

**1. Two of the three predate this branch by five months.** Alert creation timestamps on the PR ref:

```
2026-03-18T01:29:56Z  packages/kailash-kaizen/.../ai_threat_detection.py:262
2026-03-18T01:29:56Z  packages/kailash-kaizen/.../ai_behavior_analysis.py:254
2026-08-12T04:12:44Z  src/kailash/nodes/api/rest.py:676
```

Alerts created 2026-03-18 cannot have been caused by a branch that did not exist. They are long-standing findings re-surfaced in this run's result set, not regressions.

**2. `rest.py:676` has no reachable path to anything I changed.** It is `self.logger.error(f"REST API error: {error_message}")` inside `_handle_error`, with `error_message` built locally in the same function. Measured:

```
grep -c 'SecurityEventNode|AuditLogNode|secure_logging|sanitize_log|log_hygiene|nodes.auth'  src/kailash/nodes/api/rest.py   → 0
git diff --name-only origin/main...HEAD | grep -c 'nodes/api/rest.py'                                                        → 0
```

Zero references, and the file is untouched by the diff. There is no edge from my sinks to that log call.

So all three are the sanitizer-not-modelled class one hop out, exactly as you suspected — but now established rather than assumed. **#2106** carries the deferral package.

On the mechanism itself, worth recording because it changes the repo's standing assumption: I tried the MaD pack twice and **both attempts failed, measured against re-run analyses**. `summaryModel` kind `value` *cannot* work — value flow is a **subset** of taint flow, so that row tells CodeQL the taint propagates. `neutralModel` also had no effect, because `secure_logging.py` is **inside the analyzed tree**, where MaD does not override in-scope dataflow. **This means the two pre-existing `dataflow.utils.masking` rows have never worked either**, which is consistent with the 2026-04 findings they were added for still being open on main. That reasoning is now written into `sanitizers.model.yml`. The supported route is a custom query with an `isBarrier` predicate — in #2106, with a fixture requirement so the barrier is proved narrow rather than a blanket suppression.

The four notes you listed are fixed and gone from the current scan.

### F1 — you were right, and the premise was false

Verified in my own worktree and closed. `get_methods` and `challenge_mfa` dispatched from `async_run` with the raw caller `user_id`; both now derive the subject through a shared `_resolve_session_subject`, the shape `_authorize` has carried since #2026. Measured post-fix:

```
no session      → {'success': False, 'reason': 'session_required'}
forged session  → {'success': False, 'reason': 'session_invalid'}
valid session   → user_id == 'alice'   (the SESSION's subject, not the caller's 'victim' claim)
```

**Both the CHANGELOG bullet and the code comment are corrected in place and say they were wrong**, rather than being quietly dropped — leaving that sentence on the record is what re-seeds a false premise next session.

### F2 — (a) removed, not fixed; (b) fixed, and it exposed two more

**(a)** Confirmed by reading the loaded method: `CredentialManagerNode.run()` takes `**inputs`, never reads them (it uses `self.credential_name`), and returns no `"success"` key on any path. So `verify_api_key` always 401s. I **removed** my `X-API-Key` branch rather than shipping an auth path that can only reject, and filed the underlying defect as **#2108** with both-polarity ACs.

**(b)** An unusable principal now 401s instead of reverting to the body value. Fixing it surfaced two more in the same handler: the route's `except Exception` would have relabelled that 401 as a **500**, and its failure log interpolated the **raw body `user_id`** — nine lines below a site my first sweep had just fixed. That is #2088's drift argument happening inside this very change; both fixed.

### F4, F5, F6, F8, F9 — all fixed

- **F6:** `channels/session.py` (byte-identical sibling) and the `mfa.py` SMS-failure site now sanitized. **CHANGELOG claim narrowed** from "every SDK caller" to "every caller of the two security/audit NODES" — direct `logger` calls remain a call-site obligation, which is the honest scope. `jwt_auth.py` left to its owning lane, tracked in #2104.
- **F5:** the latch is keyed per NODE and names it. Your diagnosis was exact — the SDK's own `EnterpriseAuthProviderNode` opt-out burned the single process-wide warning, so a user's misconfigured node said nothing.
- **F4:** truncate-then-flatten (the map is 1:1, so byte-identical output, no 50M-call DoS lever). **F8:** `OverflowError` caught. **F9:** `_SELF_SERVICE_ACTIONS` is now READ as an allowlist, which also makes an unclassified action fail **closed**; a sweep test asserts every dispatchable action is classified.

**F3, F7, F10** judged and left: F3's main driver was the API-key branch, now removed; F7/F10 are LOW and I would rather not expand an already-large shard further — say the word if you want them in.

70 tests in the #2047 suite, 559 green across all affected surfaces.

`Refs #2102` — yes, mine: filed for scope I deliberately held (making an auth-enabled gateway **refuse** a credential-less session create, breaking across ~10 integration test files). **#2104**, **#2106**, **#2108** are also mine, all split out rather than smuggled in.

esperie:
## CodeQL disposition — `zero-tolerance.md` Rule 1b artifacts 1–3

**Measured, on head `c01d53fc9`: 11 alerts → 4.** Rule 1b's fourth artifact — the signoff — is deliberately not self-granted here.

| | before | after | how |
|---|---|---|---|
| `py/log-injection` | 4 | **0** | `.replace("\r", " ").replace("\n", " ")` as the returned expression of `sanitize_log_value` |
| `py/clear-text-logging` (`ADMIN_CAPABILITY` source) | 3 | **0** | constant renamed; value unchanged |
| `py/clear-text-logging` (`audit_log.py`) | 3 | 3 | **pre-existing on `main`** — not this branch's |
| `py/clear-text-logging` (`mfa_config` source) | 1 | 1 | false positive, proof below |

### Artifact 1 — runtime-safety proof for `#11530` (`mfa.py:89`)

`node_name` cannot carry `mfa_config` content at runtime, for two independent reasons:

1. `name=` is passed **explicitly** at `src/kailash/nodes/auth/enterprise_auth_provider.py:239` and always wins.
2. A `name` key in `mfa_config` is **refused outright** at `src/kailash/nodes/auth/enterprise_auth_provider.py:233-238`.

CodeQL cannot see either, because `MultiFactorAuthNode(name=..., **mfa_config)` lets a splat statically supply any keyword. Measured by driving the real construction with a marked value in `mfa_config` (`TestTheOptOutWarnCannotCarryMfaConfigContent`):

```
WARN records: 1
  contains node name 'p_mfa':                        True     <-- positive control
  contains mfa_config VALUE 'SEKRIT-hunter2-VALUE':  False
  contains mfa_config KEY 'totp_issuer':             False
```

### Artifact 2 — tracking issue

#2127, with three alternative acceptance criteria for a full fix.

### Artifact 3 — this link, also in the PR body

Deferred, safe per #2127.

### Artifact 4 — signoff

**Not self-granted.** Deciding whether to dismiss `#11530` as a false positive against the proof above is the merging reviewer's call.

### The three `audit_log.py` alerts are not a deferral at all

They are `#11503`/`#11504`/`#11505`, which carry **those same alert numbers on `refs/heads/main`** at lines `96`/`98`/`100`, with `created_at 2026-08-11T15:23:12Z` — the same three alerts, shifted by the +17 lines this branch adds above them.

```
ref=refs/pull/2103/merge   #11503 audit_log.py:113  created=2026-08-11T15:23:12Z
ref=refs/heads/main        #11503 audit_log.py:96   created=2026-08-11T15:23:12Z
```

An intermediate commit on this PR claimed they *were* this branch's, from reading the first code flow in a multi-flow SARIF result as though it were the only one. That claim is **corrected in place** in the CHANGELOG and in the `_actor.py` comment, with the evidence, rather than dropped. Alert identity across refs — not flow inspection — is the instrument that settles it.

### Every other job is green on this head

All 16 check runs enumerated by name; each has a `success` present. The one `build` / `cancelled` is concurrency supersession — a later `build` / `success` exists on the same head. `required_status_checks.contexts` is `[]` and `enforce_admins` is `false`, so nothing here blocks an admin merge.

esperie:
## Rule 1b artifact 4 — co-owner OVERRIDE granted, and what it does and does not cover

`zero-tolerance.md` Rule 1b requires four artifacts for a scanner deferral: a written runtime-safety proof, a tracking issue, a PR-body link, and **release-specialist signoff (or user override)**. Artifacts 1–3 were delivered by the lane. **Artifact 4 is granted here as an explicit co-owner override**, not a self-grant.

### Scope of the override: alert #11530 ONLY

`py/clear-text-logging-sensitive-data` at `src/kailash/nodes/auth/mfa.py:89`.

**Verified independently by the orchestrator before the override was sought:**

```python
# src/kailash/nodes/auth/enterprise_auth_provider.py:233-238
if "name" in mfa_config:
    raise ValueError(
        "mfa_config may not carry 'name': the MFA node's name is "
        f"derived from this provider's own name as '{self.name}_mfa'. "
        "Remove the 'name' key."
    )
self.mfa_node = MultiFactorAuthNode(name=f"{self.name}_mfa", **mfa_config)
```

The flagged line logs `node_name` in a WARN about actor enforcement being disabled. CodeQL reaches it because a `**` splat can **statically** supply any keyword and `mfa_config` trips the sensitive-name heuristic. At runtime that is unreachable twice over: `name=` is passed explicitly, and a `name` key **raises before construction**. The lane's runtime proof agrees — the WARN contains the node name, and neither the `mfa_config` value nor its keys.

**#2127** tracks clearing it properly, with three alternative acceptance criteria.

### NOT covered by this override

**Alerts #11503 / #11504 / #11505** (`audit_log.py:113/115/117`) are **pre-existing, not deferred here.** Established by alert identity across refs — the discriminating instrument, since a byte-comparison cannot distinguish a shifted alert from a new one:

```
refs/pull/2103/merge   #11503 audit_log.py:113  created=2026-08-11T15:23:12Z
refs/heads/main        #11503 audit_log.py:96   created=2026-08-11T15:23:12Z
```

Same alert numbers, same creation timestamps, +17 lines — exactly what this branch adds above them. They are on `main` today and this PR neither introduces nor resolves them. The sanitization at that sink is genuinely correct (every component of `log_message` passes through `sanitize_log_value` / `sanitize_log_structure` before assembly); CodeQL does not model those as barriers, which is the #2044 class.

### Stated plainly

This merge accepts a **red CodeQL check** on a security-touching PR. Every test job is green — `Test (Python 3.11–3.14)`, PACT, DataFlow Unit, DataFlow Infra, `test-with-infrastructure`, `Analyze Python` — each confirmed by a `success` **present** on the named job, not by absence of failure. No independent security lens re-derived the #11530 analysis; the override rests on the lane's runtime proof plus the orchestrator's verification of the guard quoted above.

PR #2123, opened 2026-08-13
chore(sync): Gate-2 build py from loom dc07a7c6cc4d

esperie:
Automated Gate-2 distribution from loom dc07a7c6cc4d.

Applied in an isolated worktree from origin/main (f0dfac30e903); the target's local checkout was NOT touched. Manifest: +123 ~409 -3.


esperie:
Rebased onto `main` (`c23836a17`) — 0 behind, 1 ahead. Verified rather than assumed:

**Rebase was clean by construction.** `main` changed exactly ONE file under the synced artifact paths in the 177 commits since the merge-base (`.claude/rules/ci-runners.operator.local.example.md`, from `42ae3607e`), and this branch does not touch it — zero overlap, no conflicts. Confirmed post-rebase: `git diff origin/main HEAD -- .claude/rules/ci-runners.operator.local.example.md` is empty, and the path-scoping fix is intact in the file.

**Diff vs main:** 616 files, +35121/−3065 — the distribution only. No `src/` or `tests/`.

**Checks run, with their discrimination stated:**

- `node --check` over all 196 synced `.js`/`.mjs` hook and bin scripts → 0 failures. Positive control: `node --check` on a deliberately malformed file correctly rejects it, so the zero is a measurement rather than an empty glob.
- `coc-manifest-integrity` → PASS (manifest ↔ probes ↔ on-disk artifacts consistent).
- `validate-xref-integrity` → PASS.
- `validate-emit` → 754 pass / 4 fail. **The 4 fails are not a regression:** `main` produces the byte-identical fail set (627 pass / 4 fail), all four rooted in `.claude/sync-manifest.yaml` being absent. That file is loom-only; this repo is `coc-build`, so `validate-emit` cannot complete here on either side. Sync moves 627 → 754 passing with the same 4 blocked.
- `pre-commit run --from-ref origin/main --to-ref HEAD` → all applicable hooks pass (Python hooks skip; the diff has no Python).

**settings.json wiring reviewed against the ships-un-enrolled constraint.** It adds two PreToolUse/Stop hooks — `worktree-forest-guard.js` and `session-notes-guard.js`. Neither reads roster, `verified_id`, enrollment, or posture state (grepped), so neither can brick a clone of this public un-enrolled repo. Smoke-tested in the rebased tree: both return `{"continue":true}`, exit 0. `session-notes-guard` has exactly one blocking path — a *truncated read* of `.session-notes` — which is the narrow structural class `hook-output-discipline.md` MUST-2 reserves `block` for; `worktree-forest-guard` has no blocking path at all.

Issue #2078, opened 2026-08-11
Tier-2 suite: 232 failures on CI from thread/fd exhaustion — blocks making Tier 2 a merge gate (#2038)


esperie:
Blocks the last acceptance criterion of #2038 (making Tier 2 gate a merge).
Filed from PR #2077, which fixed 45 consistently-failing Tier-2 tests and then
removed `--maxfail=20` so the suite could complete on CI **for the first time**.

## What was previously unknowable

`unified-ci.yml` ran Tier 2 with `continue-on-error: true`, `--maxfail=20` and a
10-minute step budget. Together those three made the true state unobservable:

- On **Python 3.11 and 3.12 the step TIMED OUT at 10 minutes on every run
  inspected** — runs `31480330313`, `31478307416`, `31476919718` — and every one
  of those jobs still reported **SUCCESS**. `continue-on-error` swallowed the
  timeout as well as the failures.
- On **3.14** the run stopped at `--maxfail=20` (14 failed, 6 errors, 152s).
- `--maxfail=20` guarantees a truncated window rather than a failure set. PR
  #2077 fixed 11 of the 20 tests the pre-change 3.14 run reported; the run then
  simply reached the next batch and stopped again at 20.

## The actual number

With `--maxfail` raised, Python 3.12, run `31482911385`:

```
174 failed, 2349 passed, 47 skipped, 4 deselected, 1 xpassed,
160 warnings, 58 errors in 175.76s (0:02:55)
```

**232 failures out of ~2,629.** The suite completes in under three minutes once
it is allowed to; the 10-minute timeout was not a slow suite, it was something
else (see below).

## It is one root cause, not 232 defects

Two signatures account for essentially all of it:

```
$ grep -cE "can't start new thread" <job log>          130
$ grep -cE "sqlite3.OperationalError: disk I/O error"  90
```

plus 21 `async_local.py:1136 Workflow execution failed: can't start new thread`,
10 `TaskException: Failed to initialize task manager: disk I/O error`, and 2
`RuntimeError: Cannot run the event loop while another loop is running`.

`can't start new thread` and sqlite `disk I/O error` are the classic paired
symptoms of **process resource exhaustion** — thread limit and file-descriptor
limit. The affected files are spread across the tier (test_async_sqlite_pool.py
17, test_todo029_api_gateway.py 13, runtime/test_sync_nodes_in_async_runtime.py
[...]

Issue #73, opened 2026-03-24
fix: resolve 3001 pyright type errors across src/kailash/


esperie:
## Problem

`pyright src/kailash/ --level error` reports **3001 errors**. This has been hidden because the CI step used `--level basic` which is not a valid pyright flag — pyright exited with code 4 (invalid argument) on every run, and the step was always "failing" but the failure was from the invalid flag, not from type errors.

## Impact

- Type safety is effectively unenforced across the entire SDK
- Real type regressions are invisible
- Contributes to CI noise and false sense of security

## Plan

1. Set pyright to `continue-on-error: true` with `--level error` (done in PR #72)
2. Incrementally fix errors by module, prioritizing:
   - `src/kailash/runtime/` (core execution path)
   - `src/kailash/trust/` (security-critical)
   - `src/kailash/nodes/` (user-facing API)
3. Once errors reach zero, remove `continue-on-error: true`

## Current state

\`\`\`
3001 errors, 0 warnings, 0 informations
\`\`\`

Related: PR #72 (fix(runtime): lifecycle management)

PR #2124, opened 2026-08-14
ci(unified): Tier 2 gates the merge — #2038 complete

### LINKED ISSUE #2038: CI: Tier-2 step is `continue-on-error: true` — a green `gh pr checks` is not evidence about integration tests
Structural finding surfaced while gating a security PR. This affects **every** PR, not one.

## The defect

`.github/workflows/unified-ci.yml` sets `continue-on-error: true` on the Tier-2 step:

```yaml
      - name: Test (Tier 2 — integration runtime tests)
        timeout-minutes: 10
        continue-on-error: true # Tier 2 has thread/I/O tests that may fail on CI runners
        run: |
          .venv/bin/python -m pytest tests/tier2_integration/ ...
```

So a Tier-2 failure does not fail the job, and `gh pr checks <N>` reports all-green regardless.

## Why this matters — it is not hypothetical

While reviewing PR #2034, a reviewer found two Tier-2 tests failing at the PR head:

```
tests/tier2_integration/channels/test_api_channel_comprehensive.py:284: test_handle_request_with_exception
tests/tier2_integration/channels/test_api_channel_comprehensive.py:448: test_emit_event_endpoint_error
```

`gh pr checks 2034` was **all-green at the same time**. Those two tests had pinned the *old leaking* behaviour of an HTTP handler — i.e. a security vulnerability written down as expected behaviour — and CI structurally could not surface the conflict. (Those particular failures are since resolved, but the blindness that hid them is not.)

## The instrument problem

The `continue-on-error` comment gives the honest reason: Tier-2 has thread/IO tests that flake on CI runners. That is a real problem. But the chosen remedy makes the check **non-discriminating**: it reports the same green whether Tier 2 passed or failed entirely. A check that cannot come out the other way carries no information, so any "CI is green" claim about integration behaviour is currently unfalsifiable.

## Acceptance criteria

- [ ] Tier-2 results are visible in the PR check surface as a distinct, non-green state when tests fail — a separate required-or-not check, an annotation, or a posted summary; not a silently-swallowed step.
- [ ] Genuinely runner-flaky tests are identified and quarantined **individually** (marker, `xfail(strict=True)`, or a documented skip with a tracking issue) rather than the whole tier being made unfailable.
- [ ] `xfail(strict=True)` is preferred over skip for anything expected to pass later, so it self-clears loudly when it starts passing (`testing.md` § deferred-implementation conformance vectors).
- [ ] Contributor docs state plainly which tiers gate a merge and which do not, so "CI is green" has a defined meaning.

## Related issues

Surfaced during the burn-down of #2015 (PR #2034). Sibling of #2002 (root `tests/regression/` has no blanket CI coverage — 1,564 of 1,566 tests never run): both are cases where the *absence* of coverage is invisible from the check surface.

esperie:
Removes `continue-on-error: true` from the Tier-2 step and deletes the paired annotation step — which is exactly what the block above the flag instructed:

> **TO COMPLETE #2038:** fix the leak (#2078), confirm a complete green run on all four interpreters, then delete this flag AND the annotation step below. Do not delete the flag without doing that first.

**Both preconditions are now met.**

## Why the flag existed twice

It came off **once before** (`d1f5ea3fb`) after 45 individually-diagnosed Tier-2 failures were fixed, and had to be **put back** (`1f6a95ba8`) when a second, unrelated cause surfaced: 174 failed / 58 errors from thread and file-descriptor exhaustion, reproducing against **unmodified** code and therefore belonging to no PR that saw it.

That second cause is **#2078**, fixed in **PR #2097** (`590983dc9`): `PythonCodeNode` set an address-space rlimit (`RLIMIT_AS`) **process-wide** and never restored it, so once any node executed, every later thread stack and file descriptor in the same pytest process drew on a capped address space. One cause, both symptom families — which is why `can't start new thread` and `sqlite3 disk I/O error` always appeared together.

## Evidence — read from step summaries, not job conclusions

A green *job* proved nothing while this flag was set, so these are the per-interpreter step summaries:

```
Python 3.11  Tier 2: 2582 passed, 47 skipped, 4 deselected, 1 xpassed
Python 3.12  Tier 2: 2582 passed, 47 skipped, 4 deselected, 1 xpassed
Python 3.13  Tier 2: 2582 passed, 47 skipped, 4 deselected, 1 xpassed
Python 3.14  Tier 2: 2582 passed, 47 skipped, 4 deselected, 1 xpassed
```

Zero failed, zero errors, all four. Compare the pre-fix figure recorded on #2078: **174 failed + 58 errors**.

## Why the annotation step goes with the flag

It existed only to make a *non-blocking* failure visible in the check surface — `::error::` plus a job-summary line saying "a green PR check does NOT mean Tier 2 passed." A blocking step needs no annotation to be seen, and leaving it would emit a warning about a holding state that no longer exists.

## The masking was not theoretical

**PR #2105 introduced a real test failure that sat invisible in this exact tier.** A vacuous test was fixed in `tests/unit/` while its byte-identical sibling in `tests/tier2_integration/` went red — and CI stayed green, because of this flag. It was found by an adversarial review, not by the check surface.

## Scope

Deliberately kept out of #2097 so the *fix* and the *merge-policy change* could be judged separately. This PR's own CI is the test of the change: Tier 2 now reds a PR the way Tier 1 does.

The unrelated `continue-on-error` on the pyright step (#73, 3001 pre-existing errors) is untouched.

## Related issues

Fixes #2038


esperie:
## This PR is RED, and that is the finding

Tier 2 fails on **all four interpreters** now that it blocks. Before assuming this PR introduced them:

```
$ gh pr view 2124 --json files
  25+/39-  .github/workflows/unified-ci.yml     # the only file
```

The diff removes comments, `continue-on-error: true`, and the paired annotation step. No source, no tests, no pytest invocation. **It cannot introduce a test failure — it can only stop hiding one.** The 23 failures are on `main` right now.

```
13 failed, 2559 passed, 47 skipped, 4 deselected, 1 xfailed, 10 errors in 55.37s
```

Cause is `kailash.utils.server_auth.ServerAuthNotConfiguredError` — PR #2100's fail-closed default (correct; it fixes CRITICAL #2072) broke Tier-2 suites that construct servers without credentials. #2100 updated the suites it could **see**; Tier 2 was masked, so its author got no signal.

Tracked as **#2129**, which blocks this PR. Holding here — merging would red every PR in the repo on failures none of them introduced, which is the exact reason the flag was added in the first place.

The right order is #2129 → then this.

Issue #2138, opened 2026-08-15
RotatingCredentialNode drives CredentialManagerNode through four operations that do not exist — rotation never fires


esperie:
Found by the #2108 AC-4 sweep ("Sweep the other `credential_manager.execute(...)` call sites for the same discarded-kwarg pattern and report each, fixed or not"). Reported rather than fixed there: the fix is a distinct feature, not a correction, and it exceeds one shard.

## The defect

`RotatingCredentialNode` drives `CredentialManagerNode` through a contract that node has never had. Ten call sites in `src/kailash/nodes/security/rotating_credentials.py` pass one of four operations:

```
operation="get_credential"       (lines 259, 318, 367)
operation="validate_credential"  (line 394)
operation="store_credential"     (lines 426, 449, 458, 486, 495)
operation="delete_credential"    (line 480)
```

None exists. Measured on the loaded node rather than inferred — `CredentialManagerNode.get_parameters()` returned `{}`, so `Node.execute` strips every one of those keywords before `run()` is reached:

```
[NODE] Unknown parameter(s) for CredentialManagerNode:
['credential_name', ..., 'operation']. Valid parameters: [].
```

and the node's return dict is `{credentials, source, validated, masked_display, metadata}` — there is no `success` key on any path, nor `credential`, nor `error`. So every one of these is unconditionally false:

```
if not credential_result.get("success"):        # line 262
if refresh_result.get("success"):               # line 324
if not store_result.get("success"):             # lines 431, 491
if not final_store_result.get("success"):       # line 454
```

`_check_credential_expiration` therefore always returns `{"needs_rotation": False, "error": "Failed to retrieve credential"}`, and rotation never fires. The node is non-functional end to end (`zero-tolerance.md` Rule 2), not merely mis-wired.

#2108 fixed the sibling half of this (`CredentialManagerNode`'s kwargs are now declared and honoured, and `MiddlewareAuthManager` got a real `APIKeyStore`), which is why the remaining call sites are worth stating precisely rather than leaving implied.

## Why this was not fixed in #2108

`CredentialManagerNode` **fetches** credentials from an environment variable, a file, or a vault. It has no write path, deliberately. Making these ten call sites work does not mean honouring their kwargs — it means building the writable, rotating credential backend they assume:

- store / delete, with a temporary-credential staging step (line 426) and rollback to the previous value on failure (lines 458, 495)
- atomic swap so a reader never observes a half-rotated credential
[...]

PR #2140, opened 2026-08-15
fix(auth): make API-key authentication actually work (#2108) and stop a non-ASCII key flooding logs (#2114)

### LINKED ISSUE #2108: MiddlewareAuthManager.verify_api_key can never succeed — CredentialManagerNode ignores its arguments and returns no success key
Surfaced during the #2103 security review (finding F2a). Split out rather than fixed there: the defect is in `MiddlewareAuthManager` / `CredentialManagerNode`, not in the #2047 actor work, and fixing it means settling `CredentialManagerNode`'s parameter contract.

## The defect

`src/kailash/middleware/auth/auth_manager.py` — `verify_api_key`:

```python
result = self.credential_manager.execute(
    operation="get_credential", credential_name=api_key
)
if not result.get("success", False):
    raise HTTPException(status_code=401, detail="Invalid API key")
```

`CredentialManagerNode.run()` — measured by reading the loaded method, not inferred:

- signature is `def run(self, **inputs)` and the body **never reads `inputs`** — it uses `self.credential_name`, fixed at construction, so **both `operation=` and `credential_name=` are silently discarded**;
- the return dict is `{"credentials", "source", "validated", "masked_display", "metadata"}` — **there is no `"success"` key on any path**.

So `result.get("success", False)` is `False` unconditionally and **`verify_api_key` always raises 401**. API-key authentication cannot succeed for any key, valid or not.

This is two distinct rule violations at once: an auth path that can only ever reject is a non-functional feature presented as a working one (`zero-tolerance.md` Rule 2), and documented kwargs accepted but with zero effect on the body are the silent-fallback mode at the API surface (Rule 3c).

## Why it went unnoticed

`enable_api_keys` gates it, and the only in-tree caller reachable without that flag is the FastAPI dependency in `get_current_user_dependency`, whose bearer-token branch is tried first and satisfies every existing test. #2103 added a second caller on the session route and **removed it again** once this was measured, rather than shipping a branch that can only log a rejection.

## Acceptance criteria

- [ ] Decide `CredentialManagerNode`'s contract: either `run()` honours `operation` / `credential_name` from `**inputs`, or `verify_api_key` stops passing them and uses a correctly-constructed node per lookup. A node that accepts kwargs and discards them is Rule 3c either way.
- [ ] `verify_api_key` checks a key the node actually returns (`validated`, or an explicit `success` the node starts emitting) — not one that never exists.
- [ ] Behavioural test, both polarities: an ISSUED key authenticates and yields its metadata; an unissued/revoked key 401s. The current suite cannot distinguish the two, which is why this survived.
- [ ] Sweep the other `credential_manager.execute(...)` call sites for the same discarded-kwarg pattern and report each, fixed or not.
- [ ] Re-enable an API-key path on the `APIGateway` session route (#2103 removed the caller, not the capability) once the above holds.

## Related

- #2103 (where it surfaced), #2102 (the session route's remaining fail-open default)

### LINKED ISSUE #2114: Non-ASCII X-API-Key drives an unauthenticated traceback per request via compare_digest TypeError
Deferred from #2100. Fails closed (401), so this is log-flooding, not an auth bypass.

## What

`secrets.compare_digest` raises `TypeError` when either `str` argument contains a character outside ASCII:

```
TypeError: comparing strings with non-ASCII characters is not supported
```

Starlette decodes request headers as **latin-1**, so any header byte `>= 0x80` reaches Python as a non-ASCII `str`. An `X-API-Key` containing one therefore reaches `_make_api_key_validator._validate` in `src/kailash/utils/server_auth.py` and raises.

The raise is caught — `JWTAuthMiddleware._dispatch_api_key` wraps the validator call and returns `_unauthorized("invalid_api_key")` — so **the request is correctly refused**. But it is caught by `logger.exception`, which writes a **full traceback per request**.

## Why it matters

That path is:

- **unauthenticated** — no credential needed to reach it
- **attacker-drivable** — one header byte, repeatable at request rate
- **unbounded** — a traceback per attempt

This is the log-volume amplification `observability.md` forbids on an error path an attacker controls. An `except: pass` would be worse, but a stack trace per anonymous request is not the right disposition either: nothing here is a server fault, so there is no diagnostic worth a traceback.

## Acceptance criteria

1. Guard before comparison — reject a non-ASCII key without entering `compare_digest` (e.g. `api_key.isascii()`).
2. Rejection stays a 401 with the same body; no behaviour change for the caller.
3. The rejection logs at most a bounded, non-traceback record (a counter or a WARN without `exc_info`).
4. A regression test driving a non-ASCII `X-API-Key` that asserts 401 **and** that no traceback was emitted, paired with a valid-key 200 control.

Note the guard must **not** short-circuit the constant-time property for the ASCII case: the existing loop deliberately compares against every configured key rather than returning early.

Refs #2072, #2100

esperie:
## Summary

Two API-key authentication defects, both measured rather than inferred.

**#2108 — API-key authentication could never succeed.** `MiddlewareAuthManager` issued and verified keys through `CredentialManagerNode`, which cannot store anything:

```
create_api_key -> NodeExecutionError: Node 'CredentialManagerNode' execution failed:
                  ValueError: Credential 'api_credentials' not found in any configured source
verify_api_key -> HTTPException 401 Invalid API key
```

Two independent causes. The node's `get_parameters()` returned `{}`, and `Node.execute` drops every keyword it does not find there, so `operation=` and `credential_name=` never reached `run()`:

```
[NODE] Unknown parameter(s) for CredentialManagerNode:
['credential_name', ..., 'operation']. Valid parameters: [].
```

And the node's return dict is `{credentials, source, validated, masked_display, metadata}` — no `success` key on any path — so `result.get("success", False)` was unconditionally False.

**#2114 — a non-ASCII `X-API-Key` drove a traceback per anonymous request.** The 401 was always correct; the log volume was the defect. Measured pre-fix, 5 requests produced 5 traceback records, and the websocket handshake produced its own:

```
AssertionError: a non-ASCII X-API-Key wrote a traceback:
  [('kailash.trust.auth.asgi', 'jwt_auth_middleware.api_key_validator_failed')]
TypeError: comparing strings with non-ASCII characters is not supported
```

## What changed

**#2108.** `CredentialManagerNode` **fetches** credentials from an environment variable, a file, or a vault; it has no write path, and giving it one would be a different node wearing this one's name. So the API-key methods stop pretending it is a key store and get one: `APIKeyStore` in `kailash.middleware.auth.api_keys`, shaped after the `TokenRevocationStore` already beside it — abstract contract, process-local `InMemoryAPIKeyStore` default, `api_key_store=` injection point for a shared backend.

- Keys stored as SHA-256 digests; plaintext returned once, at creation.
- Unknown, revoked, and expired keys answered identically — distinguishing them would tell an anonymous prober which guess had once been real.
- A store outage refuses rather than admits; a store that rejects the write does not hand back a key.
- New `revoke_api_key()`; `create_api_key()` accepts `expires_at=` / `metadata=`.
- The default store is process-local like the revocation store's, but fails **closed** (a key from another worker is refused), so it is documented rather than warned about.

Separately, `CredentialManagerNode`'s own kwargs are now real (Rule 3c): declared in `get_parameters()`, honoured in `run()`, with the cache keyed on the **effective** name. `execute()` with no arguments is unchanged.

**#2114.** Both halves closed. The presented key is rejected before `compare_digest`; a non-ASCII **configured** key is refused at construction (previously every comparison raised, including against the operator's own correct key — a silently dead deployment). The constant-time property is preserved: no configured key can be non-ASCII, so a non-ASCII presented key cannot match any of them and early rejection reveals nothing about which key it was compared against; the ASCII path still runs the full loop.

Because `JWTConfig.api_key_validator` is public configuration, any caller-supplied validator could raise the same way, so the middleware bounds it on **both** the HTTP and websocket surfaces: the first failure of each `(event, exception type)` keeps its full traceback, repeats are recorded at WARNING with a running count and no `exc_info`. Not silenced — the opposite defect.

## Enforcement-surface parity — the sweep (#2108 AC 4)

Every `credential_manager.execute(...)` call site, reported per the AC:

| Site | Disposition |
|---|---|
| `auth_manager.py` `create_api_key` / `verify_api_key` | **Fixed** — real `APIKeyStore`. |
| `access_control.py:388` `MiddlewareAuthenticationMiddleware` | **Fixed** — same defect at a third site (`action="validate_token"`, checking a `"valid"` key returned on no path), so it refused **every** request behind the comment "For now, simulating with credential manager". Now delegates to `JWTValidator`, accepts `token_verifier=`, fails closed with a one-time warning naming the wiring. |
| `rotating_credentials.py` (10 sites) | **Reported, not fixed → #2138.** Drives four operations that do not exist; closing it needs a writable backend with atomic swap and rollback — a feature, not a correction, exceeding one shard per `autonomous-execution.md` § Per-Session Capacity Budget. The node's docstring now says rotation does not fire. |

## Found on the way, fixed here

- **`X-API-Key` authenticates on `APIGateway` again — and the reason it is not dead code.** The obvious objection is that the request gate already handles API keys server-wide: `JWTAuthMiddleware._dispatch_api_key` verifies the key and leaves an `AuthenticatedUser` on `scope["state"]["user"]` (`asgi.py:305`), which `_authenticated_user_id` reads first. That objection is right about GATED deployments and wrong about this issue, because **the gate and the manager answer for different key sets**:

  | | Keys it accepts | Source |
  |---|---|---|
  | The gate's `api_key_validator` | keys the **operator configured** | `KAILASH_API_KEY_*` env vars — `server_auth.py:402`, `api_key_enabled=bool(api_keys)` |
  | `MiddlewareAuthManager.verify_api_key` | keys the manager **issued** | `create_api_key()` → `APIKeyStore` |

  A key issued through `create_api_key` is not among the operator's env-var keys, so a gated deployment REJECTS it before the route is ever reached — which is exactly the "#2108 path that cannot authenticate anyone" this PR closes. The arm therefore lives in `APIGateway._authenticated_user_id`, the single place #2102 made this gateway answer "who is calling", and is reachable where no gate is installed for the request (`require_auth=False`, `external_auth_reason=`, an operator-exempted path). It is **not** a `POST /api/sessions` arm: a per-route branch would be a second claim-precedence implementation, which is what `subject_from_claims` exists to prevent.

  It **never raises**. `verify_api_key` signals rejection with a 401 `HTTPException`; letting that escape the identity layer would bypass `_resolve_identity`, which owns the `require_auth` fail-closed decision for every route at once and which `enable_auth=False` callers rely on to keep serving openly. A rejected key resolves to "no principal", the same as an absent one. A manager with no `verify_api_key` at all — `JWTAuthManager`, the default — is reported **once per gateway**, not once per anonymous request carrying the header.

- **`EnterpriseSecurityEventNode` was unconstructable**, which made `MiddlewareAccessControlManager` unconstructable too: `__init__` opened with `assert self._db_node is not None`, reading an attribute that did not exist yet and asserting non-None on a value it set to `None` two lines later. Two sibling asserts had the same misplacement. The narrowing asserts that sit **after** initialization are correct and are kept.
- **Two more nodes had the identical defect, found by sweeping for the SHAPE rather than the symbol.** An AST sweep for "a narrowing `assert self._X is not None` standing before whatever assigns `_X`" returned exactly three classes in `src/kailash`; the other two were live and are fixed here. `EnterpriseAuditLogNode` could not be constructed at all (`AttributeError: 'EnterpriseAuditLogNode' object has no attribute '_db_node'`), and `UserManagementNode.run` asserted before its own `_init_dependencies`, so every operation died on the `None` the constructor sets with no message to say why (`NodeExecutionError: Node 'bulk_create' execution failed: AssertionError:`). Both surfaced through `tests/integration/middleware/`. A symbol grep would not have found either. The sweep now returns clean.
- Security-event emission on the auth failure path is now best-effort, so a database it may not have cannot convert a clean deny into an exception escaping an authentication call.

## Rebased onto #2139 / #2145 — and the resolution that had to be taken

This branch was authored before #2139 (server-derived gateway identity) merged. Both PRs independently fixed the SAME two root causes — the sync/async `verify_token` await bug and the `user_id`-vs-`sub` claim bug — so the conflict was semantic, not textual.

**#2139 won, and this branch's versions were deleted.** #2139 replaced `get_optional_current_user` with `_authenticated_user_id` + `_resolve_identity` + a shared `subject_from_claims`. This branch had edited the same function with a local `_verify` helper and an inline `payload.get("user_id") or payload.get("sub")`. The tempting resolution — keep main's route AND this branch's helper — passes every test while silently restoring the duplicated claim precedence that `subject_from_claims` exists to eliminate. It was not taken. Both were deleted; everything routes through the shared helper.

Invariant, verified rather than asserted:

```
$ grep -rn 'for claim in ("sub", "user_id", "uid")' src/kailash/
src/kailash/trust/auth/jwt.py:89:    for claim in ("sub", "user_id", "uid"):     <- ONE definition site

$ grep -c "_verify(\|get_optional_current_user" src/kailash/middleware/communication/api_gateway.py
0                                                                                <- zero survivors
```

The gateway contains no direct subject read at all; both arms call `subject_from_claims`. (`sso/google.py:185` and `sso/apple.py:337` read `data.get("sub")` — checked and ruled out: a single-claim read of an external OIDC provider's userinfo, no precedence chain, different issuer.)

Also preserved from main, deliberately: `api_gateway.py:662` still refuses `token_type == "refresh"`, and `asgi.py:508-531` still populates `scope["state"]["user"]` on the websocket API-key path.

**One #2114 site added by #2139 was bounded with the rest.** #2139's new websocket API-key branch logs `jwt_ws_auth_middleware.api_key_user_construction_failed` with a raw `logger.exception`. Its claims come from the key the caller presented, so leaving it unbounded would have kept the exact #2114 amplification alive on a new line. Every exception sink in `kailash.trust.auth.asgi` is now bounded.

**The X-API-Key arm was mutation-verified, not merely test-verified.** A passing test proves nothing about whether the arm is load-bearing, so the arm was disabled with a one-line mutation, the mutation was proven to reach the loaded source, and the suite was re-run:

```
$ python3 -c "import inspect, kailash...api_gateway as g; print('MUTATION present in loaded source:',
              'MUTATION: arm disabled' in inspect.getsource(g.APIGateway._authenticated_user_id))"
MUTATION present in loaded source: True

FAILED ...test_session_route_accepts_an_issued_api_key
FAILED ...test_session_route_survives_a_manager_without_api_key_support
2 failed, 29 passed
```

Restored from a `cp` backup afterwards, not `git checkout --`.

## Tests — red then green

Both files are behavioural against a real ASGI stack and a real store. **No `Mock` anywhere**: a mock validator or store accepts every call and passes identically whether the fix is present or not. Symbols introduced by the fix are imported function-locally (and the one autouse fixture degrades) so the files produce a genuine behavioural RED rather than one collection error.

RED measured on a scratch worktree at the base commit `d248afd78`, with the two test files copied in and nothing else changed:

```
tests/regression/test_issue_2108_api_key_verification.py   22 failed, 2 passed
tests/regression/test_issue_2114_non_ascii_api_key.py       6 failed, 3 passed
```

The passing ones are the intended discrimination controls — pre-fix *everything* 401s, so "an unissued key is rejected" and "a valid ASCII key logs no traceback" were already true. Without them the REDs could equally be broken routes.

GREEN on this branch:

```
$ python -m pytest tests/regression/test_issue_2108_api_key_verification.py \
    tests/regression/test_issue_2114_non_ascii_api_key.py -q
40 passed
```

Re-measured after the rebase onto `54513d077`:

```
tests/regression/{2108, 2114, 2102, 2145}                       99 passed
15 auth / gateway / jwt regression files                       471 passed
tests/unit/middleware + tests/trust/unit                      3110 passed
pre-commit run --files <all 15 branch files>                  exit 0
```

**No regressions, attributed by measurement rather than by inspection.** A throwaway worktree was built at `origin/main` and the identical command run on both:

- `tests/regression/` (full): **86 failures on main, 86 on branch, set difference empty in both directions.** The 86 are the unbuilt Rust extension (`kailash._kailash`) and absent sub-packages (`kailash.kaizen`, dataflow) under a PYTHONPATH-over-source setup.
- `tests/integration/middleware` + `unit/middleware` + `unit/nodes/admin` + `trust/unit`: main **4 failed / 11 errors** → branch **5 failed / 6 errors**. Strictly better; zero new red tests. The five setup ERRORS that disappear are the `EnterpriseAuditLogNode` fix above; two of those tests now get past setup and fail later, on a defect described under "Known, filed, not fixed here".

CI parity before first push: `ruff check src/ tests/ --select=E9,F63,F7,F82` → `All checks passed!`; the CI Tier-1 command → `5074 passed`, with 4 pre-existing observability failures reproduced identically on the base commit and caused by a missing local extra (`Prometheus reader unavailable; pip install kailash[telemetry] ...`), not by code.

## Known, filed, not fixed here — #2166

Fixing the `EnterpriseSecurityEventNode` assert made `MiddlewareAccessControlManager` constructible for the first time, and the code behind it turns out to be the same never-ran class. `check_session_access` calls `PermissionCheckNode.execute()` with three kwargs the node does not declare (`user_context`, `resource_type`, `action`), without the one it requires (`operation`), and reads `result.get("allowed")` where the node returns `{"result": {"allowed": ...}}`:

```
NodeValidationError: Required parameter 'operation' not provided.
Available resolved inputs: ['resource_id', 'cache_level', ...]
```

`check_workflow_access`'s `audit_node.execute(event_type=...)` at `access_control.py:129` is the same shape and should be swept with it.

**Not fixed here on purpose.** The correct call needs a `permission=` string, and choosing that convention is an authorization-schema decision — guessing it would ship a silently *wrong* authz check in place of a loud failure, and the only covering test needs Postgres, so there was no feedback loop to check a guess against. Filed as **#2166** with the full measurement. This is the shard bound in `autonomous-execution.md` MUST-4 applying, not a deferral.

It is also why two tests in `test_user_management_middleware_integration.py` are still red on this branch. They were red on `main` too — as fixture setup ERRORS. Here they get past setup and fail later, on #2166.

## Related issues

Fixes #2108
Fixes #2114
Refs #2166, #2138, #2103, #2102, #2047, #2139, #2145

## CodeQL — dismissed as a false positive; three candidate fixes measured-refuted

**Status: DISMISSED as a false positive. The merge-blocking language that stood in this section is retired.** Receipt, read from the alerts API rather than inferred from a green check:

```
$ gh api repos/terrene-foundation/kailash-py/code-scanning/alerts/11547
number           : 11547
rule             : py/weak-sensitive-data-hashing
state            : dismissed
dismissed_reason : false positive
dismissed_at     : 2026-08-16T11:19:00Z
dismissed_by     : esperie
path             : src/kailash/middleware/auth/api_keys.py:199
comment          : Rule premise false: the input is never a password. hash_api_key only
                   receives secrets.token_urlsafe(32) (256-bit CSPRNG). A slow KDF here
                   is a DoS amplifier - ...
```

Proof recorded in #2146. Nothing in the rebase touched `api_keys.py`.

Kept here because the *reasoning* is the durable part, and the next reader will otherwise re-propose the same three fixes. The alert was `py/weak-sensitive-data-hashing` (`rule_severity: warning`, `security_severity: high`) at `src/kailash/middleware/auth/api_keys.py`. All three candidates were ATTEMPTED and measured, not assumed away:

1. **Keyed HMAC — refuted.** The key was redesigned to `sk_<key_id>.<secret>` with a per-record-salted HMAC-SHA256 digest. The alert simply moved: `api_keys.py:117` (`hashlib.sha256`) → `api_keys.py:170` (`hmac.new`). CodeQL models HMAC under this rule too.
2. **A sanitizer-pack model — refuted by this repo's own recorded evidence.** `.github/codeql/sanitizers/sanitizers.model.yml` documents both extension kinds already measured not to work: `summaryModel` kind `value` propagates taint (the opposite of intent — the file notes its two existing `dataflow.utils.masking` rows "never cleared the findings they were added for"), and `neutralModel` has no effect on an in-source function because CodeQL analyses the real body. That file's own warning applies: *"a model row that does nothing looks exactly like a fix and stops the next person from looking for the real one."*
3. **An inline suppression comment — refuted.** `# codeql[py/weak-sensitive-data-hashing]` at the call site was pushed to be measured rather than asserted (no CodeQL CLI locally). The alert moved seven lines and persisted: `api_keys.py:170` → `api_keys.py:177`. The suppression was then removed and only the reasoning kept.

Only a genuine slow KDF satisfies the rule, and it is rejected on merit: `verify_api_key` hashes on **every request including every unauthenticated one**, so a ~10<sup>2</sup> ms KDF hands an anonymous caller a CPU-burn amplifier in exchange for defeating an offline guessing attack that 256 bits of CSPRNG entropy already defeats. A low iteration count would satisfy the scanner while providing none of the protection it asks for.

The split-key redesign is retained regardless of the alert, on its own merits: `key_id` is public, so an audit line can name *which* key was used without handling the secret; the secret is compared in constant time; salting is per-record; and finding a record is **not** authentication — a real `key_id` with a wrong secret is refused, and revocation requires the secret too, so reading an audit log cannot let anyone disable another principal's key.


esperie:
## CodeQL disposition — `py/weak-sensitive-data-hashing` deferred, safe per #2146

Recorded here to satisfy `zero-tolerance.md` Rule 1b, which requires a written runtime-safety proof citing guard lines, a tracking issue, a PR-body link, and release-specialist signoff. The first three are below and in #2146; **signoff is outstanding and this PR should not merge on my say-so alone.**

The alert:

```
py/weak-sensitive-data-hashing | warning
src/kailash/middleware/auth/api_keys.py:95
Sensitive data (password) is used in a hashing algorithm (SHA256) that is
insecure for password hashing, since it is not a computationally expensive
hash function.
```

**Runtime-safety proof.** The rule's premise is that the hashed value is a password. It is not:

1. **Every stored key is CSPRNG-generated, never caller-chosen.** `MiddlewareAuthManager.create_api_key` is the only writer to the store, takes no key argument, and generates `f"sk_{secrets.token_urlsafe(32)}"` — 256 bits — at `src/kailash/middleware/auth/auth_manager.py:421`, hashing it at `:432`. Recovering a key from its digest means inverting SHA-256 over a 256-bit uniform space, so the offline guessing attack a slow KDF exists to defeat has nothing to work against.
2. **A slow KDF here would be a net security regression.** `verify_api_key` hashes on every request, including unauthenticated ones. bcrypt/scrypt/argon2 are expensive by design (~10² ms), so adopting one lets an anonymous caller burn a full KDF of server CPU per garbage key — a DoS amplifier traded for protection the entropy already provides.

This is **not** the "same on main, therefore fine" dismissal Rule 1a blocks. The argument is about this input class and would hold identically had the alert first appeared on main. (For context: `main` already carries an open instance of this rule and 100+ open alerts overall, but that is not the reason given here.)

**Why not just silence it.** A repo-wide `query-filters` exclusion is explicitly rejected in #2146 — it would stop the rule firing for genuine password hashing everywhere else. The two acceptable full fixes (a keyed HMAC with a deployment-managed pepper, or a model in the repo's existing `.github/codeql/sanitizers` pack — the same mechanism already used for the `py/clear-text-logging-sensitive-data` false positives) both carry decisions beyond this PR's scope, and #2146 holds their acceptance criteria plus the requirement to sweep the sibling site at `packages/kailash-kaizen/.../metrics_auth.py:74,87` with whichever answer is chosen.

The reasoning also now lives in the `hash_api_key` docstring (`ff3b398f`), so a future reader meets it at the line rather than in a scanner dashboard.

esperie:
## CodeQL alert 11555 — analysis, not yet dispositioned

`py/clear-text-logging-sensitive-data` at `api_gateway.py:813`:

```python
logger.warning(
    "api_gateway.session_ownership_denied",
    extra={
        "session_id": sanitize_log_value(session_id, 128),
        "principal": sanitize_log_value(principal, 128),
    },
)
```

**The inversion worth noting: this PR is what makes that line benign.** Before it, possession of a `session_id` was sufficient to act on another user's session — that is the IDOR in #2145. After it, the gateway requires `owner == principal`, so a `session_id` appearing in a log grants an attacker nothing on its own. CodeQL flags it on a sensitive-*name* heuristic, and reports it as new only because the surrounding code changed.

**Not claiming it is therefore fine.** Two things are still true:

1. `sanitize_log_value` is explicitly **not** a redaction step — it bounds and renders, it does not withhold. That is the exact distinction PR #2176 turns on.
2. A full session id at WARN still lands in an aggregator, which is privacy-adjacent correlation data even when it is not a credential.

**Why I did not just patch it:** the obvious fix — log a fingerprint instead of the raw id — would route through `fingerprint_secret` (`url_credentials.py:667`), which is unkeyed BLAKE2b and carries its own open `py/weak-sensitive-data-hashing` alert. Fixing one alert by importing another is not a fix. Dropping the id entirely would remove the operator's ability to investigate the very denial being logged.

**Recommendation:** keep the line as-is and dispose of the alert as a false positive, on the ground that this PR removes the property CodeQL assumes (possession ⇒ access). The alternative — a keyed fingerprint helper — is worth doing, but it belongs with the `fingerprint_secret` work, not bolted onto an auth PR that is otherwise ready.

Cons of that recommendation, stated plainly: a dismissed alert is invisible to future scans of that line, so if session ids ever become bearer-like again the log site will not re-flag. That risk is what the ownership check now structurally prevents, but it is a real dependency between the two.

Dismissal is a repo-settings mutation and is the co-owner's call, so this is surfaced rather than actioned.

Issue #2119, opened 2026-08-12
test_loc_invariants::test_base_agent_loc is RED on main — base_agent.py is 1068 lines against a 1015 limit


esperie:
## Measured on main

```
packages/kailash-kaizen/src/kaizen/core/base_agent.py : 1068 lines
tests/regression/test_loc_invariants.py:18            : limit = 1015  # 882 post-refactor + 15% margin
```

`tests/regression/test_loc_invariants.py::test_base_agent_loc` fails on `main`. Reproduced independently by **two** lanes this session, each on a throwaway worktree at the then-current main with their own branches absent — so it is not attributable to any in-flight PR.

## Why it is not already fixed

`base_agent.py` grew past the cap through the #2030 / observability work. Every lane that hit it this session was working in a different file and correctly declined to fix it, because editing that file from an unrelated branch guarantees a collision with whichever lane owns it. It needs a deliberate owner, not an opportunistic fix.

## Why the guard exists

The docstring states the intent: *"base_agent.py must stay under 1015 LOC after convergence extraction"* — 882 post-refactor plus a 15% margin. So this is a **deliberate architectural invariant**, not an arbitrary lint. Two dispositions are legitimate and they are genuinely different decisions:

1. **Extract** — restore the invariant by moving cohesive responsibility out of `base_agent.py`, which is what the guard was written to force.
2. **Re-baseline** — decide the class has legitimately grown, raise the limit, and record why in the test's docstring so the next reader knows the number moved on purpose.

**Silently raising the number is the disposition to avoid**: it converts a guard into a ratchet that records whatever happened rather than what was intended.

## Acceptance criteria

- [ ] `pytest tests/regression/test_loc_invariants.py` passes on `main`.
- [ ] If re-baselined, the new limit and the reason are stated in the test docstring, and the extraction that was expected is either done or explicitly abandoned with a rationale.
- [ ] No other lane's in-flight file is touched to achieve it.

## Note on visibility

This is red on `main` and has been for some time. It surfaces in local full-suite runs but the root pytest selection in CI does not gate on it — worth confirming whether `tests/regression/` reaching CI (blocked behind #2002) would have caught it earlier.

Issue #2075, opened 2026-08-11
DataFlow pool leak on auto_migrate='warn' DDL-failure path: pool_count()=9 exceeds cap of 5


esperie:
Surfaced by turning on the root `tests/regression/` CI gate (#2002). This test
has **never run in CI** — it is one of the ~154 files #2002 identified as
uncovered — and it fails deterministically against a real Postgres.

## Reproduction

Postgres reachable at the file's default DSN
(`postgresql://test_user:test_password@localhost:5434/kailash_test`):

```
$ .venv/bin/python -m pytest \
    tests/regression/test_dataflow_pool_bridge.py::test_failed_ddl_with_warn_mode_still_bounded \
    -q -p no:xdist --timeout=300
```

**3 of 3 consecutive runs FAILED** (87.23s / 89.11s / 84.96s) with:

```
E   AssertionError: Pool leaked in warn mode: pool_count()=9 > 5 after 10
    DDL-failing DataFlow instances with auto_migrate='warn'
tests/regression/test_dataflow_pool_bridge.py:181
```

Deterministic, not a flake.

## What the test asserts

`tests/regression/test_dataflow_pool_bridge.py::test_failed_ddl_with_warn_mode_still_bounded`
creates 10 concurrent `DataFlow(pg_dsn, auto_migrate="warn")` instances whose
DDL fails (deliberate FK misordering), `await db.close_async()` on each, then
asserts the process-wide pool cap holds:

```python
set_pool_defaults(max_pool_count_per_process=5, idle_timeout=30)
...
assert AsyncSQLDatabaseNode.pool_count() <= 5
```

Observed `pool_count() == 9` against a cap of 5.

## Why it matters

`max_pool_count_per_process` is a bound, not a hint. On the `auto_migrate="warn"`
path — the legacy log-and-continue sentinel — a DDL failure leaves pools behind
that `close_async()` does not reclaim, so a long-lived process that repeatedly
constructs DataFlow instances against a failing schema exhausts Postgres
connections. The sibling test in the same file for the non-warn path passes, so
the leak is specific to the warn branch.

## Interim disposition

Marked `@pytest.mark.xfail(strict=True)` referencing this issue so the new
root-regression CI gate starts green and this test **self-clears loudly**
(XPASS -> failure) the moment the leak is fixed. It is NOT skipped and NOT
weakened — the assertion is unchanged.

## Acceptance criteria

[...]

Issue #1572, opened 2026-07-05
fix(dataflow): MySQL adapter pool not drained on close_async (Event loop is closed at GC)


esperie:
## Affected API

`dataflow.core.engine.DataFlow.close_async()` + `dataflow.adapters.mysql.MySQLAdapter`

## Summary

`MySQLAdapter.close_connection_pool()` drains its aiomysql pool correctly
(`pool.close()` + `await pool.wait_closed()`), but `DataFlow.close_async()` never
invokes it — that drain is only reached at the tail of `_discover_schema_real`. A
MySQL adapter pool created outside a completed schema-discovery therefore leaks its
connection to GC, where `aiomysql.Connection.__del__` fires `close()` on an
already-closed event loop → `RuntimeError: Event loop is closed`.

## Reproduction

Any real-MySQL DataFlow flow that calls `await db.close_async()` in teardown emits, at GC:

```
PytestUnraisableExceptionWarning: Exception ignored in: <function Connection.__del__ ...>
  aiomysql/connection.py:339 self._writer.transport.close()
  ... RuntimeError: Event loop is closed
```

(Surfaced by the #1503 conversion of `test_production_dataflow.py` to real MySQL; the
prior mock engine held no real connection so the gap was invisible.)

## Expected vs actual

- Expected: `close_async()` fully drains every live adapter pool (as `close()` /
  `close_connection_pool()` already do).
- Actual: the MySQL adapter pool is not drained by `close_async()`; the connection
  leaks and warns at GC.

## Severity

MEDIUM — resource-cleanup gap; no data-correctness impact, but leaks connections and
emits a noisy unraisable warning under any real-MySQL usage relying on `close_async()`.

## Acceptance criteria

- [ ] `close_async()` drains the MySQL adapter pool (`await close_connection_pool()` or equivalent).
- [ ] A real-MySQL regression test asserts no `ResourceWarning`/`RuntimeError` at teardown.

## Cross-SDK

The Rust SDK MySQL binding may share the equivalent connection-lifecycle gap — inspect per cross-sdk-inspection.

PR #2147, opened 2026-08-15
fix(async-sql,runtime): root-cause the CI hang chain and restore both regression gates (#2081, #2079)

### LINKED ISSUE #2079: Root regression infra-marked slice hangs on CI in DataFlow cross-loop pool disposal — blocks #2002 AC#2
Blocks AC#2 of #2002 ("infra-requiring root regression tests run in the
infrastructure workflow"). Filed from PR #2077.

## Symptom

Running the infra-marked slice of root `tests/regression/` against Postgres +
Redis service containers on `ubuntu-latest` **hangs**. The step was given a
30-minute budget and consumed all of it having executed **3 of 22** tests:

```
collected 1940 items / 1918 deselected / 1 skipped / 22 selected

tests/regression/test_dataflow_pool_bridge.py::test_failed_ddl_does_not_leak_pools_under_saturation PASSED [  4%]
tests/regression/test_dataflow_pool_bridge.py::test_failed_ddl_with_warn_mode_still_bounded FAILED [  9%]
tests/regression/test_issue_1248_localruntime_cross_loop_pool_disposal.py::test_issue_1248_worker_thread_localruntime_preserves_live_loop_pool PASSED [ 13%]
+++++++++++++++++++++++++++++++++++ Timeout ++++++++++++++++++++++++++++++++++++
##[error]The action 'Run ROOT regression infra-marked tests' has timed out after 30 minutes.
```

Job: run `31483354916`, `Test DataFlow Infra Regression (Postgres/Redis)`.

## Where it hangs

`pytest-timeout` fired its 300s banner and dumped thread stacks, but the process
did not die — `timeout_method = thread` cannot kill a thread blocked in a C
call, so the step ran to the job cap. The dump names a DataFlow async pool
thread:

```
~~~~~~~~~~~~~~~~~ Stack of dataflow_async__0 (140131247576768) ~~~~~~~~~~~~~~~~~
  File ".../threading.py", line 1002, in _bootstrap
```

and an idle executor worker:

```
~~~~~~~~~~~~~~~~~~~~~ Stack of asyncio_0 (140130136352448) ~~~~~~~~~~~~~~~~~~~~~
  File ".../concurrent/futures/thread.py", line 81, in _worker
    work_item = work_queue.get(block=True)
```

The hang is at or just after
`test_issue_1248_localruntime_cross_loop_pool_disposal.py` — i.e. in
cross-event-loop pool disposal, which is exactly what that file exists to pin.

## Does not reproduce locally

The identical selection against a real Postgres 16 on macOS completes in 244s:

```
11 passed, 11 skipped, 1 xfailed, 1915 deselected in 237.06s
```

So this is Linux-specific, container-specific, or both. Plausibly the same
resource-exhaustion family as #2078 (`can't start new thread` /
`sqlite3: disk I/O error` in Tier 2), but that is a HYPOTHESIS — the two have
not been shown to share a cause.

## Secondary observation (unresolved)

`test_failed_ddl_with_warn_mode_still_bounded` is marked
`xfail(strict=True)` for #2075, and reports `XFAIL` locally — but the CI log
shows it as **FAILED**. Either it fails on CI in a phase the xfail marker does
not cover (e.g. fixture setup), or for a different reason than the pool-count
assertion. The step never completed, so no traceback was emitted. Worth
resolving alongside the hang, because it means the step would not have started
green even without the hang.

## Interim disposition

PR #2077 **removes** the infra-marked step rather than shipping it
non-blocking. A step that holds a runner for 30 minutes on every PR is a real
cost even when it cannot fail the build, and a hang produces no useful signal
in exchange. The infra-FREE root regression gate — the substance of #2002 —
lands and is unaffected: it excludes exactly this marker set.

## Acceptance criteria

- [ ] Reproduce the hang off-CI (Linux container, Postgres + Redis service
      containers, same install set) so it can be debugged without a CI cycle.
- [ ] Identify what blocks in cross-loop pool disposal on Linux.
- [ ] Resolve the `xfail`-reported-as-`FAILED` discrepancy above.
- [ ] Restore the infra-marked step in `unified-ci.yml` (the removed block is in
      PR #2077's history, with its `env:` wiring for the four different DSN env
      var names those files read) and confirm it completes green.

### LINKED ISSUE #2081: Root tests/regression/ hangs on ubuntu CI runner (30+ tests, run never completes) — blocks #2002
Blocks #2002. Filed from PR #2077, which added the CI gate #2002 asks for and
then could not make it complete on the runner.

## Claim

Root `tests/regression/` **cannot run to completion on this repo's
GitHub-hosted `ubuntu-latest` runner.** The identical selection completes in
217s on macOS with zero failures.

| environment | result |
| --- | --- |
| macOS, from-scratch CI-equivalent venv, Python 3.12 | `1916 passed, 3 skipped, 22 deselected, 0 failed in 217.79s` |
| `ubuntu-latest`, same install block, Python 3.12 | hangs; see below |

## How it was established

Three CI runs, each tightening the instrument:

1. `-q`, `--timeout=300`, `timeout_method=thread` (the repo default), 15-minute
   step budget → **timed out**, no per-test output, nothing learnable.
2. `-v`, 40-minute budget → **timed out at 40 minutes having run ~8%**.
   pytest-timeout printed its 300s banner repeatedly while the process carried
   on: `timeout_method = thread` **cannot kill a test blocked in a C call**, so
   the suite had a timeout that could not enforce one.
3. `--timeout-method=signal` (SIGALRM), which can → hung tests now fail and
   name themselves.

Run 3, `--maxfail=30`, `--timeout=60`:

```
13 failed, 757 passed, 2 skipped, 22 deselected, 17 errors in 1420.07s (0:23:40)
```

Still truncated at `--maxfail=30` with only 757 of 1916 executed, so **30 is a
floor, not the count.**

## The 30 enumerated so far

| file | count |
| --- | --- |
| `test_create_index_identifier_safety.py` | 10 (errors) |
| `test_issue_1245_sqlite_store_per_thread_conn_leak.py` | 5 |
| `test_channel_parameters_envelope_behaviour.py` | 5 |
| `test_issue_1185_async_resume_short_circuit.py` | 3 (errors) |
| `test_issue_1708_g1_pool_metrics_production_path.py` | 2 |
| `test_eatp12_vault_quickstart.py` | 2 (errors) |
| `test_eatp12_vault_630_clearance_wiring.py` | 2 (errors) |
| `test_issue_1035_delegate_m4_tenant_hash_salt.py` | 1 |

Every one reports `Failed: Timeout (>60.0s) from pytest-timeout`. 60s is 3x the
slowest healthy test in the suite (19.94s,
`test_issue_2015_http_exception_leak.py`), so these are hangs, not slow tests.

The spread — sqlite connection pooling, Nexus channel envelope binding, async
resume, pool metrics, EATP vault, delegate tenant hashing — is too wide for
eight independent bugs. It looks like one environment-level cause.

## Probable relationship to #2078

#2078 records the Tier-2 suite failing on the same runner with 130
`RuntimeError: can't start new thread` and 90
`sqlite3.OperationalError: disk I/O error` — thread and file-descriptor
exhaustion. A process that cannot start a thread will *block* rather than raise
in some code paths, which would present exactly as these hangs. **That is a
hypothesis, not a finding** — the two have not been shown to share a cause, and
the regression hangs produced no `can't start new thread` message.

## Interim disposition

PR #2077 does **not** ship the `test-root-regression` job. A 40-minute job that
cannot complete is pure cost on every PR, and non-blocking would only make it
silent cost. So #2002's gap remains open and this issue is its blocker.

What that PR does land, which is independent of this: 45 genuinely-broken Tier-2
tests fixed, the #2023 sibling-install declarations plus their lint, corrected
contributor docs, and the diagnosis above.

## Acceptance criteria

- [ ] Reproduce off-CI. A Linux container with the runner's CPU/memory shape and
      `ulimit -u` / `-n` is the fastest path; this does not reproduce on macOS.
- [ ] Determine whether these hangs and #2078's exhaustion share a cause.
- [ ] Fix, then restore the `test-root-regression` job from PR #2077's history
      (it is complete there, including the sibling installs and the `shamir` /
      `kaizen-agents` provisioning that the suite needs to collect at all).
- [ ] Consider changing `timeout_method` from `thread` to `signal` in the root
      `pytest.ini` generally. A thread-method timeout cannot enforce a timeout,
      which is what made two CI runs unreadable.

esperie:
## Summary

The root regression suite could not complete on the CI runner. Two independent causes; both are now closed, and both removed CI gates are restored.

- **#2081** — the infra-free suite hung on `ubuntu-latest`, 30+ tests, never completing. Root cause was the process-wide `RLIMIT_AS` cap closed by #2078/#2097. **Re-measured on Linux on this branch: it no longer hangs.** All 30 previously-enumerated hangs pass.
- **#2079** — the infra-marked slice hung in DataFlow pool disposal. **Reproduced, root-caused, and fixed here.**

## #2079 — the hang, reproduced and root-caused

Reproduced in a Linux container (Postgres 16 + Redis 7) at `tests/regression/test_issue_697_pool_leak.py::test_pool_count_stays_bounded_under_lock_contention`, in isolation. Captured pending-task dump, ~10 of these at once, with the main thread parked in `selectors.select`:

```
---- TASK 'Task-633' done=False ----
  File ".../asyncpg/pool.py", line 268, in close
    await self._con.close()
Stack for <Task pending name='Task-633' coro=<PoolConnectionHolder.close() running at .../asyncpg/pool.py:268> wait_for=<Future pending ...>>
---- TASK 'Task-3' done=False ----
  File "_taskdump.py", line 38, in _one_query
    await node._adapter.disconnect()
```

Preceding log evidence from the same run:

```
asyncpg.exceptions._base.InternalClientError: got result for unknown protocol state 3
  File "src/kailash/nodes/data/async_sql.py", line 1672, in _release_quietly
    await self._pool.release(conn)
AttributeError: 'NoneType' object has no attribute 'release'
```

**Every `disconnect()`/`close()` in `async_sql.py` read its pool handle, awaited the driver close, and nulled the handle *afterwards*.** `_shared_pools` hands ONE adapter to every node on the same DSN + loop, so N concurrent callers all passed the `if self._pool:` guard and entered the driver close together while the other N−1 still had queries in flight.

For asyncpg that is unrecoverable. `Pool._check_init` gates on `_closed`, which `Pool.close()` only sets in its `finally`, so `acquire()` keeps succeeding for the whole duration of a close: a holder released past `wait_until_released()` is re-acquired by another task, `PoolConnectionHolder.close()` then runs against a connection with a live query → `_request_cancel()` → protocol state 3 → `InternalClientError`, and the stranded `await self._con.close()` waits forever for an acknowledgement that can never arrive.

**Three nested guards had the identical shape** and all had to change — `ProductionPostgreSQLAdapter.disconnect` (the outermost, and the one that actually admitted the 50 callers), `EnterpriseConnectionPool.close`, and `PostgreSQLAdapter.disconnect` — swept across every dialect.

### The fix

1. **Claim-then-close** (the root fix) — bind the handle to a local and null the attribute *before* awaiting, so exactly one caller drives the driver close.
2. **Bound the graceful close**, escalating to the driver's forced `terminate()` — the backstop for a connection already wedged by some other cause, and what asyncpg's own `Pool.close()` docstring advises. Configurable via `set_pool_defaults(close_timeout=…)`, default 5 s.

Losers wait on the winner's completion rather than returning early, so `disconnect()` stays a drain barrier and the bridge teardown cannot close the loop underneath an in-flight close (#1572). They wait on a **shielded** future, never the winner's task: in-tree callers wrap `disconnect()` in their own 1 s `wait_for`, and cancelling a waiter must not abandon the real close half-done.

Two defects the claim *widened*, so they land here too:
- `EnterpriseConnectionPool.get_connection`/`initialize` read `_pool is None` as "not yet initialized" and would have built a replacement pool mid-shutdown → both now fail closed on `_shutdown`.
- `_release_quietly` raised an opaque `AttributeError` off the nulled pool, leaving the orphaned connection with no disposition → it now terminates it and says so.

## Three checks that could not come out the other way

Found while making the infra step actually run. Each made a step green by construction.

1. **The DPI-B3 idle-pool reaper was a complete no-op.** `_PROCESS_POOL_REGISTRY` stores *adapters*; `is_idle()` lives on `EnterpriseConnectionPool`. The reaper tested `hasattr(entry, "is_idle")`, which is False for every entry it can ever hold — it walked the registry and skipped 100% of it, silently. `test_idle_pools_reaped` pins this and had never run in CI.
2. **The root conftest's Postgres probe hardcoded `:5434`.** It gates the collection-time skip for *every* `requires_postgres` test, so a runner publishing `:5432` had all of them skipped no matter what the job set in `env:`.
3. **`test_dataflow_pool_bridge.py`'s autouse `_verify_docker_services` discarded the helper's return value.** The helper prints and returns False rather than raising, so the fixture named "verify" verified nothing. **This is the whole of #2079's "xfail reported as FAILED"**: the test was `[XPASS(strict)]`, never a fixture-phase error.

On (3): the #2075 leak the `xfail(strict=True)` asserts is environment-dependent. Measured on Linux against a `postgres:16` service container: `pool_count()=1` with a pre-existing table, `0` on a clean database — both against a cap of 5. `strict=True` therefore turned every environment where the leak does *not* fire into a hard CI failure. Changed to `strict=False` with the measurements recorded at the marker; the assertion is unchanged and **#2075 stays open**.

## #2081 — the legibility defect

`thread.join()` at `local.py:2336` had no timeout and, more importantly, **no output, ever**. Exceptions inside are captured and re-raised after the join, so a raise could never hang the caller; only a block could, and then the wait was permanent and silent. That is why three CI runs produced 15 and 40 minutes of wall clock and zero actionable output.

The join is now sliced, logging a WARNING with the workflow name and the stuck thread's stack each interval. **The default stays unbounded on purpose** — a workflow may legitimately run for hours, and truncating one would trade a visible hang for invisible data loss. `LocalRuntime(sync_bridge_timeout=…)` opts into a hard bound that raises `RuntimeExecutionError`.

The bridge thread is now a **daemon**, required *by* the bound: once it expires we abandon a live thread, and a non-daemon one would be joined by `threading._shutdown` at interpreter exit — re-creating the same unbounded wait at process teardown.

`pytest.ini` now sets `timeout_method = signal` (#2081 AC#4). A thread-method timeout cannot kill a test blocked in a C call. SIGALRM is POSIX-only, so the root `conftest.py` degrades it back to `thread` with a `RuntimeWarning` where SIGALRM does not exist, rather than hard-breaking a Windows contributor.

## Measurements

All on Linux (aarch64 container, 2-CPU quota), Postgres 16 + Redis 7.

| selection | before | after |
| --- | --- | --- |
| infra-free root regression | hung on CI; 30 hangs enumerated across 8 files, 757 of 1916 executed | `1 failed, 2346 passed, 4 skipped, 23 deselected in 107.64s`, **zero** pytest-timeout hits |
| infra-marked slice | `3 failed, 17 passed` in **131.54s** incl. a 120 s hang (30 min / 3-of-22 on CI) | `20 passed, 3 skipped, 1 xpassed in 12.97s`, **exit 0** |
| `test_issue_2079_…` (new) | `5 failed in 151.21s` — two via pytest-timeout, i.e. genuine hangs | `5 passed in 6.57s` |
| `test_issue_2081_…` (new) | `6 failed` | `6 passed` |

The one remaining infra-free failure is `test_loc_invariants.py::test_base_agent_loc` (**#2119**) — red on main, reproduced identically against unmodified code, and owned by the CI-verification lane. **It will block this restored gate until #2119 lands.**

### On the new tests' discriminating power

For `test_issue_2079_…`, all five fail pre-fix on **behaviour** — `assert 50 == 1` (fifty overlapping driver closes), and two via pytest-timeout, i.e. they genuinely hang.

For `test_issue_2081_…`, five of six fail pre-fix on an **absent symbol**, which proves only that the API is new. `test_a_slow_bridge_is_not_silent_at_the_real_call_site` is the one that demonstrates the *defect*: it uses no new API, drives the real call site, and fails pre-fix on `assert []` — the empty log.

## Not fixed

- **#2075** — the warn-mode pool leak. Does not reproduce in this environment (measured above), so I cannot claim it fixed. Issue stays open; the test still asserts the bound.
- **#2119** — `base_agent.py` at 1068 lines against a 1015 limit. Red on main, out of this shard, owned by the CI-verification lane.
- **Pre-existing isort drift**, 12 files (proven pre-existing: untouched by any commit here, last modified 2026-08-12). CI does not run isort, so it is not a gate. Left alone because one of the files is `base_agent.py`, in flight in another worktree.

## Test plan

- [x] Hang reproduced off-CI in a Linux container (#2081 AC#1, #2079 AC#1) with a captured stack
- [x] Root cause identified in cross-loop pool disposal (#2079 AC#2)
- [x] `xfail`-reported-as-`FAILED` discrepancy resolved (#2079 AC#3)
- [x] Both CI jobs restored (#2081 AC#3, #2079 AC#4)
- [x] `timeout_method` → `signal`, platform-guarded (#2081 AC#4)
- [x] New regression tests fail (by hanging) before the fix and pass after
- [x] `pre-commit run --all-files` green on every file this PR touches

## Related issues

Fixes #2081
Fixes #2079
Unblocks #2002

### DIFF
--- a/.github/workflows/unified-ci.yml
+++ b/.github/workflows/unified-ci.yml
@@ -267,30 +267,113 @@ jobs:
             --maxfail=400 -q -p no:xdist --timeout=60 -rf
 
 
-  # ISSUE #2002 — the `test-root-regression` job lived HERE and is NOT SHIPPED,
-  # pending #2081. Root tests/regression/ cannot run to completion on this
-  # repo's ubuntu-latest runner: three CI runs, each with a sharper instrument,
-  # ended in a timeout or a truncated `--maxfail`. The last one
-  # (--timeout-method=signal, --timeout=60, --maxfail=30) enumerated 30 hangs
-  # across 8 files having executed 757 of 1916 tests in 23m40s — so 30 is a
-  # floor, not a count. The identical selection is 1916 passed / 0 failed in
-  # 217s on macOS.
+  # ISSUE #2002 / #2081 — RESTORED. This job was removed in PR #2077 because
+  # root tests/regression/ could not run to completion on this runner: three
+  # CI runs ended in a timeout or a truncated `--maxfail`, the last one
+  # enumerating 30 hangs across 8 files having executed 757 of 1916 tests.
   #
-  # A 40-minute job that cannot complete is pure cost on every PR, and making
-  # it non-blocking would only make that cost silent. So it is removed rather
-  # than shipped degraded, and #2002 stays open with #2081 as its blocker.
+  # Two causes, both now closed. The environment-level one was #2078: a
+  # PythonCodeNode set a process-wide, irreversible RLIMIT_AS cap, so every
+  # later mmap in the process failed — including the stack pthread_create
+  # needs. That is why it never reproduced on macOS, whose kernel rejects the
+  # setrlimit outright and so is not a discriminating instrument. Fixed in
+  # #2097. The disposal-level one was #2079, fixed in this PR.
   #
-  # The complete job — the extras set (note `shamir`, without which five EATP
-  # vault tests fail), the five editable sibling installs the suite needs to
-  # COLLECT at all, and the marker filter mirroring the DataFlow regression
-  # step — is preserved in PR #2077's history and referenced from #2081.
-  # Restore it from there once the hangs are fixed; do not re-derive it.
-  #
-  # One finding from that work is worth acting on independently of #2002: the
-  # root pytest.ini sets `timeout_method = thread`, and a thread-method timeout
-  # CANNOT kill a test blocked in a C call. Two of the three runs above were
-  # unreadable for exactly that reason — pytest-timeout printed its banner and
-  # the process carried on. `signal` is the enforceable method on Linux/macOS.
+  # Re-measured before restoring, in a Linux container with Postgres 16, on
+  # this branch: `2336 passed, 4 skipped, 22 deselected in 219.36s` with ZERO
+  # pytest-timeout hits, and every one of the 30 previously-enumerated hangs
+  # passing. The 40-minute budget is kept because a runner is slower than that
+  # container, not because the suite is expected to need it.
+  test-root-regression:
+    name: Test Root Regression Suite (Python 3.12)
+    needs: check-duplicate
+    # Skip release-prep PRs (release/v*) per ci-runners.md MUST Rule 8.
+    if: needs.check-duplicate.outputs.should-skip != 'true' && !startsWith(github.head_ref, 'release/')
+    runs-on: ubuntu-latest
+    # Must exceed the 40-minute test step plus install time.
+    timeout-minutes: 50
+    steps:
+      - uses: actions/checkout@v7
+
+      - name: Set up Python 3.12
+        uses: actions/setup-python@v7
+        with:
+          python-version: "3.12"
+
+      - name: Install uv
+        uses: astral-sh/setup-uv@v7
+
+      - name: Restore uv cache
+        uses: actions/cache@v6
+        with:
+          path: ~/.cache/uv
+          key: ${{ runner.os }}-uv-py3.12-rootregr-${{ hashFiles('pyproject.toml', 'uv.lock') }}
+          restore-keys: |
+            ${{ runner.os }}-uv-py3.12-rootregr-
+
+      - name: Install dependencies
+        # Extras are the `test` job's set PLUS `shamir`. The shamir extra is
+        # REQUIRED, not optional garnish: without shamir-mnemonic, five tests
+        # (test_eatp12_vault_3way_discrimination.py x3,
+        # test_eatp12_vault_quickstart.py x2) fail with
+        # `RuntimeError: SLIP-0039 Shamir secret-sharing requires the 'shamir'
+        # optional extra` raised from src/kailash/trust/vault/shamir.py. Those
+        # are real assertions on a real code path, not env noise — the fix is to
+        # provision the dependency, not to skip the tests.
+        #
+        # Root kailash editable BEFORE the sub-packages, per
+        # rules/deployment.md § "Sibling-Package CI Installs Root SDK Editable".
+        #
+        # ISSUE #2023 SIBLING DECLARATION — measured, do not re-derive.
+        # Selection: tests/regression/ minus the infra markers.
+        # Determined by grepping module-level sibling imports across the
+        # directory, then confirmed by building this exact venv from scratch and
+        # running the suite in it:
+        #   dataflow      -> EXERCISED (17 files) -> editable below
+        #   nexus         -> EXERCISED  (7 files) -> editable below
+        #   kaizen        -> EXERCISED  (4 files) -> editable below
+        #   kailash_mcp   -> EXERCISED  (3 files) -> editable below
+        #   kaizen_agents -> EXERCISED  (1 file)  -> editable below
+        #   kailash_ml / kailash_align / kailash_pact -> 0 files, NOT installed
+        # Several of these imports are UNGUARDED at module scope
+        # (`from nexus import Nexus`, `from kaizen.providers.registry import ...`),
+        # so a missing sibling is a COLLECTION error, not a skip.
+        run: |
+          uv venv .venv
+          uv pip install -e ".[dev,server,http-client,db-postgres,db-mysql,db-sqlite,redis,monitoring,auth,auth-azure,trust,mcp,scheduler,telemetry,data,shamir]" --python .venv/bin/python
+          uv pip install -e packages/kailash-mcp --python .venv/bin/python
+          uv pip install -e packages/kailash-dataflow --python .venv/bin/python
+          uv pip install -e packages/kailash-kaizen --python .venv/bin/python
+          uv pip install -e packages/kailash-nexus --python .venv/bin/python
+          # test_issue_443_kaizen_mcp_dependency.py does `from kaizen_agents
+          # import Delegate` — without this the test fails ModuleNotFoundError.
+          uv pip install -e packages/kaizen-agents --python .venv/bin/python
+
+      - name: Test root regression suite (infra-free)
+        timeout-minutes: 40
+        # tests/regression/ is a MIXED dir — the same shape as the DataFlow
+        # regression dir handled below. The explicit `-m` filter mirrors that
+        # step's, and is REQUIRED for the same reason: the root pytest.ini
+        # `addopts` carries NO marker filter at all, so nothing else deselects
+        # the infra-requiring tests. Keep this expression and the infra-marked
+        # step's in sync — they are exact complements, so between the two every
+        # test in root tests/regression/ runs on every PR and none runs twice.
+        #
+        # `--maxfail=30` is deliberately KEPT rather than lifted. It cost
+        # nothing on the verification run (0 failures) and it bounds the damage
+        # if a future change reintroduces a hang class: 30 named failures is a
+        # list you can act on, where an unbounded run is another 40-minute job
+        # with no output.
+        #
+        # `--timeout-method=signal` is no longer passed here — pytest.ini now
+        # sets it globally (platform-guarded in the root conftest.py), because
+        # a thread-method timeout cannot kill a test blocked in a C call and
+        # that is what made two of the three #2081 runs unreadable.
+        run: |
+          .venv/bin/python -m pytest tests/regression/ \
+            -m "not (requires_postgres or requires_mysql or requires_redis or requires_docker or integration or e2e)" \
+            --maxfail=30 -v -p no:xdist --durations=25 \
+            --timeout=60
 
   test-pact:
     name: Test PACT (Python 3.11)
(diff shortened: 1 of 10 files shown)


esperie:
## Verified on the actual `ubuntu-latest` runner

My measurements above were taken in a Linux container, which is not the same instrument as the runner. Both restored jobs have now run on the real one, at `4acd359dfc79e491ff8da1e7c38d55f018143e95`.

**`Test DataFlow Infra Regression (Postgres/Redis)` — SUCCESS.** This is the step #2079 was filed against. It previously consumed a 30-minute budget having executed 3 of 22 tests.

**`Test Root Regression Suite (Python 3.12)` — completes.** This is the claim #2081 was filed against ("cannot run to completion on this repo's GitHub-hosted `ubuntu-latest` runner"):

```
= 1 failed, 2346 passed, 4 skipped, 23 deselected, 63 warnings in 262.09s (0:04:22) =
FAILED tests/regression/test_loc_invariants.py::test_base_agent_loc - AssertionError: base_agent.py: 1068 lines (limit 1015). Code may have been re-inlined by a merge. Check git log for unexpected growth.
assert 1068 <= 1015
```

**262 seconds**, against a 40-minute job timeout that previously fired at ~8% of the suite. Zero pytest-timeout failures, zero hangs. The pass/skip/deselect counts are identical to the container run (`2346 / 4 / 23`), so the container was a faithful instrument after all — but this is the run that settles it.

The single failure is **#2119** (`base_agent.py` at 1068 lines against a 1015 limit). It is red on `main`, reproduces identically against unmodified code, is untouched by this PR, and belongs to the CI-verification lane.

**Merge ordering:** this PR restores a gate that #2119 currently fails. Either land #2119 first, or land them together — merging this alone turns every subsequent PR red on a defect it did not introduce.

esperie:
## Merge-conflict check against L4's #2144 — measured, not assumed

Concern raised: this PR and #2144 both edit `.github/workflows/unified-ci.yml`, so a 3-way conflict is guaranteed. **Measured: it is not.** The two changes touch disjoint regions of the file.

Non-destructive 3-way merge of this branch against #2144's head (`8764d7b00effed9ca451eb2755596078de83a30b`), using `git merge-tree --write-tree --messages`, which does not touch either working tree:

```
merge-tree exit: 0
6eb92c756764637de676e41130fdddf2f2e79bf0

Auto-merging .github/workflows/unified-ci.yml
```

Exit 0 with no `CONFLICT` line. Had the hypothesis been true, `merge-tree` exits non-zero and prints `CONFLICT (content): Merge conflict in .github/workflows/unified-ci.yml` — that is the result it did not produce.

Verified the resulting tree keeps **both** sides rather than silently taking one:

| | present in merged tree |
| --- | --- |
| L4 — `schedule: cron "23 2 * * *"` | line 39 |
| L4 — `permissions: pull-requests: read` (down from `write`) | line 62 |
| L4 — the `#2133` header comment block | present |
| this PR — `test-root-regression:` job | line 287 |
| this PR — `Run ROOT regression infra-marked tests` step | line 652 |

Merged file parses as valid YAML with all six jobs: `check-duplicate`, `test`, `test-dataflow`, `test-dataflow-infra`, `test-pact`, `test-root-regression`.

**Why they do not collide:** #2144's edits are confined to the file header (trigger block, permissions, comments, lines 1-62). This PR's edits are job bodies at lines ~287 and ~652. Nothing overlaps.

**Sequencing unchanged:** #2144 still merges first, and I will rebase and re-confirm on the real merge rather than rely on this dry run. This only establishes that the resolution is mechanical and that L4's nightly `schedule:` trigger is not at risk.

esperie:
## Correction — the `pytest.ini` change was inert, and two green Windows runs proved nothing

Retracting a claim from `a2451c6b5` and from the PR body above. Fixed in `ceed16587`.

### What I claimed

That `pytest.ini` "now sets `timeout_method = signal` globally (platform-guarded in the root `conftest.py`)".

### What is actually true

Root `pytest.ini` carries **both** `timeout` and `timeout_method` inside a **`[pytest:local]`** section. pytest reads **only `[pytest]`**. Measured, identically on this branch and on `main`:

```
config.getini("timeout")        -> ''
config.getini("timeout_method") -> ''
config._env_timeout             -> None
```

So my edit changed a dead letter. `pytest.ini` is now reverted to main's version byte-for-byte, and `--timeout-method=signal` is passed **explicitly** on both restored root-regression steps, where it is actually read.

### How this surfaced, because two green CI runs asserted the opposite

To test whether the `conftest.py` SIGALRM guard was load-bearing on Windows, I pushed a throwaway branch with the guard **removed** and ran the `trust-plane.yml` Windows matrix. All four `windows-latest` legs came back **green** — which reads as "the guard is unnecessary".

It is not. Reproduced directly in an isolated directory: with `timeout_method = signal` and any active timeout on a platform without SIGALRM, pytest-timeout reaches `signal.signal(signal.SIGALRM, handler)` and the run dies before a single test executes:

```
INTERNALERROR> AttributeError: module 'signal' has no attribute 'SIGALRM'. Did you mean: 'SIGABRT'?
============================ no tests ran in 0.02s =============================
```

**Both Windows runs were non-discriminating instruments.** The ini timeout is dead *and* zero tests under `tests/trust/` carry a `@pytest.mark.timeout` marker, so no timer is ever armed and `timeout_method` is never consulted. Neither the green control nor the green mutation could have come out the other way. Per instrument discipline, a non-reddening mutation leaves two hypotheses — inert guard, or inert mutation — and here it was the *instrument* that was inert.

The guard is **kept**, with its docstring corrected to name its real trigger (the command line, not the ini). pytest-timeout has no fallback of its own for a missing SIGALRM — its only coercion is `signal` → `thread` off the main thread — so the guard is what stands between a `--timeout-method=signal` invocation on Windows and a collection-time crash. Verified in both polarities locally: degrades to `thread` with a `RuntimeWarning` where SIGALRM is absent, leaves `signal` untouched where it exists.

### A separate defect this exposed, not fixed here

**The repo's advertised global `timeout = 10` has never applied to any test run.** Same dead-section cause. Moving it into `[pytest]` would arm a 10-second per-test cap across the entire suite for the first time, and several tests in the root regression selection alone exceed it (14.68s, 9.19s, 9.01s). That is a measured shard of its own, not a drive-by change on this PR.

This is the fourth "check that could not come out the other way" found in this lane, after the no-op idle reaper, the hardcoded `:5434` Postgres probe, and the `_verify_docker_services` fixture that verified nothing.

esperie:
## Both restored gates GREEN on `ubuntu-latest` — and the `signal` method is now provably live

At `ceed16587`, after moving `--timeout-method=signal` off the dead ini section and onto the two steps explicitly.

**`Test Root Regression Suite (Python 3.12)` — SUCCESS:**

```
timeout: 60.0s
timeout method: signal
=== 2348 passed, 4 skipped, 23 deselected, 63 warnings in 266.67s (0:04:26) ====
```

**`Test DataFlow Infra Regression (Postgres/Redis)` — SUCCESS.**

Three things this settles, each of which was previously either unproven or wrong:

1. **The suite that "cannot run to completion on this runner" completes in 266.67s**, against a 40-minute job timeout that previously fired at ~8%. **Zero** pytest-timeout hits.
2. **`timeout method: signal` is confirmed in the header.** That is the discriminating evidence the earlier `pytest.ini` edit never had — the dead `[pytest:local]` section produced `getini("timeout_method") == ''` and the header showed no timeout line at all. The CLI flag reaches pytest-timeout; the ini section did not.
3. **`test_base_agent_loc` now passes** — #2119 landed via #2143, and `pull_request` CI runs against the merge of this head with current `main`, so the merge result is green. This removes the merge-ordering blocker I flagged earlier: #2143 is in.

Count moved 2346 → 2348 (the two `test_loc_invariants` tests that were red).

Remaining sequencing: #2144 is still open; I will rebase onto `main` after it merges and confirm — as a positive check, not an absence of conflict markers — that L4's `schedule: cron "23 2 * * *"`, `pull-requests: read`, and the #2133 comment block are all present in the rebased file.

esperie:
## Rebased onto main (`2e3ddd9e7`) — L4's markers confirmed present, all checks green

Head `799a72626`. Rebase was clean (6 commits replayed, no conflict), as the `merge-tree` dry run predicted.

### Positive confirmation of L4's #2144 content — each named as PRESENT

Not "no conflict markers" — each grepped for and located after the rebase:

| L4 marker | present at |
| --- | --- |
| `schedule:` → `- cron: "23 2 * * *"` | `unified-ci.yml:43` |
| `permissions:` → `pull-requests: read` | `unified-ci.yml:62` |
| the `# Issue #2133` comment block | `unified-ci.yml:7` |

Parsed rather than only grepped:

```
schedule: [{'cron': '23 2 * * *'}]
permissions: {'contents': 'read', 'pull-requests': 'read'}
jobs: ['check-duplicate', 'test', 'test-dataflow', 'test-dataflow-infra', 'test-pact', 'test-root-regression']
```

My own two restorations survive alongside: `test-root-regression:` at `:287`, `Run ROOT regression infra-marked tests` at `:663`.

### Both gates green on the rebased head

```
timeout: 60.0s
timeout method: signal
=== 2403 passed, 4 skipped, 23 deselected, 65 warnings in 277.91s (0:04:37) ====
```

`Test DataFlow Infra Regression (Postgres/Redis)` — SUCCESS. **Zero pytest-timeout hits. Zero non-green checks on the PR.**

Count moved 2348 → 2403: L4's #2144 marker-determinism fix (#2152) plus the three new SIGALRM tests below. That shift is expected — `str(item.function)` embeds an ASLR-randomized address, so ~3% of tests were being excluded non-deterministically before #2144.

### The Windows gap is now closed by execution, not by assertion

The `--timeout-method=signal` degrade branch in root `conftest.py` had never been executed anywhere — no workflow this PR touches runs a Windows matrix (`trust-plane.yml` is the repo's only `windows-latest` job; its `paths:` filter is `src/kailash/trust/**`). Three tests now execute it, and they ran on this very job:

```
test_signal_degrades_to_thread_where_sigalrm_is_absent PASSED
test_signal_is_left_intact_where_sigalrm_exists PASSED
test_an_explicitly_requested_thread_method_is_untouched PASSED
```

They simulate the platform condition rather than requiring the platform, so they run on **every** Linux job instead of none.

**Mutation-verified, not asserted.** Removing the guard from `conftest.py`:

```
E  AssertionError: the SIGALRM degrade branch did NOT fire
E  assert 'signal' == 'thread'
1 failed, 2 passed
```

The 2 that stay passing are the negative controls — that is what separates a real red from an always-failing suite. `conftest.py` restored clean afterwards (`git diff --stat` empty).

This **inverts** the earlier Windows result recorded above: a throwaway branch with the same guard removed ran the Windows matrix and came back green on all four legs. That run could not discriminate — the ini timeout is a dead letter and zero tests under `tests/trust/` carry a timeout marker, so no timer is ever armed and `timeout_method` is never consulted. Green there was consistent with the guard being both necessary and unnecessary.

Ready to merge on your call. Not merging.

esperie:
## Stash trap filed as #2158

Per the request to record the near-miss rather than let a third occurrence find it: **#2158**.

A bare `git stash -u` on a **clean** tree stashes nothing and pushes no entry; the paired `git stash pop` then pops whatever was already on top — in my case a pre-existing stash from another session, landing 137 conflicted paths in this worktree. Both commands exit `0`, and with `-q` there is no output at all.

Reproduced in a scratch repo and quoted verbatim in the issue, so the next reader does not have to trust the narrative.

Why it matters beyond the nuisance: the natural remedy for the resulting mess is `git reset --hard` on a dirty tree — which is exactly the operation `.claude/rules/git.md` § "Destructive Working-Tree Ops" gates. The trap manufactures the situation where the documented-risky command looks like the obvious fix.

What made this recoverable rather than a loss, for the record: before resetting I confirmed all four commits were pushed, ran `comm -12` between my commits' file list and the conflicted set (**empty** — no file of mine was involved), and confirmed `stash@{0}` still held the foreign content intact.

Issue #2074, opened 2026-08-11
packages/kailash-kaizen/tests/regression/ is not run by CI — 107 files / 855 tests are green-by-absence


esperie:
## Summary

`packages/kailash-kaizen/tests/regression/` appears in **no** pytest invocation in `.github/workflows/test-kailash-kaizen.yml`. Every regression test written for kaizen in that directory has never run in CI.

```
$ grep -n "tests/regression" .github/workflows/test-kailash-kaizen.yml
  (no matches)

$ ls packages/kailash-kaizen/tests/regression/*.py | wc -l
107
$ grep -rhc "^def test_\|^    def test_" packages/kailash-kaizen/tests/regression/*.py | paste -sd+ - | bc
855
```

The workflow runs two invocations: `tests/unit/llm/` + `tests/cross_sdk_parity/`, then an explicit list of ~30 `tests/unit/<tier>/` directories. `tests/regression/` is in neither, and the list is enumerated by hand, so a new top-level test directory is invisible by default rather than picked up.

## How it was found

PR #2068 added two regression files here for #2030 / #2022. They passed locally and would have merged reporting nothing — never collected, never run. Moving them to `tests/unit/core/` (which IS in the list) produced the first real signal:

```
tests/unit/core/test_issue_2022_config_error_not_swallowed.py ..........  [ 12%]
tests/unit/core/test_issue_2030_agent_io_logging.py ..................    [ 12%]
=== 6642 passed, 88 skipped, 1633 deselected in 215.81s ===
```

Before the move those two lines did not appear at all. That absence is the whole defect: a green job is equally consistent with "the tests passed" and "the tests were never collected", so the job's conclusion carries no information about them.

## Why this is worth fixing rather than absorbing

A regression test's entire purpose is to fail if a fixed bug returns. One that is never collected cannot do that, and it is *worse than no test*, because it advertises coverage that does not exist — the same class as a silent skip, and the class several recent issues (#2022's Tier-3 tests skipping without a model env var; #2038's `continue-on-error`) have been burning down.

I have not audited what the 855 tests would report if switched on; some may be red, some may need markers or infra. **That audit is the work**, and it should not be assumed to be clean.

## Suggested approach

- [ ] Run `tests/regression/` locally/in a scratch job first to establish the true baseline before gating it (expect failures; they are findings, not blockers to filing).
[...]

Issue #1973, opened 2026-07-25
kaizen: a2a.py pre-existing hardening — un-awaited coroutine in latency calc + silent except:pass


esperie:
Surfaced by the Wave-R holistic redteam + editor diagnostics on `packages/kailash-kaizen/src/kaizen/nodes/ai/a2a.py` (pre-existing, NOT introduced by the #1952 a2a residual fix).

## Findings
- Lines ~377/383/389: `Operator '*' not supported for Coroutine and float` — a coroutine appears to be multiplied without being awaited (latency simulation), likely a real async bug producing wrong values or a runtime error.
- Line ~1865: `except Exception: pass` on the summarization `super().run()` — silent swallow with no observability log (a rule-based fallback follows, but the failure is invisible). zero-tolerance R3 / observability.

## Acceptance
- [ ] The coroutine at 377/383/389 is awaited (or the type error resolved)
- [ ] The 1865 silent swallow logs at WARN before falling back
- [ ] Behavioral coverage for both

PR #2160, opened 2026-08-15
fix(kaizen): dead MCP auto-registration, pytest-base-url collision, stale-wheel test placement + wire 812 never-run kaizen-agents tests into CI

esperie:
Two pre-existing kaizen test failures that **block #2159 from merging**. A conftest bug was auto-marking ~15% of the kaizen suite as needing infrastructure it does not need, so these were silently deselected from every infra-free CI step. Fixing that marker bug releases them, and they fail. They are pre-existing, not caused by the marker fix.

## 1. MCP tool auto-registration was dead (product bug)

`test_mcp_tool_wrapper_closure_capture.py::test_pre_fix_wrapper_shape_is_rejected_by_real_registration` failed with `ValueError: Functions with **kwargs are not supported as tools`.

The test's name suggested it might merely be asserting a rejection wrongly. It is not — driving the real path shows the **product** is broken:

```
a.expose_as_mcp_server("probe", tools=["greet"])
-> ValueError: Functions with **kwargs are not supported as tools
```

`expose_as_mcp_server` raised for **every** agent method under FastMCP 2.12.4, so agent-as-MCP-server was dead for every agent. A prior fix removed `_bound_method` from the wrapper signature and filed this second rejection as version-noise "tracked separately"; it is the same defect one layer down. The server builds a tool's published schema from the wrapper's signature and annotations, and a bare `**kwargs` wrapper carries neither.

Fix: `functools.wraps(bound_method)` on the wrapper, so it carries the method's signature (via `__wrapped__`) and annotations. Registration now succeeds **and** the tool advertises the method's real parameters — without it the schema would have been empty even where registration passed, leaving a tool no client could call correctly.

Also: one un-publishable method no longer takes the whole server down. Auto-discovery sweeps every public attribute so it always meets one — those warn and are skipped; an explicit `tools=[...]` entry still raises rather than returning a server silently missing what the caller named.

The test that stood in for this asserted a bare `**kwargs` probe registers — a shape the server rejects — so it pinned framework behaviour that does not hold and could not red against broken source. It now drives `expose_as_mcp_server` and asserts the published schema.

## 2. `pytest-base-url` fixture collision (test bug)

`test_issue_1720_b1b_embed_cutover.py` — every parametrized case errored at setup with `ScopeMismatch: You tried to access the function scoped fixture base_url with a session scoped request object`.

`pytest-base-url` ships a session-scoped `base_url` fixture plus a session-scoped autouse `_verify_url(request, base_url)`. The parametrize argname `base_url` overrode it with a function-scoped one, so the session-scoped consumer could no longer request it.

Fix: renamed the argname to `wire_base_url`. The colliding name is ours and local. Widening the parametrized value to session scope would leak one wire's value across the others, and narrowing a third-party plugin's fixture is not ours to do. A comment records the collision so a later cleanup does not rename it back.

## 3. `kaizen-agents` regression suite wired into CI — 812 tests that had NEVER run

**This is the most consequential change in the PR.**

`packages/kaizen-agents/tests/` appeared in **no pytest invocation in any workflow**. The only `packages/kaizen-agents` matches anywhere in `.github/workflows/` were `paths:` **trigger** filters in `unified-ci.yml` — those decide which workflow runs, not what pytest collects. The entire suite had never executed in CI. Same never-ran class as #2074 and #2038, in a package nobody had checked.

Baseline measured **before** wiring it in, per #2074's instruction not to assume clean:

```
812 passed, 2 deselected, 0 failed
```

Wired into the existing `test-kaizen` job in `test-kailash-kaizen.yml` (no new workflow file). That job already installs `kaizen-agents` editable, so `kaizen_agents` resolves to the branch rather than to whatever wheel is on the runner — load-bearing, see below. `packages/kaizen-agents/**` added to both `paths:` filters; without it a PR touching only that package would still run none of its tests. Job name left unchanged so no required status check is renamed.

Scope: the `regression` tier only, which is verified green. `tests/unit/`, `integration/`, and `e2e/` in that package remain unrun — not baselined here, and not wired blind. That is a separate shard in the #2074 family.

## 4. The two `test_issue_1973_a2a_capability_match` failures — stale wheel, not a product bug

Reported as a third blocker. The propagation was never broken. The tests sat in `packages/kailash-kaizen/`, whose `pytest.ini` sets `pythonpath = src` (its own only), so `kaizen_agents` fell through to an installed wheel:

```
kailash-kaizen ctx : ~/.pyenv/.../site-packages/kaizen_agents/__init__.py   [0.9.11, STALE]
kaizen-agents  ctx : packages/kaizen-agents/src/kaizen_agents/__init__.py   [0.13.0, branch]

0.9.11 : result = matcher(task)
branch : result = matcher(task, config=reasoning_config, correlation_id=correlation_id)
```

The bridge's legacy `except TypeError` fallback then re-calls positionally and `config` arrives `None` — reported as a live config-propagation bug against source that propagates config fine.

Fix: moved `TestReasoningBridgesPreserveConfigPropagation` (3 tests) to `packages/kaizen-agents/tests/regression/`, beside `test_issue_1981_degraded_judgment.py`, which covers the same module for the same reason. `kaizen-agents` depends on `kailash-kaizen`, so a test in the depended-upon package importing its own dependent inverted the dependency. That package's `conftest.py` puts its own `src` at `sys.path[0]`, so the assertions always describe the branch. **No production code changed — both sides were already correct.**

These would have passed in CI in either location (CI installs the sibling editable); the failure was local-only. The relocation is what makes the outcome independent of install mode — and it is only safe because of the CI wiring in section 3.

A related guard was added in `packages/kailash-kaizen/conftest.py`: an advisory terminal-summary report naming any first-party package served from outside the checkout, with the fix command. Advisory, not fatal — testing against a released wheel on purpose is legitimate, and failing would red every suite on any machine with a stale sibling.

## Evidence

RED (before):
```
1 failed, 9 passed, 6 errors in 0.74s
```

GREEN (after):
```
17 passed in 1.84s
```

Both new tests were mutation-checked: removing `functools.wraps` reds both new MCP tests; removing the un-publishable guard alone reds `test_unpublishable_method_does_not_abort_the_whole_server`.

Wider suites: `packages/kailash-kaizen/tests/regression/` 1627 passed, `tests/unit/core/` 1096 passed, `packages/kaizen-agents/tests/regression/` 812 passed / 0 failed under the new CI step's exact command and marker filter.

No `skip`/`xfail` was used, and no test was deleted.

## Related issues

Refs #2160
Unblocks #2159

### DIFF
--- a/.github/workflows/test-kailash-kaizen.yml
+++ b/.github/workflows/test-kailash-kaizen.yml
@@ -278,11 +278,17 @@ on:
     branches: [feat/*, feature/*, fix/*, docs/*, session/*, session-*, chore/*]
     paths:
       - "packages/kailash-kaizen/**"
+      # kaizen-agents is built and tested by this workflow (see the
+      # kaizen-agents regression step). Without this path a PR touching ONLY
+      # kaizen-agents would run none of its tests -- the same never-ran gap
+      # #2074 closed for kailash-kaizen's own regression suite.
+      - "packages/kaizen-agents/**"
       - ".github/workflows/test-kailash-kaizen.yml"
   pull_request:
     branches: [main]
     paths:
       - "packages/kailash-kaizen/**"
+      - "packages/kaizen-agents/**"
       - ".github/workflows/test-kailash-kaizen.yml"
   workflow_dispatch:
 
@@ -467,6 +473,31 @@ jobs:
         # passed, 425 deselected, 8 xfailed, 0 failed.
         timeout-minutes: 10
         working-directory: packages/kailash-kaizen
+        run: |
+          ../../.venv/bin/python -m pytest \
+            tests/regression/ \
+            -m 'not (integration or e2e or requires_azure or requires_docker or requires_google or benchmark or requires_postgres or requires_redis or requires_ollama or requires_db or requires_llm or llm_execution or real_llm or ollama_validation or cost or landing_ai or stress or endurance or requires_infrastructure or requires_kafka or requires_mongodb or memory_intensive or requires_isolation or data_intensive or flaky or slow or load or migration or openai or anthropic or google or docker or ollama or integration_llm or tier3 or performance)' \
+            --timeout=30 --maxfail=50 -q
+
+      - name: Test (kaizen-agents regression suite, infra-free)
+        # packages/kaizen-agents/tests/ appeared in NO pytest invocation in ANY
+        # workflow -- the only `packages/kaizen-agents` matches were the
+        # `paths:` TRIGGER filters in unified-ci.yml, which decide which
+        # workflow runs, not what pytest collects. The whole suite had
+        # therefore never run in CI: the same never-ran gap as #2074 (this
+        # job's kailash-kaizen regression step above) and #2038.
+        #
+        # This job is where it belongs: it already installs kaizen-agents
+        # editable (see the install step), so `kaizen_agents` resolves to the
+        # branch rather than to whatever wheel is on the runner. That
+        # resolution is load-bearing -- against a stale INSTALLED wheel these
+        # tests red while the branch source is already correct.
+        #
+        # Same broad exclusion set as the steps above. Baseline established
+        # before wiring this in (per #2074's instruction not to assume clean):
+        # 812 passed, 2 deselected, 0 failed.
+        timeout-minutes: 10
+        working-directory: packages/kaizen-agents
         run: |
           ../../.venv/bin/python -m pytest \
             tests/regression/ \
--- a/packages/kailash-kaizen/conftest.py
+++ b/packages/kailash-kaizen/conftest.py
@@ -22,6 +22,7 @@ as the un-scrubbed-secret gap this file's cost-guard half already closes.
 import os
 import sys
 from pathlib import Path
+from typing import Iterator, Tuple
 
 import pytest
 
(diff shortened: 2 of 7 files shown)


esperie:
## Third blocker (`test_issue_1973_a2a_capability_match.py`) — not a product or test bug

Added to my lane as two more pre-existing failures. They are **not fixable in code, because the code on both sides is already correct.** No change was made to the product or the tests.

**Root cause: `kaizen_agents` resolves to a stale non-editable wheel that shadows the repo source.**

```
installed: kaizen-agents 0.9.11
branch:    kaizen-agents 0.13.0
```

The `#1973` config-propagation fix exists only in the branch:

```
site-packages/kaizen_agents/patterns/_reasoning_bridge.py:94
    result = matcher(task)                                          # config dropped
packages/kaizen-agents/src/kaizen_agents/patterns/_reasoning_bridge.py
    result = matcher(task, config=reasoning_config, correlation_id=correlation_id)
```

The bridge's legacy `except TypeError` fallback then re-calls positionally, so `config` arrives as `None` — which is exactly what the test reports.

**Why the earlier A/B read as "pre-existing either way":** it swapped the conftest but held `PYTHONPATH` constant, and that path omits `packages/kaizen-agents/src`. Both arms therefore imported the same stale wheel. Varying the one thing that matters:

```
ARM A  imports site-packages/kaizen_agents/...     ->  2 failed, 20 passed
ARM B  imports packages/kaizen-agents/src/...      ->  22 passed
```

**These do not block the root-cause PR.** Both workflows install every sibling editable (`uv pip install -e packages/kaizen-agents` — `test-kailash-kaizen.yml:383`, `unified-ci.yml:350,607`), so CI resolves repo source and these two pass there. The trap is local-only.

Local fix:
```
uv pip install -e packages/kaizen-agents
```

### Guard added

Nothing in the suite pointed at this, in either direction — a stale wheel reds correct code, and a wheel *newer* than the branch greens a regression the branch actually has. Both look like ordinary pytest output. Added an advisory terminal-summary report (`packages/kailash-kaizen/conftest.py`) naming any first-party package served from outside the checkout, with the fix command:

```
============== first-party packages NOT served from this checkout ==============
  kaizen_agents: imported from /Users/.../site-packages/kaizen_agents/__init__.py
                 repo source /Users/.../packages/kaizen-agents/src/kaizen_agents
                 fix: uv pip install -e packages/kaizen-agents
Results above describe the INSTALLED code, not this branch.
```

Advisory, not fatal: testing against a released wheel on purpose is legitimate, and failing would red every suite on any machine with a stale sibling. Say the word if you want it escalated to a hard failure.

### Evidence

All three blocker files together, CI-equivalent resolution: **39 passed**.
Full `tests/regression/` with the guard active: **1630 passed, 1 skipped, 24 xfailed**.

Guard verified in both directions: fires in the stale arm, silent in the clean arm.

Issue #2168, opened 2026-08-16
Design: check_url performs a live DNS lookup during Endpoint construction


esperie:
## Summary

`Endpoint.base_url` routes every URL through `url_safety.check_url()`, which performs a **live DNS lookup** as part of its DNS-rebinding defense. Constructing an `Endpoint` therefore performs network I/O, and fails closed (`InvalidEndpoint: reason=resolution_failed`) whenever the hostname does not resolve.

This issue raises the design question. It does **not** propose a change — the current behaviour may well be correct, and it has security implications either way.

## Evidence

`src/kaizen/llm/deployment.py:129` calls `check_url(v)` inside a Pydantic `field_validator(mode="before")`:

```python
@field_validator("base_url", mode="before")
@classmethod
def _validate_base_url(cls, v: Any) -> Any:
    ...
    check_url(v)
```

`check_url`'s signature offers an opt-out that this call site does not use, and does not expose:

```python
def check_url(url: str, *, resolve_dns: bool = True) -> None:
```

`src/kaizen/llm/url_safety.py:367-371`:

```python
    if not any_resolved and host_lc not in _HTTP_LOCALHOST_ALLOWLIST:
        # Resolution failed entirely — treat as InvalidEndpoint with
        # `resolution_failed`. Caller can retry with resolve_dns=False if the
        # host is intentionally unreachable in the test environment.
        _reject("resolution_failed", url)
```

The comment names an affordance ("caller can retry with `resolve_dns=False`") that **is not reachable through `Endpoint`** — there is no parameter, config field, or context var that plumbs `resolve_dns` from an `Endpoint`/`LlmDeployment` caller down to `check_url`.

## Why this is worth deciding deliberately

1. **Construction is not the moment of egress.** The lookup happens when the object is built, not when a request is sent. A DNS result cached at construction is stale by send time, so the rebinding window it closes is narrow, while the cost (network I/O in a constructor) is paid on every construction.

2. **It makes a validity check depend on ambient network state.** A perfectly well-formed endpoint for a region whose DNS has not yet been published is rejected as invalid. `tests/unit/llm/auth/test_register_bedrock_region.py:104-112` already documents working around exactly this:

   > *"The SSRF guard's DNS resolution check rejects unresolvable hostnames with `resolution_failed`, which is correct behavior — DNS for an unreleased AWS region does NOT resolve until AWS publishes it."*

[...]

PR #2169, opened 2026-08-16
test(kaizen): make the Tier-1 unit tier hermetic (no DNS, no egress)

esperie:
## Summary

**Unblocks `main`.** Five tests under `packages/kailash-kaizen/tests/unit/llm/` are failing on `main` with `kaizen.llm.errors.InvalidEndpoint: reason=resolution_failed`. This makes the Tier-1 unit tier hermetic, which fixes them — and fixes the 250 more that were failing for the same reason without anyone noticing.

## These tests are not a regression

They were **deselected for as long as they have existed**. The old kaizen conftest matched the 2-letter fragment `"ai"` against test names — `azure_**ai**_foundry`, `open**ai**` — and marked them `requires_ollama`. Replacing that heuristic with fixture-graph derivation released them, and they failed on their first ever run. This is first-run exposure.

## The actual defect

`Endpoint.base_url` routes every URL through `url_safety.check_url()` (`deployment.py:129`), whose DNS-rebinding defense calls `socket.getaddrinfo` and rejects with `resolution_failed` when a hostname does not resolve (`url_safety.py:367-371`). **Constructing an endpoint performs a live DNS lookup.**

The dependency stayed invisible because two fake fixture hostnames split by accident:

```
myresource.openai.azure.com        RESOLVES        -> 20.89.12.205   (Azure wildcard record)
myfoundry.services.ai.azure.com    does NOT resolve (gaierror)
```

Both are fake. Both take the same code path. The only reason the other tests passed is that one fake hostname happens to land under a real Azure wildcard record and the runner has network.

Measured by poisoning `socket.getaddrinfo` to raise for every host, against the **unfixed** code:

| scope | with DNS | without DNS |
|---|---|---|
| `tests/unit/llm/` | 5 failed, 1610 passed | **255 failed**, 1360 passed |
| `tests/unit/providers`, `rag`, `strategies` | 323 passed | **6 failed**, 317 passed |

So `tests/unit/` had a hidden network dependency: on an air-gapped or DNS-restricted runner the *passing* tests fail too. 29 of the fixture hostnames resolve only because the runner has real internet DNS.

## The fix

Test-side only, in `packages/kailash-kaizen/tests/unit/conftest.py` — the file that already carries an autouse LLM stub for exactly this reason (`rules/testing.md` § 3-Tier). Two autouse patches:

1. **In-process resolver.** Unknown hosts resolve to a fixed public address; literal IPs pass through unchanged; `localhost` and the cloud-metadata hostnames keep their real addresses. It patches the same seam the existing SSRF rebinding tests already use, so a test that drives resolution itself just re-patches and wins for its own duration.

2. **Non-loopback connects refused.** Resolving a name is only half the dependency. Several tests deliberately drive an ungoverned egress and swallow the result (`except Exception: pass`), so they were fast *only because the lookup failed first*. Once names resolve they black-hole for their full 60s timeout — and on a runner where the address is reachable they would issue a real request from the unit tier.

Two details worth flagging, both found the hard way:

- The RFC 5737 documentation ranges (`192.0.2.0/24`, `198.51.100.0/24`, `203.0.113.0/24`) are unusable as the stub address: Python's `ipaddress` reports all three as `.is_private`, so the guard's own `_is_private_ipv4` would reject them and every endpoint construction would fail.
- Literal IPs must pass through unchanged. Remapping them silently redirected the ollama fallback's `127.0.0.1:11434` to a public address, turning a 0.05s connection refusal into 75s black-holed across its retry budget.

## What was NOT done

`check_url`'s production behaviour is unchanged. Whether a DNS lookup belongs inside endpoint construction is a real design question with security implications — raised as **#2168** with the evidence, not decided here. Note that `check_url`'s own comment names a `resolve_dns=False` affordance that is **not reachable through `Endpoint`**; that is part of #2168.

No test was skipped, xfailed, or given a hostname that happens to resolve.

## Verification

**Direction 1 — the 5 previously-failing tests.**

```
BEFORE: 5 failed, 73 passed in 15.97s
AFTER:  78 passed in 34.92s
```

**Direction 2 — the rest no longer depend on DNS.** Real `socket.getaddrinfo` poisoned to raise for every host, and instrumented to count any call that escapes the stub:

```
tests/unit/llm/          1615 passed in 21.98s
[nodns] real getaddrinfo calls that escaped the stub: 0

full tests/unit/         11151 passed, 152 skipped in 1415s
[nodns] real getaddrinfo calls that escaped the stub: 0
```

The instrument is known to discriminate: the same plugin produced 255 failures against the unfixed code.

`pre-commit run` is green end-to-end, including the `Run Tier 1 unit tests` hook.

## Note for reviewers

Two **pre-existing** wall-clock perf assertions flake under heavy parallel load and are unrelated to this change (both pass 5/5 in isolation, and timings are identical with and without this fix — 4.46/4.68/4.79s vs 4.65/4.38s):

- `tests/unit/examples/test_simple_qa_async.py::TestSimpleQAAsyncPerformance::test_async_execution_overhead` — `assert elapsed < 5.0`, ~4.5s unloaded, so ~10% headroom
- `tests/unit/nodes/test_async_node_comprehensive.py::TestAsyncNodePerformance::test_concurrent_execution_performance`

They are not addressed here; flagging them rather than silently leaving them for the next reader.

## Related issues

Unblocks `main`. Follow-up design question: #2168

### DIFF
--- a/packages/kailash-kaizen/src/kaizen/core/token_counter.py
+++ b/packages/kailash-kaizen/src/kaizen/core/token_counter.py
@@ -283,10 +283,24 @@ class TokenCounter:
 
         # Fallback: character-based estimate
         if not self._tiktoken_warning_shown:
-            logger.warning(
-                "Using character-based token estimation. "
-                "Install tiktoken for accurate counting: pip install tiktoken"
-            )
+            if TIKTOKEN_AVAILABLE:
+                # tiktoken IS installed, so the encoder lookup itself failed --
+                # most often the one-time fetch of the BPE table from
+                # openaipublic.blob.core.windows.net. Telling the operator to
+                # install a package they already have sends them the wrong way;
+                # the preceding "Failed to get encoder" warning carries the
+                # real cause.
+                logger.warning(
+                    "Using character-based token estimation: tiktoken is "
+                    "installed but its encoding data could not be loaded "
+                    "(see the preceding 'Failed to get encoder' warning). "
+                    "Counts are approximate until the encoding is available."
+                )
+            else:
+                logger.warning(
+                    "Using character-based token estimation. "
+                    "Install tiktoken for accurate counting: pip install tiktoken"
+                )
             self._tiktoken_warning_shown = True
 
         return self._estimate_tokens(text)
--- a/packages/kailash-kaizen/tests/integration/core/test_token_counter_tiktoken.py
+++ b/packages/kailash-kaizen/tests/integration/core/test_token_counter_tiktoken.py
@@ -1,0 +1,61 @@
+"""Tier-2 tests for TokenCounter against REAL tiktoken encoding data.
+
+Moved out of ``tests/unit/core/test_token_counter.py``. These assert the
+content of tiktoken's own BPE tables ("Hello, world!" is exactly 4 tokens in
+``cl100k_base``), and tiktoken does not ship those tables in its wheel — it
+fetches them from ``openaipublic.blob.core.windows.net`` on first use. A test
+that downloads 1.7MB before it can assert is an integration test regardless of
+which directory it sits in (``rules/testing.md`` § 3-Tier: Tier-1 MUST NOT
+touch the network).
+
+They passed for as long as they did because a developer machine, and any CI
+runner with a warm tiktoken cache, already had the file on disk. On a cold
+cache with no network ``_get_encoder`` returns None and ``count`` falls back
+to the character estimate, so ``count("Hello, world!")`` returns 3 rather
+than 4.
+
+The Tier-1 half of this contract — that ``count`` DELEGATES to an encoder when
+one exists rather than silently estimating — is covered offline with a stub
+encoder in ``tests/unit/core/test_token_counter.py::TestEncoderDelegation``.
+What lives here is only the part that genuinely needs the real tables.
+"""
+
+import pytest
+
+from kaizen.core.token_counter import TokenCounter
+
+try:
+    import tiktoken
+
+    TIKTOKEN_AVAILABLE = True
+except ImportError:
+    TIKTOKEN_AVAILABLE = False
+
+
+@pytest.mark.integration
+@pytest.mark.skipif(
+    not TIKTOKEN_AVAILABLE,
+    reason="tiktoken not installed - tests require tiktoken",
+)
+class TestTiktokenSpecific:
+    """Tests that specifically require real tiktoken encoding data."""
+
+    def test_tiktoken_exact_count(self):
+        """Test exact token count with tiktoken."""
+        counter = TokenCounter()
+        # "Hello, world!" is exactly 4 tokens in cl100k_base
+        tokens = counter.count("Hello, world!", encoding_name="cl100k_base")
+        assert tokens == 4
+
+    def test_tiktoken_encoding_for_model(self):
+        """Test tiktoken-based encoding selection."""
+        import tiktoken
+
+        counter = TokenCounter()
+        # Get encoding directly from tiktoken
+        expected = tiktoken.encoding_for_model("gpt-4")
+        actual = counter._get_encoder("cl100k_base")
+        assert actual is not None
+        # Verify they produce same result
+        text = "Test text"
+        assert len(expected.encode(text)) == len(actual.encode(text))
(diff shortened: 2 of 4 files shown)


esperie:
## Positive control for the DNS-unavailable demonstration

"Green with DNS stubbed" cannot on its own distinguish *hermetic* from *the stub silently did nothing*. So the stub was deliberately sabotaged two different ways, against `tests/unit/llm/`, with the real resolver still poisoned to raise:

| control | what it breaks | result |
|---|---|---|
| `_stub_getaddrinfo` raises `gaierror` | fixture installed but useless | **255 failed**, 1360 passed |
| `_STUB_PUBLIC_IPV4` → `10.0.0.1` | stub answers, but with a private address | **255 failed**, 1360 passed |
| neither (as merged) | — | **1615 passed**, 0 escapes |

Both controls red **exactly the 255** tests that fail against unfixed code, so the fixture is demonstrably load-bearing rather than inert.

The second control is the sharper one: it fails for a *different reason*, which shows the guard consumes the address the stub returns and not merely the fact that resolution succeeded.

```
$ POISON=private pytest tests/unit/llm/test_deployment_resolver_azure.py ... | grep -oE "reason=[a-z_]+" | sort | uniq -c
   8 reason=private_ipv4
```

versus `reason=resolution_failed` under the raising control.

Rebased onto `main` @ `77f1af4ed`; both directions re-verified after the rebase (1615 passed with DNS, 1615 passed with the resolver poisoned, 0 real `getaddrinfo` calls escaping).

esperie:
## Second hidden Tier-1 network dependency, found by the egress block

The connect-refusal added here caught a second pre-existing dependency in the same class as the DNS one. tiktoken does not ship its BPE tables in the wheel — it fetches `cl100k_base.tiktoken` from `openaipublic.blob.core.windows.net` on first use. Two tests in `tests/unit/core/test_token_counter.py` asserted against those tables, so they passed only where the tiktoken cache was already warm.

Reproduced by pointing `TIKTOKEN_CACHE_DIR` at an empty directory — exactly those 2 fail, the other 36 pass:

```
Failed to get encoder cl100k_base: HTTPSConnectionPool(host='openaipublic.blob.core.windows.net',
port=443): ... NewConnectionError(... [Errno 61] Tier-1 unit tests MUST NOT touch the network
(blocked connect to ('93.184.216.34', 443)); see tests/unit/conftest.py)
```

`_get_encoder` then returns None and `count` falls back to the character estimate, so `count("Hello, world!")` returns 3 instead of 4.

### Not moved wholesale to Tier 2

Moving both would have **deleted** the coverage rather than relocating it:

- no workflow invokes `packages/kailash-kaizen/tests/integration/`;
- the kaizen job additionally runs `-m 'not (integration or e2e or ...)'`;
- `test-kailash-kaizen.yml` already carries a comment recording this same failure mode for `tests/regression/` under #2074 — *"had therefore never run in CI."*

So the halves are split by what each actually needs:

- **Tier 2** — assertions that need the real BPE tables (exact count of 4, parity with `tiktoken.encoding_for_model`). Those test tiktoken's data, not ours.
- **Tier 1** — the regression-relevant half, offline, via a stub encoder: that `count()` **delegates** to an available encoder instead of silently estimating. The stub counts words, which differs from the character estimate for the same string, so a regression to the fallback is visible rather than merely plausible.

### A test that could not fail

`TestEncoderCaching` guarded its assertion with `if encoder1 is not None:` — on any runner without the encoding data it asserted nothing and passed. Zero caching coverage, presenting as green. It now asserts unconditionally and checks the underlying lookup ran exactly once for two calls.

**Positive control:** forcing `TokenCounter._get_encoder` to return None reds all three (both delegation tests and the caching test), so they discriminate.

### Warning message

The one-time fallback WARNING said *"Install tiktoken for accurate counting: pip install tiktoken"*, which sends an operator to reinstall a package they already have when the real cause is a failed encoding fetch. It now branches on `TIKTOKEN_AVAILABLE` and points at the preceding `Failed to get encoder` line.

### Verification

Full unit tier under **both** adverse conditions simultaneously — cold tiktoken cache and the real resolver poisoned to raise:

```
11152 passed, 152 skipped, 0 failed in 838s
```

Reconciles exactly with the earlier single-condition run (11151 passed + 1 pre-existing perf flake = the same 11152 collected; the flake passed this time).

One honesty note on the instrument: that run used `-n 8`, and the `[nodns]` escape counter aggregates per-process, so it reports the controller only and printed nothing. The **0 escapes** figure quoted earlier comes from the serial `tests/unit/llm/` runs, where the counter is meaningful. The poisoning itself still applies in every xdist worker, so the green result stands on its own.

### Not decided here

Wiring `tests/integration/` into CI is a real coverage gap (#2074 is the precedent), but it adds CI cost and job time, so it is a maintainer call rather than something to slip into this PR.

Issue #2171, opened 2026-08-16
Nexus rate-limit middleware fingerprints a client IP with an unkeyed hash — the privacy it claims does not hold


esperie:
Surfaced during the open-HIGH CodeQL triage. Not the finding CodeQL reported at this line (that is alert 11465, `py/clear-text-logging`, already dispositioned under #2044) — this is a different, real defect at the same site, which is why alert-by-alert triage would not have caught it.

## The code and its own stated intent

`packages/kailash-nexus/src/nexus/auth/rate_limit/middleware.py:182-190`:

```python
# identifier is whatever _identifier_extractor returns -- an IP or
# user id (PII), and by convention sometimes an API key. Log a
# fingerprint so operators can still correlate repeat offenders;
# the raw value stays confined to the rate-limit backend key.
logger.warning(
    "Rate limit exceeded: identifier_fp=%s, path=%s, retry_after=%ds",
    fingerprint_secret(str(identifier)),
    path,
    retry_after,
)
```

The stated goal is that **"the raw value stays confined to the rate-limit backend key"** — i.e. the log gets a correlation tag, not the PII.

## Why the goal is not met

`fingerprint_secret` is unkeyed BLAKE2b truncated to 8 hex chars (`src/kailash/utils/url_credentials.py:730-733`). The default extractor is documented at `:30` and `:99-100` as returning **"user_id from JWT or IP address"**.

For the IP case the input space is 2^32. Against an unkeyed fast hash that is exhaustible in minutes on commodity hardware, so the fingerprint in the log is equivalent to the plaintext IP for anyone who can read the log. The privacy control does not do the job its comment describes.

`fingerprint_secret`'s own docstring is consistent with this and is not the problem — it says plainly:

> Fingerprints MUST NOT be treated as secrets ... anyone who knows the plaintext can reproduce the fingerprint.

The defect is the **call site** relying on a property the helper never claimed.

## Contrast: the keyed helper exists

Three lines below `fingerprint_secret` in the same module:

```python
_PROCESS_KEY_MATERIAL = _secrets.token_bytes(32)

def process_local_config_key(payload: str, *, length: int = 16) -> str:
```

whose docstring states the exact reasoning this call site needs: *"Because the key is secret and random, the derived cache key is not a brute-forceable oracle for the credentials in its pre-image, which a bare digest would be."*

[...]

Issue #2170, opened 2026-08-16
kaizen/llm/url_safety.py documents its URL tag as "non-reversible" when it is not (Rule 3e)


esperie:
Surfaced during the open-HIGH CodeQL triage while investigating alert 11474.

## The false claim

`packages/kailash-kaizen/src/kaizen/llm/url_safety.py:44`:

```python
def _url_fingerprint(raw: str | None) -> str:
    """Produce a short, non-reversible tag for a URL.
```

It delegates to `fingerprint_secret`, which is **unkeyed** BLAKE2b truncated to 8 hex characters (32 bits):

```python
# src/kailash/utils/url_credentials.py:730-733  (function spans 667-733)
digest_bytes = max(1, (length + 1) // 2)
return hashlib.blake2b(value.encode("utf-8"), digest_size=digest_bytes).hexdigest()[
    :length
]
```

A URL is **low-entropy and enumerable**. Against an unkeyed fast hash, anyone holding the logged tag can enumerate candidate URLs and recover the original. "Non-reversible" is false for this input class.

## The canonical helper says so itself

`fingerprint_secret`'s own docstring, same file:

> Fingerprints MUST NOT be treated as secrets. They are derived deterministically from the plaintext with no secret keying material; **anyone who knows the plaintext can reproduce the fingerprint.** Do not use fingerprints for access control or verification.

So the SDK already documents the property correctly in one place and incorrectly in another. That is the `zero-tolerance.md` Rule 3e shape: a doc claim about a code-surface property that does not hold.

## Why it is load-bearing rather than cosmetic

The keyed helper **exists** and is a different function three lines below:

```
667: def fingerprint_secret(...)              <- unkeyed; _url_fingerprint calls THIS
740: _PROCESS_KEY_MATERIAL = _secrets.token_bytes(32)
743: def process_local_config_key(...)        <- keyed; NOT what url_safety uses
```

`process_local_config_key`'s docstring explains exactly why keying matters: *"Because the key is secret and random, the derived cache key is not a brute-forceable oracle for the credentials in its pre-image, which a bare digest would be."*

A reader who trusts "non-reversible" may fingerprint something that genuinely needs to stay secret, and get no protection. The docstring is the interface here.

## Acceptance criteria

- [ ] Either correct the docstring to state what the tag actually is (a stable, **enumerable-input-reversible** correlation tag, not a confidentiality control), or switch the call to keyed material if confidentiality is intended — decide which, do not leave both readings available.
[...]

PR #2176, opened 2026-08-16
fix(security): redact credentials at every log sink they reach (#2167)

### LINKED ISSUE #2167: Credentials reach log sinks unredacted: AuditLogNode event_data and HTTPRequestNode url/headers/body (CodeQL 11503-11507)
Two REAL defects from the open-HIGH CodeQL triage. Both are credential material reaching a clear-text log sink. Five alerts, two underlying bugs.

| Alert | Site | Defect |
|---|---|---|
| 11503 / 11504 / 11505 | `src/kailash/nodes/security/audit_log.py:113,115,117` | A (one bug — three sinks share one value) |
| 11506 / 11507 | `src/kailash/nodes/api/http.py:893,888` | B |

## Defect A — `AuditLogNode` logs caller `event_data` unredacted

The three alerts are one `if/elif/else` over a single `log_message` built once at `:104-109`, so one fix closes all three.

`event_data` is a free-form caller bag passed through `sanitize_log_structure` → `sanitize_log_value`. That helper's own docstring states it is **not** the control needed here:

> **NOT a redaction step.** A short attacker-chosen value survives on purpose -- it is usually the only diagnostic there is. For withholding credential material by key name, compose `kailash.nodes.auth._log_hygiene.redact_mapping` instead.

So a credential up to `_DEFAULT_LOG_VALUE_CHARS = 256` reaches the log **byte-intact**. Non-printables are flattened and length is bounded — that is anti-log-injection hygiene, and it is working; it is the redaction claim that does not hold.

**Real unredacted producers, not hypothetical:**
- `packages/kailash-kaizen/src/kaizen/nodes/auth/sso.py:160-171` passes the raw IdP attribute bag as `"raw_attributes": attributes`. The core-SDK sibling for the identical bag *does* redact — `src/kailash/nodes/auth/sso.py:1094` uses `redact_mapping`. That divergence is the SDK's own classification of the bag as credential-bearing.
- `src/kailash/nodes/auth/session_management.py:1143` splices `**(metadata or {})` in raw.
- `src/kailash/runtime/local.py:4217-4222`, `:4297-4302` forward `event_data` raw.
- `src/kailash/nodes/auth/directory_integration.py:420-425` passes raw directory-sync stats.

**The fix location is the repo's own stated design**, `src/kailash/nodes/auth/_log_hygiene.py:23-26`:

> The durable fix belongs one layer down, inside `SecurityEventNode.execute` and `AuditLogNode.execute`, so that EVERY caller in the SDK is covered rather than only this package's.

`SecurityEventNode` is included: its `log_message` does **not** carry `metadata`, so its exposure is via the returned-and-rendered bag rather than the log line, but it is the same class and the same prescribed fix.

## Defect B — `HTTPRequestNode` / `AsyncHTTPRequestNode` log credentials

Flagged lines log `url` with no masking, on the **default** path (`log_requests` defaults False, so `:893` fires on every async request). No masking helper is imported anywhere in `http.py`. The repo's canonical masker states the binding contract this violates, `src/kailash/utils/url_credentials.py:299-312`:

> Single source of truth for credential masking across the entire Kailash codebase. Every log line, error message, exception payload, or stdout writeback that mentions a connection string MUST route the URL through this helper first.

`mask_url` explicitly handles `http://`/`https://`, masking `user:password@` userinfo and credential query params. A caller-supplied `https://svc:secret@api/...` or a presigned `...?X-Amz-Signature=...` is logged verbatim.

**The unflagged lines in the same block are worse than the flagged ones:**

```python
889:  self.logger.info(f"Headers: {headers}")          # Authorization: Basic <base64>
891:  self.logger.info(f"Body: {json_data or data}")   # OAuth token exchange body
```

`_apply_authentication:774-776` builds `base64.b64encode(f"{auth_username}:{auth_password}")` — an encoding, not a mask, so the password is trivially recoverable. And `src/kailash/nodes/auth/sso.py:968-973` really does POST a token exchange through this node with `client_secret` in `token_data`.

**The sync class is byte-identical and was NOT flagged.** `http.py:539-544` mirrors `:888-893`; CodeQL only reported the async class. Fixing one and leaving the other is the `rules/security.md` § Multi-Site Kwarg Plumbing shape — both must land together.

## Acceptance criteria

- [ ] `AuditLogNode.execute` composes `redact_mapping` at the `event_data` boundary; `SecurityEventNode.execute` does the same for `metadata`. Existing `sanitize_log_structure` behaviour (log-injection defence) is preserved, not replaced.
- [ ] `redact_mapping` lives at a canonical location importable by `nodes/security/` without reaching into a private `nodes/auth/` module; existing importers updated, no duplicate implementation left behind.
- [ ] `http.py` routes `url` through `mask_url`, redacts `headers`, and redacts or drops the body log — in **both** `HTTPRequestNode` and `AsyncHTTPRequestNode`.
- [ ] Regression tests that FAIL on current code and pass after, for each defect, asserting on the emitted log stream (a forged or leaked value is produced at format time, so counting `LogRecord`s cannot discriminate).
- [ ] Alerts 11503, 11504, 11505, 11506, 11507 close on the next analysis.

## Related

- Triage of all 36 open HIGH alerts; existing dispositions #2044 / #2106 / #2127 / #2146
- `rules/security.md` § Multi-Site Kwarg Plumbing (the sync/async pair)
- `zero-tolerance.md` Rule 3 (a control that reports success without doing its job)

esperie:
## Summary

Five CodeQL HIGH alerts pointed at two sinks. The regression test written for them surfaced a third that is far wider than either.

**A — audit sinks (alerts 11503/11504/11505).** `AuditLogNode` / `SecurityEventNode` passed the caller's free-form `event_data` bag through `sanitize_log_structure` alone. That helper's own docstring says it is **not** a redaction step — it decides *how* a value renders, not *whether* it may be recorded — so a credential up to its 256-char bound reached the log byte-intact. Redaction is now **composed** with sanitization rather than substituted for it: dropping either half trades a credential leak for a log-injection hole, or the reverse.

**B — HTTP nodes (alerts 11506/11507).** `HTTPRequestNode` / `AsyncHTTPRequestNode` logged the request URL, the assembled `Authorization` header, and the request body unmasked. Both classes are fixed; CodeQL flagged only the async one and the sync sibling is byte-identical.

**C — `Node.execute` (found by B's test, flagged by nothing).** It logged every node's full validated input dict. That is not one node's leak but **every** node's: `auth_password`, `api_key`, `secret_key`, and connection strings all landed in clear text. DEBUG is not a defence — debug logging is routinely enabled in staging and shipped to an aggregator. The call is guarded by `isEnabledFor` so the structure walk costs nothing when DEBUG is off, and keys are preserved so the diagnostic still reports which inputs bound.

`redact_mapping` moves from `nodes/auth/_log_hygiene` (private to that package) to `kailash.utils.secure_logging`, beside `sanitize_log_value`, because nodes outside that package must not reach into its private module. The five auth call sites follow the move. No deprecation shim — `_log_hygiene` is private.

## Verification

The tests assert on the **emitted stream** via a real handler and formatter. A credential is rendered at format time, so inspecting `LogRecord.args` cannot tell a redacted call from an unredacted one.

Discrimination established, not assumed:

| run | result |
| --- | --- |
| the 11 new tests vs **unfixed main** | **10 failed** |
| the 11 new tests vs this branch | **11 passed** |

The 11th is the deliberate no-false-positive control (`test_http_node_still_logs_a_usable_diagnostic`) and passes on both by design.

Full local sweep (`tests/unit tests/security tests/regression`, untruncated both sides):

```
main   : 115 failed, 7141 passed, 50 errors
branch : 114 failed, 7153 passed, 50 errors
new failures introduced: 0
```

`pre-commit` run per-file across all 13 changed files: 14 hooks each, 0 failures.

## Related issues

Fixes #2167

## CodeQL deferrals on this PR — Rule 1b

Two sets of alerts are deferred as false positives rather than suppressed. **No suppression comments, no rule disabling, no config changes** — the alerts stay visible and each has a tracking issue carrying a runtime-safety proof.

| Tracking issue | Alerts | Disposition |
|---|---|---|
| **#2178** | 5 × `py/clear-text-logging-sensitive-data` at the `http.py` **fix lines** | The rule follows the taint into the masking helpers; each guard was executed against a password-bearing input to prove no credential survives. |
| **#2181** | 11474 (`py/weak-sensitive-data-hashing`), 11484 + 11485 (`py/clear-text-logging`) | Rule premises do not hold: authentication completes before any hashing and no fingerprint gates an auth decision; the two logging sinks carry a filesystem path matched by identifier NAME. |

**Rule 1b condition 4 — release-specialist signoff — is NOT given.** These should not be dismissed in the security tab until it is recorded, or until dismissed by someone with the authority to do so.

**Scope caveat carried from #2181, repeated here because it is the sentence most likely to be misread:** the 11474 disposition covers **the rule only**. `fingerprint_secret` is genuinely unkeyed and low-entropy inputs genuinely reach it — that weakness is real and tracked as **#2170** (a docstring claiming "non-reversible" when it is not) and **#2171** (a rate-limit privacy control that does not conceal an enumerable IP). Neither is closed by this dismissal.

### DIFF
--- a/src/kailash/nodes/api/http.py
+++ b/src/kailash/nodes/api/http.py
@@ -6,6 +6,7 @@ the best features from both the original HTTPRequestNode and HTTPClientNode.
 
 import asyncio
 import base64
+import json
 import time
 from enum import Enum
 from typing import Any
@@ -35,6 +36,41 @@ from kailash.utils.resource_manager import (
     ResourcePool,
     managed_resource,
 )
+from kailash.utils.secure_logging import redact_mapping, sanitize_log_value
+from kailash.utils.url_credentials import mask_error_text, mask_url
+
+
+def _log_safe_body(body: Any, limit: int = 500) -> str:
+    """Render a request/response body for a log line without leaking credentials.
+
+    Issue #2167. These nodes logged bodies verbatim, and the bodies genuinely
+    carry credentials in both directions: ``nodes/auth/sso.py`` POSTs an OAuth
+    token exchange through this node with ``client_secret`` in the payload, and
+    the RESPONSE to that exchange is a JSON object whose ``access_token`` field
+    is the credential itself.
+
+    Structured bodies keep their shape and lose only credential-named values, so
+    the diagnostic survives. An OPAQUE body -- a raw string or bytes that is not
+    JSON -- is withheld entirely and reported by type and size: there are no keys
+    to redact by, so there is no way to withhold the secret half while keeping
+    the rest, and guessing would be the "control that reports success without
+    doing its job" shape.
+    """
+    if body is None:
+        return "<empty>"
+    if isinstance(body, (dict, list)):
+        return sanitize_log_value(str(redact_mapping(body)), limit)
+    if isinstance(body, (bytes, bytearray)):
+        return f"<{len(body)} bytes withheld>"
+    if isinstance(body, str):
+        try:
+            parsed = json.loads(body)
+        except (ValueError, TypeError):
+            return f"<{len(body)} chars withheld>"
+        if isinstance(parsed, (dict, list)):
+            return sanitize_log_value(str(redact_mapping(parsed)), limit)
+        return f"<{len(body)} chars withheld>"
+    return f"<{type(body).__name__} withheld>"
 
 
 class HTTPMethod(str, Enum):
@@ -535,13 +571,31 @@ class HTTPRequestNode(Node):
             request_kwargs["data"] = data
 
         # Execute request with retries
+        # CREDENTIAL-SAFE LOGGING (issue #2167, CodeQL 11506/11507).
+        #
+        # `url` goes through `mask_url` because it is the SDK's single source of
+        # truth for credential masking -- its own docstring: "Every log line,
+        # error message, exception payload, or stdout writeback that mentions a
+        # connection string MUST route the URL through this helper first." It
+        # handles http(s) userinfo (`https://svc:pw@host`) and credential query
+        # params (a presigned `?X-Amz-Signature=...`). Nothing in this file
+        # imported any masker before this change.
+        #
+        # `headers` is redacted because `_apply_authentication` puts the
+        # credential THERE: `Authorization: Basic <base64(user:pass)>`. Base64
+        # is an encoding, not a mask -- the password is trivially recoverable
+        # from a log. The redactor is separator-normalized so the default
+        # `X-API-Key` header name matches too.
+        #
+        # NOTE the else-branch is the DEFAULT path (`log_requests` defaults
+        # False), so the unmasked URL was logged on EVERY request.
         if log_requests:
-            self.logger.info(f"Request: {method} {url}")
-            self.logger.info(f"Headers: {headers}")
+            self.logger.info("Request: %s %s", method, mask_url(url))
+            self.logger.info("Headers: %s", redact_mapping(headers))
             if data or json_data:
-                self.logger.info(f"Body: {json_data or data}")
+                self.logger.info("Body: %s", _log_safe_body(json_data or data))
         else:
-            self.logger.info(f"Making {method} request to {url}")
+            self.logger.info("Making %s request to %s", method, mask_url(url))
 
         response = None
         response_time: float = 0.0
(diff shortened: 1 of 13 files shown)


esperie:
## CodeQL: 1 real finding fixed, 5 deferred with proof

**Fixed (real):** `py/unused-import` at `session_management.py:25` — dead `redact_mapping` import. It was already unused on main and this PR only relocated it, which is why CodeQL saw it as new. Removed rather than carried; "unchanged on main" is not a reason to keep dead code past a scan that surfaced it.

**Deferred (5 HIGH `py/clear-text-logging-sensitive-data`, tracked in #2178):** the flagged lines *are the fix* — `mask_url(url)`, `redact_mapping(headers)`, `_log_safe_body(...)`. CodeQL's taint tracker sees `url` reaching a log call and does not model the in-source sanitizers as barriers (the same limitation measured under D1, where both `sanitizers.model.yml` forms were non-working for in-source functions).

Runtime-safety proof, executed against password-bearing inputs:

```
PASS  http.py:594/979  mask_url(url) [userinfo] -> https://***@api.example.com/v1
PASS  http.py:594/979  mask_url(url) [query]    -> https://api.example.com/v1?api_key=%2A%2A%2A
PASS  http.py:975      redact_mapping(headers)  -> {'Authorization': '[REDACTED]', 'X-API-Key': '[REDACTED]'}
PASS  http.py:977      _log_safe_body(body)     -> {'password': '[REDACTED]', 'user': 'alice'}
```

Dismissal is a repo-settings mutation and is the co-owner's call, not mine — #2178 exists so the deferral is tracked rather than silent.

## Addendum since the PR opened

Probing `redact_mapping` **directly** (not through a sink) found three key-matching gaps that were live even with the sink tests green — sink tests only exercise the keys producers in this repo happen to emit:

- `pwd` was uncovered (`passwd` contains no adjacent `p-w-d`, so that fragment never matched it)
- dotted (`X.API.Key`, the SAML/OIDC attribute-bag shape) and spaced (`api key`) keys missed the `api_key` fragment, because only `-` was folded

Each new assertion was verified to RED against the pre-fix matcher.

**Walking sets was tried and reverted.** Measured, it closed nothing: a set holds only hashables, so it never contains a dict, so there is no key to match — the walk returned the same credential as a list. A shape-pinning test records that decision so the walk is not re-added later in the belief that it redacts something.

Known accepted limitation: a credential inside an innocently-named free-text value (`{"note": "user pw is …"}`) is not reachable by key-based or URL-based redaction.

Issue #1952, opened 2026-07-24
fix(kaizen): remaining silent-mock / fabricated-content sites after #1947 (env fallback + EmbeddingGeneratorNode)


esperie:
Follow-up to #1947, surfaced by its /redteam. #1947 closed the silent-mock (fabricated-content-as-real) class at `LLMAgentNode` and `IterativeLLMAgentNode`. Two same-class sites remain, each requiring its own construction-site prod-audit + test sweep (out of #1947's node-scoped shard):

## 1. `detect_provider_from_env()` returns `"mock"` when keyless (HIGH — highest blast radius)

`packages/kailash-kaizen/src/kaizen/core/_provider_env.py` — the env resolver's final fallback is `return "mock"`. The Agent surface (`Agent._get_provider_for_config()`) and ~30 RAG dispatch sites pass this resolved string straight into `LLMAgentNode`'s `provider`. After #1947, a real (non-test) caller with **no** `OPENAI_API_KEY`/`ANTHROPIC_API_KEY` and no explicit provider still silently gets `"mock"` (never `None`), so it passes the #1947 fail-loud gate and returns fabricated content. This is the primary user-facing surface (Agent + RAG), triggered by a common misconfiguration (forgotten/mis-named key).

This is deliberately scoped out of #1947 (the `_provider_env.py` docstring names it a separate, higher-blast-radius follow-up) because changing the fallback to fail-loud alters dispatch for every Agent/RAG site and anyone relying on keyless mock.

Note: this is **distinct** from #1948 item 3 (`workflow_generator.py` `llm_provider or "openai"`) — a silent-*openai* fallback errors loudly at the API boundary rather than fabricating content. The `-> "mock"` fabrication risk is not captured by that framing.

## 2. `EmbeddingGeneratorNode` silent-mock (MED)

`packages/kailash-kaizen/src/kaizen/nodes/ai/embedding_generator.py:239` — `provider = kwargs.get("provider", "mock")`, and its `get_parameters()` provider param declares no default. A forgotten provider yields silent mock embeddings returned as real vectors — the same integrity class on the embeddings surface.

## Acceptance criteria

- [ ] Audit every Agent/RAG/embedding construction site for keyless behavior; decide the fail-loud contract for the env-fallback path (raise vs require explicit `provider="mock"` for keyless dev).
- [ ] `EmbeddingGeneratorNode` fails loud (or requires explicit `provider="mock"`) on an unresolved provider.
[...]

PR #2185, opened 2026-08-16
fix(kaizen): stop shipping a mock embedder as the production default (#2174)

### LINKED ISSUE #2174: kaizen/memory/vector.py ships a MOCK embedder as the production default (zero-tolerance Rule 2)
Surfaced incidentally during the open-HIGH CodeQL triage while checking alert 11251 (`py/weak-sensitive-data-hashing`, MD5). **The alert itself is a false positive** — the hashed input is arbitrary memory text, not credential material. The real defect is what the hash is doing there at all.

## The defect

`packages/kailash-kaizen/src/kaizen/memory/vector.py`:

```python
self.embedding_fn = embedding_fn or self._default_embedder
```

and `_default_embedder` is, by its own docstring and comments:

```python
"""Default mock embedder for testing."""
# Simple hash-based embedding for testing
hash_val = int(hashlib.md5(text.encode()).hexdigest(), 16)
```

A caller who does not pass `embedding_fn` gets **hash-derived pseudo-embeddings** presented as real ones. That is `zero-tolerance.md` Rule 2 — a mock in a production path — and the failure mode is the worst kind: it does not raise, it does not warn, it returns plausible vectors of the right shape.

## Why this is the sharpest finding in the batch

Every downstream operation still "works": similarity search returns results, ranked, with scores. They are just **meaningless** — MD5 of the text carries no semantic signal, so nearest-neighbour over these vectors is nearest-neighbour over hash noise. A user gets confidently wrong retrieval with no error, no log line, and no way to notice from the outside.

This is the same shape as the rest of the session's findings: **a control that reports success without doing its job.** An audit sanitizer that does not redact, a privacy fingerprint that does not conceal, a permission check that cannot run, an embedder that does not embed.

## Secure-default question

Per `rules/security.md` § Secure-Default, a feature whose default makes it a silent no-op must instead fail closed **or** emit a loud one-time WARN naming the wiring. Neither happens here.

## Acceptance criteria

- [ ] `_default_embedder` no longer serves as the production default. Preferred: fail closed — require `embedding_fn`, raising a typed error naming the wiring. If backward-compat forbids that, emit a loud one-time WARN at init/first-use naming what is degraded and how to wire a real embedder.
- [ ] If retained for tests, rename to make its status unmistakable at the call site (e.g. `_hash_embedder_for_tests`) and keep it out of the default path.
- [ ] A test asserting the default path does NOT silently produce hash-derived vectors.
- [ ] Sweep for sibling "default mock" wiring elsewhere in `kaizen/memory/` and `kaizen/nodes/ai/` — this pattern rarely appears once.

## Related
- Alert 11251 — false positive for its own rule; this issue is what was found while checking it
- `zero-tolerance.md` Rule 2; `rules/security.md` § Secure-Default

esperie:
Fixes #2174.

## The defect

`VectorMemory.__init__` did:

```python
self.embedding_fn = embedding_fn or self._default_embedder
```

and `_default_embedder` was, by its own docstring, `"""Default mock embedder for testing."""` — MD5 of the text expanded to a 128-dimensional 0/1 vector.

A caller who omitted `embedding_fn` got hash-derived pseudo-embeddings presented as real ones. Nothing raised, nothing logged, and everything downstream kept working: similarity search returned ranked results, with scores. The ranking was over MD5 noise.

## Disposition: fail closed

`rules/security.md` § Secure-Default allows two outcomes — fail closed, or a loud one-time WARN if backward-compat forbids it. **The WARN branch was not needed**: a sweep of every construction found no production caller relying on the default. Every one was a test, an example, or a docs snippet. So `embedding_fn` is now required and its absence raises `ConfigurationError` naming the wiring.

This mirrors **#1952** on `EmbeddingGeneratorNode`, which removed the identical hazard one surface over — *"a forgotten provider used to become 'mock' here and silently return fabricated mock vectors as real embeddings."* Same shape, same disposition; the two surfaces now agree.

The hash function survives as a module-level `hash_embedder_for_tests`, out of the default path. The name is the whole point: as `VectorMemory._default_embedder` it silently served everyone who omitted the argument.

## Sibling sweep (AC4) — one worse instance, one verified clean

**`SimpleEmbeddingProvider.embed_text` — worse than the original.** Both the non-200 arm and a bare `except Exception:` fell back to `_hash_embedding`, with no raise and no log:

```python
except Exception:
    # Fallback to simple hash-based embedding
    embedding = self._hash_embedding(txt)
```

and the returned `EmbeddingResult` carried `model=self.model_name` — so the fabricated vectors were **labelled with the real model's provenance** and were indistinguishable downstream. `hybrid_search.py:337` constructs this provider, so the degraded path was live, not theoretical.

Both arms now raise a typed `EmbeddingUnavailable`. `_hash_embedding` is **deleted** rather than left reachable.

**`EmbeddingGeneratorNode._generate_mock_embedding` — CLEAN.** Reachable only through an explicit `provider == "mock"`; #1952 already removed the silent default. Recorded here so nobody re-checks it.

## Collateral: the examples and docs taught the broken pattern

- **`docs/reference/memory-patterns-guide.md`** documented `VectorMemory(embedding_fn=None, top_k=5)` under the heading *"Default: Mock hash-based embedder (testing only)"*. The defect was the documented API.
- **`memory-showcase/demo.py`** printed *"Key Feature: Semantic search - finds related conversations"* while ranking hash noise, with its questions annotated `# Related to Q1`. It now wires the stand-in explicitly, states plainly that the ranking is not semantic, and shows the real-embedder swap.
- **`rag-research/workflow.py`** forwards `memory_config["embedder"]`, which was `None` whenever unset — the forwarded-default case a call-site grep does not surface. Now documented as required.

## Verification

**Positive control**, because a test that cannot fail is the same defect class as the bug. Reverting `vector.py` to `main` leaves the new tests RED, then GREEN after restore:

```
vector.py at main:  4 failed, 1 passed
    E   Failed: DID NOT RAISE ConfigurationError
    E   Failed: DID NOT RAISE ConfigurationError
    E   AssertionError: assert not True
         +  where True = hasattr(VectorMemory, '_default_embedder')
after restore:      5 passed
```

The one test that passes both ways is `test_supplied_embedder_is_the_one_used` — supplying an embedder always worked, so it correctly does not discriminate on this change.

**The sibling tests point at a non-loopback host deliberately.** `SimpleEmbeddingProvider`'s default is `http://localhost:11434`, which the unit tier's egress block permits — on a developer machine with ollama running, the request *succeeds*, and a test written against the default would pass for the wrong reason and then fail in CI. I hit exactly that (`DID NOT RAISE`) while writing them.

```
1193 passed, 10 skipped   (tests/unit/memory + tests/unit/nodes + tests/unit/examples)
```

Full-tier run in progress; will report.

## Related issues

Fixes #2174

### DIFF
--- a/packages/kailash-kaizen/docs/reference/memory-patterns-guide.md
+++ b/packages/kailash-kaizen/docs/reference/memory-patterns-guide.md
@@ -235,8 +235,18 @@ class LongConversationAgent(BaseAgent):
 ```python
 from kaizen.memory.vector import VectorMemory
 
-# Default: Mock hash-based embedder (testing only)
-memory = VectorMemory(embedding_fn=None, top_k=5)
+# embedding_fn is REQUIRED. Omitting it (or passing None) raises
+# ConfigurationError. It used to fall back to a hash-based pseudo-embedder,
+# which returned vectors of the right shape with no semantic signal, so
+# similarity search silently ranked MD5 noise (#2174).
+memory = VectorMemory(embedding_fn=custom_embedder, top_k=5)
+
+# For tests ONLY, an explicitly-named stand-in is available. It carries no
+# semantic signal — use it to exercise storage/retrieval mechanics, never to
+# assert on ranking quality.
+from kaizen.memory.vector import hash_embedder_for_tests
+
+memory = VectorMemory(embedding_fn=hash_embedder_for_tests, top_k=5)
 
 # Custom embedder (production)
 def custom_embedder(text: str) -> List[float]:
--- a/packages/kailash-kaizen/examples/1-single-agent/memory-showcase/demo.py
+++ b/packages/kailash-kaizen/examples/1-single-agent/memory-showcase/demo.py
@@ -21,7 +21,7 @@ from kaizen.core.base_agent import BaseAgent, BaseAgentConfig
 from kaizen.memory.buffer import BufferMemory
 from kaizen.memory.knowledge_graph import KnowledgeGraphMemory
 from kaizen.memory.summary import SummaryMemory
-from kaizen.memory.vector import VectorMemory
+from kaizen.memory.vector import VectorMemory, hash_embedder_for_tests
 from kaizen.signatures import InputField, OutputField, Signature
 
 # ============================================================================
@@ -200,8 +200,18 @@ def demo_vector_memory():
         "Perfect for: Large knowledge bases, RAG applications, finding related context"
     )
 
-    # Create agent with VectorMemory
-    memory = VectorMemory(top_k=2)
+    # Create agent with VectorMemory.
+    #
+    # NOTE: this demo wires the explicitly-named test stand-in so it runs with
+    # no embedding service. That function hashes text, so the retrieval below
+    # is NOT semantically meaningful — it demonstrates the STORAGE and
+    # RETRIEVAL mechanics only. Swap in a real embedder to see actual semantic
+    # matching:
+    #
+    #     from sentence_transformers import SentenceTransformer
+    #     encoder = SentenceTransformer("all-MiniLM-L6-v2")
+    #     memory = VectorMemory(top_k=2, embedding_fn=lambda t: encoder.encode(t).tolist())
+    memory = VectorMemory(top_k=2, embedding_fn=hash_embedder_for_tests)
     agent = DemoAgent(DemoConfig(), memory=memory)
 
     # Conversation with semantically related and unrelated questions
@@ -231,6 +241,10 @@ def demo_vector_memory():
 
     print(
         "\n  Key Feature: Semantic search - finds related conversations, not just recent"
+    )
+    print(
+        "  (This run uses hash_embedder_for_tests, so the ranking above is NOT"
+        " semantic. Wire a real embedder to see the feature itself.)"
     )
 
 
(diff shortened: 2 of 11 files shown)


esperie:
## CI caught a gap my local sweep had

First push went red. My local verification ran `tests/unit/`, but the kaizen job also runs `tests/regression/`, and:

```
tests/regression/test_issue_f31_fu5_numpy_lazy_import.py::test_hash_embedding_and_search_still_work_with_numpy_present
AttributeError: 'SimpleEmbeddingProvider' object has no attribute '_hash_embedding'
```

Recording it rather than quietly fixing it, because the shape is worth noting: **"the full unit tier is green" was true and still not sufficient** — the suite I chose did not cover the job's scope. Same family as the truncation traps from earlier today, one level up: a complete answer to a narrower question than the one that mattered.

### Why the test was repointed rather than deleted or relaxed

That test's stated purpose is *"the lazy-import refactor must not change numeric behavior"*, and the numeric path it actually pins is `InMemoryVectorStore.search_similar`, which resolves numpy via `require_numpy("vector similarity search")`. `_hash_embedding` was **incidental** there — a convenient source of a 384-dim ndarray, not the subject of the regression.

So the probe vector is now built directly through `require_numpy`. F31/FU5 coverage is intact and arguably tighter: the helper under test is called explicitly in the test rather than reached through a production method that no longer exists. Shape, store round-trip, and cosine self-similarity assertions are unchanged.

Renamed to `test_vector_search_still_works_with_numpy_present` — the old name described hash embedding, which the test no longer exercises and which no longer exists. Confirmed the name appears in no workflow or pytest config before renaming.

### Verification after the fix

```
tests/regression/    1643 passed, 1 skipped, 24 xfailed
tests/unit/         11160 passed, 152 skipped        (11152 baseline + the 8 new tests here)
```

CI is now green on `d05b00123`, including the kaizen job that was red.

PR #2192, opened 2026-08-18
fix(auth): mounted-subapp declaration no longer claims a WARN that never fired (#2113)

### LINKED ISSUE #2113: mounted_subapp_auth_kwargs asserts a WARN that never fired when the parent used external_auth_reason
Deferred from #2100. Low severity — a false statement in an audit record, not an open route.

## What

`src/kailash/utils/server_auth.py::mounted_subapp_auth_kwargs` is binary on `parent_is_authenticated`:

```python
if parent_is_authenticated:
    state = "which authenticates every request before routing to this mount"
else:
    state = ("which has authentication explicitly disabled; it logged that "
             "exposure once at its own construction")
```

Callers pass `parent_is_authenticated=self._auth_config is not None`. But `_auth_config` is `None` in **two** different situations:

1. `require_auth=False` — the parent logged `server_auth.disabled` (a WARNING). The string above is true.
2. `external_auth_reason=...` — the parent logged `server_auth.external` (an INFO). **The string above is false**: no WARN was ever emitted, and authentication is not "explicitly disabled", it is provided by something outside the server.

Under Nexus, case 2 is the normal path (Nexus installs its own `nexus.auth.jwt.JWTMiddleware` and declares it externally), so **every mounted workflow gets a record asserting a warning that never fired**.

## Why it matters

The whole point of `external_auth_reason` is that the declaration names who authenticates. A derived record that contradicts it degrades the audit trail an operator uses to answer "was this mount gated, and by what?" — and it points them at a WARN they will not find in the log.

## Acceptance criteria

1. Make the state **tri-state** rather than binary: authenticated-by-parent / explicitly-disabled / externally-authenticated. The caller must pass enough information to distinguish case 1 from case 2 (the parent knows which branch `resolve_server_auth` took).
2. The external case names what the parent declared, rather than claiming a WARN.
3. A test per branch asserting the rendered string, including one that fails if the external case renders the disabled text.

Refs #2072, #2100

esperie:
Recovered during a worktree consolidation sweep — this branch existed only in a local worktree and had **never been pushed**. A worktree prune would have destroyed it.

## What

`mounted_subapp_auth_kwargs` asserted a WARN that never fired. The declaration claimed a warning path the code did not take, so an operator relying on that warning to notice a misconfigured mounted sub-app got silence.

Reason is now a tristate rather than a claim, with the WARN emitted where the declaration says it is.

## State — read before merging

- Rebased onto current `main` (`96ab3d87a`); 0 behind.
- Its regression suite passes: `tests/regression/test_issue_2113_mounted_subapp_auth_reason_tristate.py` → **6 passed**.
- **NOT independently re-verified beyond that.** The commit predates the release work by several days and has not had the discrimination check this session applied elsewhere — nobody has established that those 6 tests RED against unfixed code. Treat the suite as necessary, not sufficient, and establish the red before merging.

## Related issues

Fixes #2113


esperie:
Red established — the 6 tests discriminate.

Method: in the PR worktree (rootdir pinned and verified in the same invocation), reverted ONLY the three source files to `main` while keeping the test file, then re-ran.

Positive control that the revert reached the code under test:

```
MODULE_FILE: .../l4-post-merge-verify/src/kailash/utils/server_auth.py
HAS_PARAM (expect False): False
```

Against unfixed code:

```
rootdir: /Users/esperie/repos/kailash/build/.kailash-py-wt/l4-post-merge-verify
test_parent_authenticated_by_its_own_gate                                   PASSED
test_parent_explicitly_disabled_auth                                        FAILED
test_parent_externally_authenticated_does_not_claim_a_warn_that_never_fired FAILED
test_workflow_server_mounts_sub_app_naming_the_real_external_reason         FAILED
test_workflow_api_gateway_mounts_sub_app_naming_the_real_external_reason    FAILED
test_workflow_server_disabled_case_still_names_the_real_warn                PASSED
========================= 4 failed, 2 passed =========================
```

Restored from a `cp` backup; `git status --porcelain` empty; re-ran → `6 passed`.

The 2 that pass on both sides are the control arms — the parent-authenticated branch and the `WorkflowServer` disabled branch are unchanged by this fix, so passing on both sides is correct for them, not vacuity. The 4 that flip are exactly the tristate branch and the two call-site plumbing assertions.

Checks on head `0209de2bceb9d419a8981f2e15071f6857c9d1d6`: 23 pass, 11 skipping, 0 failing.

